PR #1, opened 2020-04-02
Profiling endpoints for mcs

Alevsk:
start and stop profiling endpoints, test includes and many fixes for
policies


minio_copyright.txt
```diff
@@ -0,0 +1,15 @@
+This file is part of MinIO Console Server
```
harshavardhana, line 1: We shouldn't add this file, it should be just called NOTICE which is already added in the repo

Alevsk: done

cesnietor: you might neet to do `make swagger-gen` again since that will remove all the copyright lines from all the code.

PR #4, opened 2020-04-02
Users List & Creation

bexsoft:
## What does this do?
This PR fixes the Users list & User creation modal. Still need to add the new functionality for Groups (WIP).

Also there's an issue with incomplete PATH that is being fixed on this.

## How to test?
- Pull branch
- run yarn install inside portal-ui folder
- run mcs in port 5000 and review


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/elazarl/go-bindata-assetfs v1.0.0
+	github.com/go-bindata/go-bindata v3.1.2+incompatible // indirect
```
dvaldivia, line 7: this line is going to get removed every time we do `go mod tidy`

bexsoft: Ok, is there a way to avoid this?

dvaldivia: on the `mcs` folder run `go mod tidy` and the only do `go get` for the `gobinddata` inside the `portal-ui` folder

PR #28, opened 2020-04-06
Secure Middleware

Alevsk:
adding secure middleware to enforce security headers, most
of the options can be configured via env variables

### DIFF
--- a/cmd/mcs/main.go
+++ b/cmd/mcs/main.go
@@ -58,7 +58,7 @@ var appCmds = []cli.Command{
 
 func newApp(name string) *cli.App {
 	// Collection of m3 commands currently supported are.
-	commands := []cli.Command{}
+	var commands []cli.Command
 
 	// Collection of m3 commands currently supported in a trie tree.
 	commandsTree := trie.NewTrie()
--- a/restapi/config.go
+++ b/restapi/config.go
@@ -17,10 +17,15 @@
 package restapi
 
 import (
+	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/minio/minio/pkg/env"
 )
+
+var Port = 9090
+var TLSPort = 9443
 
 func getAccessKey() string {
 	return env.Get(McsAccessKey, "minioadmin")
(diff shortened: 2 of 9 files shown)


restapi/consts.go
```diff
@@ -21,4 +21,26 @@ const (
 	McsAccessKey   = "MCS_ACCESS_KEY"
 	McsSecretKey   = "MCS_SECRET_KEY"
 	McsMinIOServer = "MCS_MINIO_SERVER"
+	ProductionMode = "PRODUCTION_MODE"
```
harshavardhana, line 24: All ENVs as a practice should be prefixed with MCS_ to avoid any standard ENV name clashes. Like in `mc` all ENVs have `MC_` and minio its `MINIO_`

restapi/consts.go
```diff
@@ -21,4 +21,26 @@ const (
 	McsAccessKey   = "MCS_ACCESS_KEY"
 	McsSecretKey   = "MCS_SECRET_KEY"
 	McsMinIOServer = "MCS_MINIO_SERVER"
+	ProductionMode = "MCS_PRODUCTION_MODE"
+	Hostname       = "MCS_HOSTNAME"
+	TlsHostname    = "MCS_TLS_HOSTNAME"
```
harshavardhana, line 26: Tls is called TLS generally capitalize it.

harshavardhana: you may also use `Mcs` prefix for all constants to be consistent with other constants?

restapi/config.go
```diff
@@ -57,3 +58,121 @@ func getMinIOEndpointIsSecure() bool {
...
+	if allowedHosts != "" {
+		return strings.Split(allowedHosts, ",")
+	} else {
+		return []string{}
+	}
+}
+
+// If SSLRedirect is set to true, then only allow HTTPS requests. Default is false.
+func getSSLRedirect() bool {
+	return strings.ToLower(env.Get(SecureSSLRedirect, "false")) == "true"
```
harshavardhana, line 123: I think we should avoid `true/false` instead perhaps use `on/off` to be consistent with server behavior for its boolean flags.

cmd/mcs/main.go
```diff
@@ -63,9 +67,22 @@ func main() {
...
+	server.Host = restapi.GetHostname()
 	server.ConfigureAPI()
 	server.Port = *portFlag
 
+	tlsCertificatePath := *tlsCertificateFlag
+	tlsCertificateKeyPath := *tlsCertificateKeyFlag
+	if tlsCertificatePath != "" && tlsCertificateKeyPath != "" {
+		server.TLSCertificate = flags.Filename(tlsCertificatePath)
+		server.TLSCertificateKey = flags.Filename(tlsCertificateKeyPath)
+		// We need to discuss with team if running both servers (http & https) is required
```
harshavardhana, line 79: I think we should default http redirect to https if https is enabled by default, listening is okay on http as well. But nothing is served from `http`.

restapi/config.go
harshavardhana: This perhaps is dependent on `https` is enabled or not, so it may not be needed as ENV to turn on/off

restapi/config.go
```diff
@@ -57,3 +58,121 @@ func getMinIOEndpointIsSecure() bool {
...
+	if allowedHosts != "" {
+		return strings.Split(allowedHosts, ",")
+	} else {
+		return []string{}
+	}
+}
+
+// AllowedHostsAreRegex determines, if the provided AllowedHosts slice contains valid regular expressions. Default is false.
+func getSecureAllowedHostsAreRegex() bool {
+	return strings.ToLower(env.Get(SecureAllowedHostsAreRegex, "false")) == "true"
```
harshavardhana, line 86: I think we should avoid true/false instead perhaps use on/off to be consistent with server behavior for its boolean flags.

cmd/mcs/main.go
```diff
@@ -30,6 +30,10 @@ import (
 
 var portFlag = flag.Int("port", 9090, "Port to run this service on")
 
+// TLS certificate configuration flags
+var tlsCertificateFlag = flag.String("tls-certificate", "", "")
```
harshavardhana, line 34: perhaps we should now migrate to the `minio/cli` package before we get a lot of options? Golang flag parser is quite limited..

dvaldivia: I can do that on a separate PR if you'd like @Alevsk

dvaldivia: added on #31

PR #36, opened 2020-04-06
mcs delete bucket event notification api

cesnietor:
To delete do `DELETE` on `/api/v1/buckets/{bucket_name}/events/{arn_name}`
Response will be empty if successful. Returns same errors as minio/mc.


swagger.yml
```diff
@@ -182,6 +182,28 @@ paths:
             $ref: "#/definitions/error"
       tags:
         - UserAPI
+  /buckets/{bucket_name}/events/{name}:
```
dvaldivia, line 185: the `event` definition doesn't return a name, should it be id? most of the them id is also empty. 

```
notificationConfig:
    type: object
    required: 
      - arn
    properties:
      id:
        type: string
      arn:
        type: string
      events:
        type: array
        items:
          $ref: "#/definitions/notificationEventType"
        title: "filter specific type of event. Defaults to all event (default: '[put,delete,get]')"
      prefix:
        type: string
        title: "filter event associated to the specified prefix"
      suffix:
        type: string
        title: "filter event associated to the specified suffix"
```

cesnietor: it is `arn`

cesnietor: id is not returned

cesnietor: example would be `/api/v1/buckets/bucktest1s/events/arn:minio:sqs::test:postgresql`

dvaldivia: does that delete all the bucket events? how could you possibly delete only one? multiple events share the same `arn`

cesnietor: that deletes the one specifically for that bucket. It doesn't delete the event, it deletes the bucket notification on that bucket. you can still list the events that were created. https://docs.min.io/docs/minio-bucket-notification-guide.html#PostgreSQL

dvaldivia:
@bexsoft my only concern is that it deletes all the events for the bucket, regardless, I created a PR for `mc` so we can delete only 1 event https://github.com/minio/mc/pull/3145

PR #42, opened 2020-04-07
Implements remove user API

bexsoft:
## What does this do?

This PR implements the remove user API


restapi/admin_users.go
```diff
@@ -129,3 +137,32 @@ func getUserAddResponse(params admin_api.AddUserParams) (*models.User, error) {
 	}
 	return user, nil
 }
+
+//removeUser invokes removing an user on `MinioAdmin`, then we return the response from API
+func removeUser(client MinioAdmin, accessKey string) error {
```
cesnietor, line 142: can you please add ctx as an input parameter? similar to how we do it on other functions, I think addUser will also need to have it like that to keep consistency but that can be a separate pr

bexsoft: Sure! WIP

bexsoft: Updated

PR #47, opened 2020-04-07
Logout endpoint

Alevsk:
Delete in memory session when user logout from mcs


restapi/user_logout_test.go
```diff
@@ -0,0 +1,21 @@
...
+	mcCmd "github.com/minio/mc/cmd"
+	"github.com/minio/mcs/restapi/sessions"
+)
+
+// TestLogout tests the case of deleting a valid session id
+func TestLogout(t *testing.T) {
+	cfg := mcCmd.Config{}
+	// Creating a new session
+	sessionID := sessions.GetInstance().NewSession(&cfg)
+	// Test Case 1: Delete a session Valid sessionID
```
cesnietor, line 15: we should probably start by checking that 
the sessionID is on sessions:
Pseudo:
```
s := sessions.GetInstance()
_, ok := s.sessions[sessionID] // (here I wouldn't use ValidateSession) cause that function should be tested separately
verify that it is ok
logout
_, ok := s.sessions[sessionID]  
verify that !ok
```
This will actually test that the session got deleted

Alevsk: `ValidateSession` already has some test in `sessions_test.go`. The logout function internally checks if the session got deleted correctly

Alevsk: I was also thinking that we should mock the sessions singleton, right now we are using the same ones as mcs use, but maybe on a different PR since all this session mechanism will change with an incoming change to support STS from Minio :)

cesnietor: `logout` function checks inside but we are technically testing that the logout function is doing its job, that's why I propose that.
Up to you.

PR #49, opened 2020-04-08
Users-Group Update API

bexsoft:
## What does this do?

This PR Creates the Users-Group Update functions, when the user sets a groups list, for each item if is not assigned and is in the list, the group will be assigned to the user, if the group is assigned originally & is not in the new list, the group will be removed from the memberOf array

## How to test

1. In postman Go, set verb as `PUT` & URL as `/api/v1/users/{user_name_to_modify}/groups`
2. As body, send an object with this:
```
{
   "groups": ["group_1", "group_2", "group_3"]
}
```

3. Send the request, the API must return the user with the groups assigned on the top
4. Remove one of the groups of the body request and send it again
5. The API must return the user without the removed user


restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
 	log.Println("User removed successfully:", params.Name)
 	return nil
 }
+
+// updateUserGroups invokes getGroupDescription() to get the current users in the specified group on `MinioAdmin`,
+// then we add/delete the user to the specified groups and return the response to the client
+
+func updateUserGroups(ctx context.Context, client MinioAdmin, user string, groupsToAssign []string) (*models.User, error) {
+	parallelUserUpdate := func(groupName string, originGroups []string) chan bool {
```
dvaldivia, line 186: Can you add a comment for future generations to understand why we have to update each group individually?

bexsoft: sure!

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
 	log.Println("User removed successfully:", params.Name)
 	return nil
 }
+
+// updateUserGroups invokes getGroupDescription() to get the current users in the specified group on `MinioAdmin`,
+// then we add/delete the user to the specified groups and return the response to the client
+
```
cesnietor, line 184: there should be no space before the function so that it appears as documentation on the function.
```suggestion
```

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
 	log.Println("User removed successfully:", params.Name)
 	return nil
 }
+
+// updateUserGroups invokes getGroupDescription() to get the current users in the specified group on `MinioAdmin`,
```
cesnietor, line 182: I don't see the function calling getGroupDescription is that the right comment on the function?

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
...
+
+			userToAddRemove := []string{user}
+
+			gAddRemove := madmin.GroupAddRemove{
+				Group:    groupName,
+				Members:  userToAddRemove,
+				IsRemove: isRemove,
+			}
+
+			err := client.updateGroupMembers(ctx, gAddRemove)
```
cesnietor, line 217: we already have a function for this, `updateGroupMembers()` on admin_groups.go we should use it to avoid duplicate code.

restapi/admin_users_test.go
```diff
@@ -32,6 +32,8 @@ import (
 var minioListUsersMock func() (map[string]madmin.UserInfo, error)
 var minioAddUserMock func(accessKey, secreyKey string) error
 var minioRemoveUserMock func(accessKey string) error
+var minioUpdateUserGroupsMock func(user string, groupsToAssign []string) (*madmin.UserInfo, error)
+var minioGetUserInfo func(accessKey string) (madmin.UserInfo, error)
```
cesnietor, line 36:
```suggestion
var minioGetUserInfoMock func(accessKey string) (madmin.UserInfo, error)
```
To keep consistency we need to call it mock to know it is part of testing

restapi/admin_users.go
bexsoft: Added 
```
// Each group must be updated individually because there is no way to update all the groups at once for a user,
	// we are using the same logic as 'mc admin group add' command
```
in line 244 to understand why the channels are being processed in that way

restapi/admin_users_test.go
```diff
@@ -48,6 +50,16 @@ func (ac adminClientMock) removeUser(ctx context.Context, accessKey string) erro
...
 }
 
+//mock function of updateUserGroups()
+func (ac adminClientMock) updateUserGroups(ctx context.Context, user string, groupsToAssign []string) (*madmin.UserInfo, error) {
+	return minioUpdateUserGroupsMock(user, groupsToAssign)
+}
+
+//mock function of getUserInfo()
+func (ac adminClientMock) getUserInfo(ctx context.Context, accessKey string) (madmin.UserInfo, error) {
+	return minioGetUserInfo(accessKey)
```
cesnietor, line 60: we need to call it here the mock function with the context minioGetUserInfoMock(ctx, accesskey)

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
...
+			if err != nil {
+				log.Println("Error updating", groupName, err)
```
cesnietor, line 219: no need to log this since it will be logged on the top level function that calls this. Else, it will appear when we test it.

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
...
+
+		return chProcess
+	}
+
+	userInfoOr, err := client.getUserInfo(ctx, user)
+
+	if err != nil {
+		log.Println("error getting user:", err)
+		return nil, err
+	}
```
cesnietor, line 237: could you please abstract this and add tests to it? similar to how we are doing for instance on `groupInfo()`

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
...
+
+		if !success {
+			channelHasError = true
+		}
+	}
+
+	if channelHasError {
+		errRt := errors.New("there was an error updating the groups")
+		return nil, errRt
+	}
```
cesnietor, line 264: we should return error instead of bool  in the goroutine that will handle this part of the code and jus return nil if no error.

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
...
+
+	userInfo, err := client.getUserInfo(ctx, user)
+
+	if err != nil {
+		log.Println("error getting user:", err)
+		return nil, err
+	}
+
```
cesnietor, line 272: same comment `could you please abstract this and add tests to it? similar to how we are doing for instance on groupInfo()`

restapi/admin_users.go
```diff
@@ -167,3 +178,125 @@ func getRemoveUserResponse(params admin_api.RemoveUserParams) error {
...
+	// create a minioClient interface implementation
+	// defining the client to be used
+	adminClient := adminClient{client: mAdmin}
+
+	user, err := updateUserGroups(ctx, adminClient, params.Name, params.Body.Groups)
+
+	if err != nil {
+		log.Println("error getting user", err)
```
cesnietor, line 299: here we can just log `error updating user's groups:`

restapi/utils.go
```diff
@@ -30,3 +32,29 @@ func DifferenceArrays(a, b []string) []string {
 	}
 	return diff
 }
+
+// IsElementInSlice returns true if the string belongs to the slice
```
cesnietor, line 36: we would need tests for this functions as well

restapi/utils.go
```diff
@@ -30,3 +32,29 @@ func DifferenceArrays(a, b []string) []string {
...
+	for _, e := range a {
+		if e == b {
+			return true
+		}
+	}
+
+	return false
+}
+
+// UniqueKeys returns an array without duplicated keys
```
cesnietor, line 49: we would need tests for this functions as well

restapi/admin_users_test.go
```diff
@@ -32,6 +32,8 @@ import (
 var minioListUsersMock func() (map[string]madmin.UserInfo, error)
 var minioAddUserMock func(accessKey, secreyKey string) error
 var minioRemoveUserMock func(accessKey string) error
+var minioUpdateUserGroupsMock func(user string, groupsToAssign []string) (*madmin.UserInfo, error)
```
cesnietor, line 35: this function will never be used, here we only mock functions that are defined under MinioAdmin interface, in this pr you only added getUserInfo so that is the only one that needs to be mocked.

restapi/admin_users.go
bexsoft: That comment was from a previous iteration of this function, forgot to update the comment. Updating it accordingly

restapi/admin_users.go
bexsoft: Removed space

restapi/admin_users.go
bexsoft: Updated comment to: 
```
// updateUserGroups invokes getUserInfo() to get the old groups from the user,
// then we merge the list with the new groups list to have a shorter iteration between groups and we do a comparison between the current and old groups.
// We delete or update the groups according the location in each list and send the user with the new groups from `MinioAdmin` to the client
```

restapi/admin_users_test.go
bexsoft: Updated function name

restapi/admin_users.go
bexsoft: Removed log line

restapi/admin_users_test.go
bexsoft: Resolved

restapi/admin_users_test.go
bexsoft: Removed Mocked

restapi/admin_users.go
bexsoft: Removed and lonked to updateGroupMembers from admin_groups

restapi/admin_users.go
bexsoft: Updated error message & added groups instead of error

restapi/utils.go
bexsoft: Added test suite for utils library

restapi/utils.go
bexsoft: Added test suite for utils library

restapi/admin_users.go
bexsoft: Abstracted in a separated function & created tests for it

restapi/admin_users.go
bexsoft: Abstracted in a separated function & created tests for it

cesnietor:
LGTM tested

PR #50, opened 2020-04-08
API to list and add Notification Endpoints

dvaldivia:
Adds APIs to list all notification endpoints (arns) with status details and to add a new Notification Endpoint for any of the supported services


restapi/admin_config.go
```diff
@@ -141,9 +142,8 @@ func getConfigResponse(params admin_api.ConfigInfoParams) (*models.Configuration
 }
 
 // setConfig sets a configuration with the defined key values
-func setConfig(client MinioAdmin, name *string, kvs []*models.ConfigurationKV, arnResourceID string) error {
-	config := buildConfig(name, kvs, arnResourceID)
-	ctx := context.Background()
+func setConfig(ctx context.Context, client MinioAdmin, configName *string, kvs []*models.ConfigurationKV) error {
```
dvaldivia, line 145: I rewired this function so it can be re-used by other functions that need to set a configuration

swagger.yml
```diff
@@ -1057,3 +1092,74 @@ definitions:
...
+          - nsq
+          - mysql
+          - postgres
+          - elasticsearch
+          - redis
+      account:
+        type: string
+      properties:
+        type: object
+        additionalProperties:
```
cesnietor, line 1164: making it `additionalProperties` makes it ambiguous we are using currently `configurationKV` for the configuration which makes more clear what we expect. Regardless, both, `configuration` and notifications, should be consistent on how we build them and handle them

restapi/admin_notification_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// set all the config values if found on the param.Body.Properties
+	for k, val := range params.Body.Properties {
+		configs = append(configs, &models.ConfigurationKV{
+			Key:   k,
+			Value: val,
+		})
+	}
+	// if account was provided, set it as part of the endpoint config
+	if params.Body.Account != "" {
```
cesnietor, line 144: this will always be true since you evaluate above `if params.Body.Account == "" `

restapi/admin_notification_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+
+	err := setConfig(ctx, client, &configName, configs)
+	if err != nil {
+		log.Println(err)
+		return nil, err
+	}
+
+	props := params.Body.Properties
+	// don't return any password field
+	delete(props, "password")
```
cesnietor, line 156: why not? you are admin.

restapi/admin_notification_endpoints_test.go
```diff
@@ -0,0 +1,114 @@
...
+		params *admin_api.AddNotificationEndpointParams
+	}
+	tests := []struct {
+		name          string
+		args          args
+		mockSetConfig func(kv string) error
+		want          *models.NotificationEndpoint
+		wantErr       bool
+	}{
+		{
```
cesnietor, line 44: can you  add a small description what each test does and what it is expected to test?

restapi/admin_notification_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+		log.Println("error getting notification endpoint list:", err)
+		return nil, err
+	}
+	return notfEndpointResp, nil
+}
+
+func addNotificationEndpoint(ctx context.Context, client MinioAdmin, params *admin_api.AddNotificationEndpointParams) (*models.NotificationEndpoint, error) {
+	if params.Body.Account == "" {
+		return nil, errors.New("`account` can't be empty")
+	}
```
cesnietor, line 106: not needed if we do validation, just add `required` on the definition

restapi/admin_notification_endpoints_test.go
```diff
@@ -0,0 +1,114 @@
...
+				params: &admin_api.AddNotificationEndpointParams{
+					HTTPRequest: nil,
+					Body: &models.AddNotificationEndpoint{
+						Account: "1",
+						Properties: map[string]string{
+							"host":     "localhost",
+							"user":     "user",
+							"password": "passwrd",
+						},
+						Service: "postgres",
```
cesnietor, line 58: since we are performing a Switch, yes, we need to test all the cases. To have the highest coverage as possible.

restapi/admin_notification_endpoints_test.go
```diff
@@ -0,0 +1,114 @@
...
+				return errors.New("invalid config")
+			},
+			want:    nil,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// mock function response from setConfig()
+			minioSetConfigKVMock = tt.mockSetConfig
```
cesnietor, line 103: nice! I like this way.

dvaldivia: 👯

restapi/admin_notification_endpoints.go
dvaldivia: We should debate this with AB, I think `mc` does return the password

swagger.yml
dvaldivia: but `additionalProperties` is the way to make a map/dictionary on OpenAPI, I feel wrapping it in an array it's not very resty, lets discuss it on slack.

swagger.yml
```diff
@@ -1057,3 +1092,74 @@ definitions:
...
+          - amqp
+          - kafka
+          - mqtt
+          - nats
+          - nsq
+          - mysql
+          - postgres
+          - elasticsearch
+          - redis
+      account:
```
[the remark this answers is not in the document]
dvaldivia: I removed that check since the account is optional

swagger.yml
```diff
@@ -1057,3 +1092,74 @@ definitions:
...
+        type: string
+      properties:
+        type: object
+        additionalProperties:
+          type: string
```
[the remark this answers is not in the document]
dvaldivia: lol, that's true, it slip my mind

restapi/admin_config.go
```diff
@@ -178,7 +171,16 @@ func setConfigResponse(name string, configRequest *models.SetConfigRequest) erro
 	// create a MinIO Admin Client interface implementation
 	// defining the client to be used
 	adminClient := adminClient{client: mAdmin}
-	if err := setConfig(adminClient, swag.String(name), configRequest.KeyValues, configRequest.ArnResourceID); err != nil {
+	configName := name
+	// if arnResourceID is not empty the configuration will be treated as a notification target
+	// arnResourceID will be used as an identifier for that specific target
+	// docs: https://docs.min.io/docs/minio-bucket-notification-guide.html
+	if configRequest.ArnResourceID != "" {
```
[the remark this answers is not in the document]
dvaldivia: I know, but the `setConfig` function should stay true to it's capability so it can be re-used by other functions, the arn concatenation has more to do with notifications, perhaps I can wrap the `setConfig` to keep the response more like the rest.

restapi/admin_notification_endpoints_test.go
dvaldivia: the name of the tests is meant to describe what it's testing

restapi/admin_notification_endpoints_test.go
dvaldivia: done

swagger.yml
```diff
@@ -1057,3 +1092,74 @@ definitions:
...
+          - webhook
+          - amqp
+          - kafka
+          - mqtt
+          - nats
+          - nsq
+          - mysql
+          - postgres
+          - elasticsearch
+          - redis
```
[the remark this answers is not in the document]
cesnietor: how was this resolved? I still see it duplicated on `notificationEndpoint` and `notificationEndpointItem`

restapi/admin_config.go
```diff
@@ -141,26 +142,28 @@ func getConfigResponse(params admin_api.ConfigInfoParams) (*models.Configuration
...
 	if err := client.setConfigKV(ctx, *config); err != nil {
 		return err
 	}
 	return nil
 }
 
-// buildConfig builds a concatenated string including name and keyvalues
-// e.g. `region name=us-west-1`
-func buildConfig(name *string, kvs []*models.ConfigurationKV, arnResourceID string) *string {
+func setConfigWithARN(ctx context.Context, client MinioAdmin, configName *string, kvs []*models.ConfigurationKV, arn string) error {
```
cesnietor, line 153: so it is not a ARN itself, it is an ID so the function should be called like setConfigWithArnResourceID or something like that.
The docs says :
```
Note that, you can add as many PostgreSQL server endpoint configurations as needed by providing an identifier (like "1" in the example above) for the PostgreSQL instance and an object of per-server configuration parameters
```
so we can use setConfigWithInstanceID or something like that to be consistent, just as a proposal

cesnietor, line 153:
```suggestion
func setConfigWithARN(ctx context.Context, client MinioAdmin, configName *string, kvs []*models.ConfigurationKV, arnResourceID string) error {
``` since it is not an arn

File: restapi/admin_notification_endpoints.go [hunk 14]
```diff
@@ -0,0 +1,178 @@
...
+
+	// set all the config values if found on the param.Body.Properties
+	for k, val := range params.Body.Properties {
+		configs = append(configs, &models.ConfigurationKV{
+			Key:   k,
+			Value: val,
+		})
+	}
+	// if account was provided, set it as part of the endpoint config
+	if params.Body.Account != "" {
```
cesnietor, line 141: what is Account? is it the ID then we need to use here `setConfigWith...` to use the ID

swagger.yml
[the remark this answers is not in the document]
dvaldivia: they are different that `notificationEndpointItem ` has a `status` and doesn't returns the `properties`, it's a light weight listing

restapi/admin_config.go
dvaldivia: but this is taking the whole ARN, not the instance id, it's assumed that the instance ID is part of the ARN, on mc it is not treated separately, there's places where the ARN instance ID is treated separately (adding a new notification endpoint) but that code in turn builds the new ARN for the config name (notify_postgres:1,notify_postgres:3,notify_mysql:instanceID)

dvaldivia: the R in ARN stands for Resource (ARN = Amazon Resource Name) so ARNResourceId would mean amazon resource name resource identifier xD hahaha

restapi/admin_notification_endpoints.go
dvaldivia: yes, it's the ID, this is the code I mentioned above that handles it separately, internally it's called AccountID so I'll rename this field

swagger.yml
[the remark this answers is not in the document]
cesnietor: this is regarding the service :
```
service:
        type: string
        enum:
          - webhook
          - amqp
          - kafka
          - mqtt
          - nats
          - nsq
          - mysql
          - postgres
          - elasticsearch
          - redis
```
this can be abstracted in a separate definition

restapi/admin_config.go
cesnietor: so not resource it should be `account_id` : from docs.
```
arn:partition:service:region:account-id:resource-id
```
that is what you are defining not a whole ARN

cesnietor: the arn is the whole thing `arn:partition:service:region:account-id:resource-id`
```
The following are the general formats for ARNs. The specific formats depend on the resource. To use an ARN, replace the italicized text with the resource-specific information. Be aware that the ARNs for some resources omit the Region, the account ID, or both the Region and the account ID.

arn:partition:service:region:account-id:resource-id
arn:partition:service:region:account-id:resource-type/resource-id
arn:partition:service:region:account-id:resource-type:resource-id
```
https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

PR #51, opened 2020-04-09
[WIP] Connected add user-Groups to add user

bexsoft:
## What does this do?

This PR connects the add user-groups API to the add user modal box

## How to Test

1. Create a new user using the add user module on mcs, keep the Inspetor opened in the network tab
2. Fill the information and save the user (remember the groups that you selected)
3. Check the network request & verify that the new user has the groups assigned in the groups request
4. Go to Groups tab & check each of the groups that you added to the user, the user must be selected in all of them


cesnietor:
I was thinking that we might need to make this a single call cause I saw that we do first `Create` and then `Set` On groups. Right now we are trying to call as few apis as possible not sure if it is ok to call two functions on create. cc @Alevsk

cesnietor:
Create and set groups worked fine using the test steps Just while trying to `edit` a user is returning me an error since we are doing a put on `api/v1/users/{user}`.
Screen Shot 2020-04-08 at 7 09 03 PM
Screen Shot 2020-04-08 at 7 08 38 PM
Should we address this issue here?

bexsoft:
Edit user is not implemented yet, that screen will change completely, that’s why this is in an independent PR

swagger.yml
```diff
@@ -239,6 +239,11 @@ paths:
           required: true
           schema:
             $ref: '#/definitions/addUserRequest'
+        - name: body
```
dvaldivia, line 242: doesn't this require some generated swagger files to be updated via `make swagger-def` ?

bexsoft: PR was WIP, Updated this file

bexsoft:
Closed it to avoid issues on files.

PR #62, opened 2020-04-09
Implemented User-Groups integration for mcs

bexsoft:
## What does this do?

This PR connects the add user-groups API to the add user modal box

## How to Test

1. Create a new user using the add user module on mcs, keep the Inspetor opened in the network tab
2. Fill the information and save the user (remember the groups that you selected)
3. Check the network request & verify that the new user has the groups assigned in the groups request
4. Go to Groups tab & check each of the groups that you added to the user, the user must be selected in all of them


swagger.yml
```diff
@@ -251,6 +251,49 @@ paths:
...
+          description: A successful response.
+          schema:
+            $ref: "#/definitions/user"
+        default:
+          description: Generic error response.
+          schema:
+            $ref: "#/definitions/error"
+      tags:
+        - AdminAPI
     delete:
```
cesnietor, line 297: I don't see this being used, is this part of the update User ?

restapi/admin_users.go
```diff
@@ -189,6 +198,35 @@ func getUserInfo(ctx context.Context, client MinioAdmin, accessKey string) (*mad
...
+		return nil, err
+	}
+
+	// create a minioClient interface implementation
+	// defining the client to be used
+	adminClient := adminClient{client: mAdmin}
+
+	user, err := getUserInfo(ctx, adminClient, params.Name)
+	if err != nil {
+		log.Println("error adding user:", err)
```
cesnietor, line 216:
```suggestion
		log.Println("error getting user:", err)
```

swagger.yml
bexsoft: This will be implemented in the feature that I'm working on right now, I added it right now since I noticed that there were conflicts in this file in other PRs

restapi/admin_users.go
```diff
@@ -116,15 +125,15 @@ func getListUsersResponse() (*models.ListUsersResponse, error) {
...
 	// Calls into MinIO to add a new user if there's an error return it
-	err := client.addUser(ctx, *accessKey, *secretKey)
-	if err != nil {
+	if err := client.addUser(ctx, *accessKey, *secretKey); err != nil {
 		return nil, err
 	}
 
-	userElem := &models.User{
-		AccessKey: *accessKey,
+	userElem, errUG := updateUserGroups(ctx, client, *accessKey, groups)
```
cesnietor, line 134: we should probably just call this function if groups is not empty, else all the code inside will be run .

PR #65, opened 2020-04-13
Removed console messages in mcs-users module

bexsoft:
## What does this do?
Removed console messages in mcs-users module, also detected a couple of non-used dependencies that needed to be removed.


portal-ui/src/screens/Console/Common/FormComponents/RadioGroupSelector/RadioGroupSelector.tsx
```diff
@@ -127,6 +127,7 @@ export const RadioGroupSelector = ({
             {selectorOptions.map(selectorOption => {
               return (
                 <FormControlLabel
+                  key={`rd-${name}-${selectorOption.value}`}
```
cesnietor, line 130: is this part of this change?

bexsoft: Yes, this was throwing an error message in the console

PR #68, opened 2020-04-14
[WIP] Added bulk functionality for add users to groups

bexsoft:
## What does this do?

Added functionality in users module to add multiple users to multiple groups at once.

## How to test

1. Go into Users module and check that the add to group is disabled when no users are selected in the list
2. Select multiple users in the list, the add to group button should get enabled
3.  Click on the add to group button, a modal box to select multiple groups should be opened
4. Select the desired groups and click on the save button, the modal box should close and the users should get unchecked
5. Check on each user & review the groups, the user must have assigned the selected groups on step 4


restapi/admin_users.go
```diff
@@ -402,3 +411,109 @@ func getUpdateUserResponse(params admin_api.UpdateUserInfoParams) (*models.User,
...
+
+	// create a minioClient interface implementation
+	// defining the client to be used
+	adminClient := adminClient{client: mAdmin}
+
+	usersList := params.Body.Users
+	groupsList := params.Body.Groups
+
+	if err := addUsersListToGroups(ctx, adminClient, usersList, groupsList); err != nil {
+		log.Println("error updating groups bulk users")
```
dvaldivia, line 514: you should print the error too so we understand what went wrong

bexsoft: Fixed

restapi/admin_users.go
```diff
@@ -402,3 +411,109 @@ func getUpdateUserResponse(params admin_api.UpdateUserInfoParams) (*models.User,
...
+
+			userToAddRemove := []string{user}
+
+			updateReturn := updateGroupMembers(ctx, client, groupToAssign, userToAddRemove, false)
+
+			chProcess <- updateReturn
+
+			return updateReturn
+		}()
+
```
cesnietor, line 431: go func doesn't need to return error, also updateGroupMembers returns error so it is easier to understand what it returns
so we should keep calling it `err` and handle it accordingly.
```suggestion
		go func() {
			defer close(chProcess)

			userToAddRemove := []string{user}

			err := updateGroupMembers(ctx, client, groupToAssign, userToAddRemove, false)

			if err != nil {
			   chProcess <- err
			   return // (this might not be needed)
			}
		}()
```

restapi/admin_users.go
```diff
@@ -402,3 +411,109 @@ func getUpdateUserResponse(params admin_api.UpdateUserInfoParams) (*models.User,
...
+		return chProcess
+	}
+
+	parallelUserRead := func(user string) chan error {
```
cesnietor, line 435: this function is more updating than doing user read, is this the right name of the function?

restapi/admin_users.go
```diff
@@ -402,3 +411,109 @@ func getUpdateUserResponse(params admin_api.UpdateUserInfoParams) (*models.User,
...
+			for _, chanRet := range groupsUpdate {
+				locError := <-chanRet
+
+				if locError != nil {
+
+					channelHasError = locError
+				}
+			}
+
+			chUserRead <- channelHasError
```
cesnietor, line 459: if we have an error we should return right away
Also it would simplify code.
```suggestion
				if chanRet != nil {
				         chUserRead <- chanRet
				         return
				}
			}
```

restapi/admin_users.go
```diff
@@ -402,3 +411,109 @@ func getUpdateUserResponse(params admin_api.UpdateUserInfoParams) (*models.User,
...
+	if len(errorsUpdates) > 0 {
+		errorMessages := ""
+
+		for _, errorItem := range errorsUpdates {
+			errorMessages = errorMessages + ", " + errorItem.Error()
+		}
+
+		errRt := errors.New(500, "there was an error updating the user-groups bulk"+errorMessages)
+		return errRt
+	}
```
cesnietor, line 492: with the changes above we won't need this since if only one failed we will know that it failed performing the whole update, which is what would matter to us. 
Also you can use fmt.Errorf e. g.`err := fmt.Errorf("user %q (id %d) not found", name, id)`

restapi/admin_users.go
```diff
@@ -402,3 +411,109 @@ func getUpdateUserResponse(params admin_api.UpdateUserInfoParams) (*models.User,
...
+
```
cesnietor, line 493: also can you make the functions smaller in code and try to refactor them ? this will make the code much cleaner and easy to read. Thanks in advance.

restapi/admin_users.go
bexsoft: Refactored function

restapi/admin_users.go
bexsoft: Refactored function

restapi/admin_users.go
bexsoft: Refactored function

restapi/admin_users.go
bexsoft: Refactored function

restapi/admin_users.go
bexsoft: Refactored function

swagger.yml
```diff
@@ -336,6 +336,25 @@ paths:
             $ref: "#/definitions/error"
       tags:
         - AdminAPI
+  /user/groups:
```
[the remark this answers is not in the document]
bexsoft: We need to define the new paths for swagger

PR #70, opened 2020-04-17
STS integration, JWT auth and Stateless MCS

Alevsk:
This commit changes the authentication mechanism between mcs and minio to an sts
(security token service) schema using the user provided credentials, previously
mcs was using master credentials. With that said in order for you to
login to MCS as an admin your user must exists first on minio and have enough
privileges to do administrative operations.

```
./mc admin user add myminio alevsk alevsk12345
```

```
cat admin.json

{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "admin:*",
        "s3:*"
      ],
      "Resource": [
        "arn:aws:s3:::*"
      ]
    }
  ]
}

./mc admin policy add myminio admin admin.json
```

```
./mc admin policy set myminio admin user=alevsk
```


restapi/user_login.go
```diff
@@ -59,28 +42,36 @@ func registerLoginHandlers(api *operations.McsAPI) {
...
+	if err != nil {
+		log.Println(err)
 		return nil, ErrInvalidCredentials
 	}
-	// if we made it here, the credentials work, generate a session
-	sessionID := sessions.GetInstance().NewSession(cfg)
-
-	return &sessionID, nil
+	// if we made it here, the minioCredentials work, generate a jwt with claims
+	jwt, err := auth.NewJWTWithClaimsForClient(&tokens, getMinIOServer())
```
Alevsk, line 58: @harshavardhana I noticed on minio/minio we are using `req.URL.Query().Encode()` https://github.com/minio/minio/blob/master/cmd/rest/client.go#L65 (essentially a URL) as the audience value for JWT, please advice if we should be doing the same here, currently I'm using `getMinIOServer()`

pkg/auth/jwt/config.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package jwt
+
+import "github.com/minio/minio/pkg/env"
+
+func GetHmacJWTSecret() string {
+	return env.Get(McsHmacJWTSecret, "C34AgmdPwoXqt91yt1Zw5fZMjMHdbPyS6QXXTJJvSl1mSzx7HmrrdqLvAG5O3n8P")
```
dvaldivia, line 22: is there a way to randomize this if it's not set?

restapi/admin_arns.go
```diff
@@ -53,8 +54,8 @@ func getArns(ctx context.Context, client MinioAdmin) (*models.ArnsResponse, erro
 }
 
 // getArnsResponse returns a list of active arns in the instance
-func getArnsResponse() (*models.ArnsResponse, error) {
-	mAdmin, err := newMAdminClient()
+func getArnsResponse(sessionID string) (*models.ArnsResponse, error) {
```
dvaldivia, line 57: This makes me think if we should introduce a context-like parameter on all funcitons to pass down both `context.Context` and session, but we can do that later.

cesnietor: yes I was thinking the same, not in context but probably we need to create an Interface for all Response functions 
and we can do just e.g.  `mcs_api.getArnsResponse() ..` and internally mcs_api.SessionID can be taken from it.

restapi/user_login.go
```diff
@@ -59,28 +42,36 @@ func registerLoginHandlers(api *operations.McsAPI) {
...
-func login(mc McCmd, accessKey, secretKey *string) (*string, error) {
-	// Probe the credentials
-	cfg, pErr := mc.BuildS3Config(getMinIOServer(), *accessKey, *secretKey, "", "auto")
-	if pErr != nil {
+// login performs a check of minioCredentials against MinIO
+func login(credentials MCSCredentials) (*string, error) {
+	// try to obtain minioCredentials,
+	tokens, err := credentials.Get()
+	if err != nil {
+		log.Println(err)
```
cesnietor, line 54: Since the top level function is already logging the error, please remove it to avoid duplicated logs.
```suggestion
```

restapi/user_login.go
```diff
@@ -59,28 +42,36 @@ func registerLoginHandlers(api *operations.McsAPI) {
...
 		return nil, ErrInvalidCredentials
 	}
-	// if we made it here, the credentials work, generate a session
-	sessionID := sessions.GetInstance().NewSession(cfg)
-
-	return &sessionID, nil
+	// if we made it here, the minioCredentials work, generate a jwt with claims
+	jwt, err := auth.NewJWTWithClaimsForClient(&tokens, getMinIOServer())
+	if err != nil {
+		log.Println(err)
```
cesnietor, line 60: Since the top level function is already logging the error, please remove it to avoid duplicated logs.
```suggestion
```

restapi/user_logout.go
```diff
@@ -38,21 +37,22 @@ func registerLogoutHandlers(api *operations.McsAPI) {
...
+	credentials.Expire()
 }
 
 // getLogoutResponse performs logout() and returns nil or error
-func getLogoutResponse(sessionID string) error {
-	if err := logout(sessionID); err != nil {
+func getLogoutResponse(jwt string) error {
+	creds, err := getMcsCredentialsFromJWT(jwt)
+	if err != nil {
+		return err
```
cesnietor, line 49: can you please add a log to see where it failed?

restapi/client.go
```diff
@@ -107,20 +109,66 @@ func (c mcS3Client) removeNotificationConfig(arn string, event string, prefix st
...
+	return credentials.NewSTSAssumeRole(getMinIOServer(), credentials.STSAssumeRoleOptions{
+		AccessKey: accessKey,
+		SecretKey: secretKey,
+		Location:  location,
+	})
+}
+
+// getMcsCredentialsFromJWT returns the *minioCredentials.Credentials associated to the
+// provided jwt, this is useful for running the Expire() or IsExpired() operations
+func getMcsCredentialsFromJWT(jwt string) (*credentials.Credentials, error) {
```
cesnietor, line 148: Can you please add tests for this function?

restapi/consts.go
```diff
@@ -46,4 +46,6 @@ const (
 	McsSecureReferrerPolicy                  = "MCS_SECURE_REFERRER_POLICY"
 	McsSecureFeaturePolicy                   = "MCS_SECURE_FEATURE_POLICY"
 	McsSecureExpectCTHeader                  = "MCS_SECURE_EXPECT_CT_HEADER"
+
+	McsHmacJWTSecret = "MCS_HMAC_JWT_SECRET"
```
cesnietor, line 50: can you please add how to use this on a README?

pkg/auth/jwt/parser.go
```diff
@@ -0,0 +1,316 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package jwt
+
+// This file is a re-implementation of the original code here with some
+// additional allocation tweaks reproduced using GODEBUG=allocfreetrace=1
+// original file https://github.com/dgrijalva/jwt-go/blob/master/parser.go
+// borrowed under MIT License https://github.com/dgrijalva/jwt-go/blob/master/LICENSE
+
```
cesnietor, line 23: can you please add tests for the functions defined in this package? SetExpiry, SetAudience, parseUnverifiedMapClaims, ParseWithClaims, etc.

cesnietor: or probably just the ones that are more than just simple assignment

Alevsk: Hi, all the functions inside `parser.go` are already tested by the `TestJWTAuthenticate` and `TestIsJWTValid` tests on `jwt_test.go`, ie `IsJWTValid` uses `JWTAuthenticate` that uses `NewMapClaims` and `ParseWithClaims`, internally uses `Valid`, the same is true for `parseUnverifiedMapClaims`, `base64Decode`, `ParseWithClaims`, `NewMapClaims`, `Valid` and `Lookup` are public functions to match the implemented functions on `jwt.StandardClaims` go.

restapi/user_login.go
harshavardhana: that is for a different reason for internode communication @Alevsk

restapi/client.go
```diff
@@ -107,20 +109,66 @@ func (c mcS3Client) removeNotificationConfig(arn string, event string, prefix st
...
+	claims, err := auth.JWTAuthenticate(jwt)
 	if err != nil {
 		return nil, err
 	}
+	creds := credentials.NewStaticV4(claims.AccessKeyID, claims.SecretAccessKey, claims.SessionToken)
```
harshavardhana, line 153: I am not sure if its a good idea to send claims in JWT they are publicly visible @Alevsk and can be used to make external communication.  You only need `claims.Subject` - are you thinking to avoid storing this information the server side?

Alevsk: Yes @harshavardhana, I want mcs to be stateless and to don't put all the eggs on the same basquet, claims are temporary credentials that can be stored on client side, similar to an access token in oauth2 (this doesn't mean claims should not be confidential, they should only be shared between client and the application), if an attacker compromise them he will gain temporary access but the same is true for the subject claim approach, however both mechanism doesn't allow you to refresh the existing credentials for that user must input his access/secret keys again.

Alevsk: I don't see any security improvement in storing claims on the server and then using them via the subject claim besides an attacker not being able to talk directly to the minio server, he will have the same level of privileges in both scenarios, talking directly to minio or through mcs and I do see a future scalability problem if mcs stores sessions in memory, maybe there's something that I'm not seeing :p, I'll like to read @aead thoughts on this.

harshavardhana: But subject claim doesn't allow access to secrets which is the key here. The problem in this approach is that state such as secret is moved to client, it is not a correct architecture because this disallows us extending the expiry.

Alevsk: I encrypted the JWT data claim :)

Alevsk:
@aead can you review the `pkg/auth/jwt.go` that contains the encrypt/decrypt functions for the jwt claims? thanks

pkg/auth/jwt.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	cipherText := gcm.Seal(nonce, nonce, plaintext, nil)
+	return cipherText, nil
+}
+
+// Decrypts a blob of data using AEAD (AES-GCM) with a pbkdf2 derived key
+func decrypt(data []byte) []byte {
+	block, err := aes.NewCipher(derivedKey)
+	if err != nil {
+		panic(err.Error())
```
harshavardhana, line 162: Avoid panics as much as needed unless you really want your entire handlers to crash @Alevsk return errors back to the client if you see a malformed request.

Alevsk: Good catch, I removed the panics from the encrypt func but forgot the decrypt, I'll change them to return errors

pkg/auth/jwt.go
```diff
@@ -0,0 +1,180 @@
...
+		log.Println(err)
+		// we return a generic error that doesn't give any information to attackers
+		return nil, errReadingToken
+	}
+	// claimsTokens contains the decrypted STS claims
+	return claimTokens, nil
+}
+
+// NewJWTWithClaimsForClient generates a new jwt with claims based on the provided STS credentials
+// NewJWTWithClaimsForClient first encrypt the clients and the sign them
```
cesnietor, line 94: which one is it? XD

Alevsk: NewJWTWithClaimsForClient do both actions XD

PR #74, opened 2020-04-21
Tables replacement in mcs

bexsoft:
## What does this do?

Replaced all the tables in mcs to be consistent with the new design

## How to test

Go into each section and check that all the information is shown correctly and the table interactions doesn't get broken

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/ListBuckets.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/ListBuckets.tsx
@@ -16,34 +16,20 @@
 
 import React from "react";
 import { createStyles, Theme, withStyles } from "@material-ui/core/styles";
-import Table from "@material-ui/core/Table";
-import TableBody from "@material-ui/core/TableBody";
-import TableCell from "@material-ui/core/TableCell";
-import TableHead from "@material-ui/core/TableHead";
-import TableRow from "@material-ui/core/TableRow";
-import Paper from "@material-ui/core/Paper";
 import Grid from "@material-ui/core/Grid";
-import api from "../../../../common/api";
-import { Bucket, BucketList } from "../types";
-import {
-  Button,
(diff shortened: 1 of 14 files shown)


cesnietor:
Not sure if part of this pr. when there is no group selected on creation of a user the group parameter needs to be sent as empty `[]`
Screen Shot 2020-04-21 at 11 31 35 AM

cesnietor:
After deletion of a user the page doesn't update:
Screen Shot 2020-04-21 at 11 32 25 AM
Screen Shot 2020-04-21 at 11 32 31 AM

cesnietor:
When a bucket is deleted an error occurs:
To reproduce this I only had one bucket and deleted it.
Screen Shot 2020-04-21 at 11 32 51 AM

cesnietor:
After deletion of a single bucket, the page doesn't load.
Screen Shot 2020-04-21 at 11 35 05 AM

bexsoft:
> Not sure if part of this pr. when there is no group selected on creation of a user the group parameter needs to be sent as empty `[]`
> Screen Shot 2020-04-21 at 11 31 35 AM

Seems like mc is now returning null when the users list is empty instead of an empty array like it used to do before. Fixed issue

bexsoft:
> After deletion of a user the page doesn't update:
> Screen Shot 2020-04-21 at 11 32 25 AM
> After delete:
> Screen Shot 2020-04-21 at 11 32 31 AM

This seems to be an old issue, it only happens when you have one only one user in the list and you want to delete it, Will file a new issue for this

bexsoft:
> When I try to edit a bucket this breaks.
> Screen Shot 2020-04-21 at 11 32 51 AM

Fixed. same issue as cannot update user

bexsoft:
> After deletion of a single bucket, the page doesn't load.
> Screen Shot 2020-04-21 at 11 35 05 AM

Fixed

PR #75, opened 2020-04-21
IDP: OpenID integration for mcs

Alevsk:
This PR adds support for oidc in  mcs, to enable idp
authentication you need to pass the following environment variables and
restart mcs.

```
MCS_IDP_URL=""
MCS_IDP_CLIENT_ID=""
MCS_IDP_SECRET=""
MCS_IDP_CALLBACK=""
```

Make sure the mcsAdmin policy exists on minio.

## Test 1

- Create and configure a new account in auth0
- Create a new application and get the application credentials (clientId, secret, url, etc) on auth0
- Set the mcs callback on auth0
- Create a new user in auth0
- Start MCS with the idp env variables
- Click on the Login button and authenticate using your auth0 account
- You should be able to access MCS_IDP_SECRET

## Test 2

- Disable the user you just created on auth0
- Click on the Login button and authenticate using your auth0 account
- You should be unable to login because the user is disabled in auth0


portal-ui/src/screens/LoginPage/LoginPage.tsx
```diff
@@ -67,49 +69,85 @@ const styles = (theme: Theme) =>
...
         "transparent linear-gradient(333deg, #281B6F 1%, #271260 13%, #120D53 83%) 0% 0% no-repeat padding-box;",
     },
-      oceanBg:{
-          backgroundImage:'url(/images/BG_Illustration.svg)',
-          backgroundRepeat:'no-repeat',
-          backgroundPosition:'bottom left',
-          height:'100%',
-          width:'100%',
-      },
+    oceanBg: {
```
dvaldivia, line 72: Are you running prettier?

pkg/auth/idp/oauth2/provider.go
```diff
@@ -0,0 +1,226 @@
...
+	// - RedirectURL is the URL to redirect users going through
+	//   the OAuth flow, after the resource owner's URLs.
+	// - Endpoint contains the resource server's token endpoint
+	//   URLs. These are constants specific to each server and are
+	//   often available via site-specific packages, such as
+	//   google.Endpoint or github.Endpoint.
+	// - Scopes specifies optional requested permissions.
+	ClientID string
+	config   Configuration
+	provider *oidc.Provider
```
cesnietor, line 92: does `Provider` struct above is a `*oidc.Provider` ? or has a `*oidc.Provider`, if it `has` it is ok like this, but if it is, we should probably define it like :
```
...
config   Configuration
*oidc.Provider
...
```
Also probably the name here `Provider` is weird since as I said seems to have a provider within. (just a bit confusing)

pkg/auth/idp/oauth2/provider.go
```diff
@@ -0,0 +1,226 @@
...
+		Endpoint:     provider.Endpoint(),
+		Scopes:       scopes,
+	}
+	client.config = &config
+	client.provider = provider
+	client.ClientID = clientID
+
+	return client, nil
+}
+
```
cesnietor, line 123: Can you please add comments on exported types.

restapi/user_login.go
```diff
@@ -81,12 +94,88 @@ func getLoginResponse(lr *models.LoginRequest) (*models.LoginResponse, error) {
...
+	redirectURL := ""
+	if oauth2.IsIdpEnabled() {
+		loginStrategy = models.LoginDetailsLoginStrategyRedirect
+		// initialize new oauth2 client
+		oauth2Client, err := oauth2.NewClient(ctx, oauth2.GetIdpURL(), oauth2.GetIdpClientID(), oauth2.GetIdpSecret(), oauth2.GetIdpCallbackURL(), nil)
+		if err != nil {
+			return nil, err
+		}
+		// Validate user against IDP
+		idpClient := &auth.IdentityProvider{Client: oauth2Client}
```
cesnietor, line 110: why is it called idpClient if it is an IdentityProvider?

restapi/user_login.go
```diff
@@ -81,12 +94,88 @@ func getLoginResponse(lr *models.LoginRequest) (*models.LoginResponse, error) {
...
+func getLoginOauth2AuthResponse(lr *models.LoginOauth2AuthRequest) (*models.LoginResponse, error) {
+	ctx := context.Background()
+	if oauth2.IsIdpEnabled() {
+		// initialize new oauth2 client
+		oauth2Client, err := oauth2.NewClient(ctx, oauth2.GetIdpURL(), oauth2.GetIdpClientID(), oauth2.GetIdpSecret(), oauth2.GetIdpCallbackURL(), nil)
+		if err != nil {
+			return nil, err
+		}
+		// initialize new identity provider
+		idpClient := &auth.IdentityProvider{Client: oauth2Client}
```
cesnietor, line 137: why is it called idpClient if it is an IdentityProvider?, we feed this in a function as a provider.

portal-ui/src/screens/LoginPage/types.ts
```diff
@@ -0,0 +1,20 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+export interface LoginDetails {
+  loginStrategy: string;
+  redirect: string;
+}
```
bexsoft, line 20: Please add an empty space at the end of the file
```suggestion
}

```

portal-ui/src/screens/LoginPage/types.ts
```diff
@@ -0,0 +1,20 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+export interface LoginDetails {
```
bexsoft, line 17: Shouldn't we call ILoginDetails to follow the convention of other interfaces?

pkg/auth/idp/oauth2/provider.go
```diff
@@ -0,0 +1,226 @@
...
+	PhoneNumber       string                 `json:"phone_number"`
+	PhoneVerified     bool                   `json:"phone_verified"`
+	Picture           string                 `json:"picture"`
+	UpdatedAt         string                 `json:"updated_at"`
+	UserID            string                 `json:"user_id"`
+	UserMetadata      map[string]interface{} `json:"user_metadata"`
+	Username          string                 `json:"username"`
+}
+
+// VerifyIdentity will contact the configured IDP and validate the user identity based on the authorization code
```
cesnietor, line 149: is this being used? I don't see it I just see the use of `IdentityProviderClient.VerifyIdentity `. but not sure if this is used except from tests.

restapi/user_login.go
```diff
@@ -81,12 +94,88 @@ func getLoginResponse(lr *models.LoginRequest) (*models.LoginResponse, error) {
...
+		adminClient := adminClient{client: mAdmin}
+		accessKey := identity.Email
+		secretKey := utils.RandomCharString(32)
+		// Create user in MinIO
+		_, err = addUser(ctx, adminClient, &accessKey, &secretKey, []string{})
+		if err != nil {
+			log.Println("error adding user:", err)
+			return nil, err
+		}
+		// assign the "mcsAdmin" policy to this user
```
cesnietor, line 157: what happens if the next functions fail after the user was created? we should delete it to make it more `atomic`, else we would have users floating around.

pkg/auth/idp/oauth2/provider.go
```diff
@@ -0,0 +1,226 @@
...
+	ClientID string
+	config   Configuration
+	provider *oidc.Provider
+}
+
+// derivedKey is the key used to compute the HMAC for signing the oauth state parameter
+// its derived using pbkdf on MCS_IDP_HMAC_PASSPHRASE with MCS_IDP_HMAC_SALT
+var derivedKey = pbkdf2.Key([]byte(getPassphraseForIdpHmac()), []byte(getSaltForIdpHmac()), 4096, 32, sha1.New)
+
+// NewClient instantiates a new oauth2 client using the configured credentials
```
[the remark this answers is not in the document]
Alevsk: I renamed it as `NewOauth2ProviderClient` to differentiate between different providers we may add in the future.

pkg/auth/idp/oauth2/provider.go
Alevsk: it has, ill rename the variable names so it's more clear.

pkg/auth/idp/oauth2/provider.go
Alevsk: VerifyIdentity is being used via the identityProvider interface on

```
func loginOauth2Auth(ctx context.Context, provider *auth.IdentityProvider, code, state string) (*oauth2.User, error) {
	userIdentity, err := provider.VerifyIdentity(ctx, code, state)
	if err != nil {
		return nil, err
	}
	return userIdentity, nil
}
```

PR #82, opened 2020-04-28
Add mcs admin trace api

cesnietor:
Trace Api uses websocket to send trace information, a
    valid jwt token needs to be sent either on the header
    or as a cookie of the ws request to start.
    Three goroutines are needed to ensure communication
    if read hearbeat fails all trace should stop by cancelling
    the context. WaitGroups are needed to ensure all
    goroutines finish gracefully.

> Note: To test this a websocket client is needed.
-You can test it by doing `/ws/trace`
-PingWait is 15 sec.


bexsoft:
Please update your branch with the latest from master

cesnietor:
> Please update your branch with the latest from master

done

restapi/admin_trace.go
```diff
@@ -0,0 +1,196 @@
...
+func startTraceInfo(conn WSConn, client MinioAdmin) (mError error) {
+	// a WaitGroup waits for a collection of goroutines to finish
+	wg := sync.WaitGroup{}
+	// a cancel context is needed to end all goroutines used
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Set number of goroutines to wait. wg.Wait()
+	// waitsuntil counter is zero (all are done)
+	wg.Add(3)
```
Alevsk, line 100: I read the PR description regarding 3 goroutines but I don't have all the context, is this need it for some type of quorum? would someone want to increase this to 5 or 7 in the future? If the answer is yes an env variable should be added for user to configure this value.

restapi/admin_trace.go
```diff
@@ -0,0 +1,196 @@
...
+				log.Println("error on serviceTrace:", traceInfo.Err)
+				ch <- traceInfo.Err
+				return
+			}
+			// Serialize message to be sent
+			traceInfoBytes, err := json.Marshal(shortTrace(&traceInfo))
+			if err != nil {
+				fmt.Println("error on json.Marshal:", err)
+				ch <- err
+				return
```
Alevsk, line 158: This is missing in the test coverage, it's possible to make json.Marshal to return an error, ie: shortTrace returns a value that cannot be converted to a JSON string?, otherwise it's fine

restapi/admin_trace.go
```diff
@@ -0,0 +1,196 @@
...
+	s.Query = t.ReqInfo.RawQuery
+	s.FuncName = t.FuncName
+	s.StatusCode = t.RespInfo.StatusCode
+	s.StatusMsg = http.StatusText(t.RespInfo.StatusCode)
+	s.CallStats.Duration = t.CallStats.Latency.String()
+	s.CallStats.Rx = t.CallStats.InputBytes
+	s.CallStats.Tx = t.CallStats.OutputBytes
+
+	if host, ok := t.ReqInfo.Headers["Host"]; ok {
+		s.Host = strings.Join(host, "")
```
Alevsk, line 191: This is missing in the test coverage, it's possible to mock the info.Trace object to simulate a request without the Host header? otherwise is fine.

restapi/ws_handle.go
```diff
@@ -0,0 +1,207 @@
...
+	ReadBufferSize:  0,
+	WriteBufferSize: 1024,
+}
+
+const (
+	// websocket base path
+	wsBasePath = "/ws"
+
+	// Time allowed to read the next pong message from the peer.
+	pingWait = 15 * time.Second
```
Alevsk, line 42: can this be configured by the user via env variables?

restapi/ws_handle.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// Maximum message size allowed from peer. 0 = unlimited
+	maxMessageSize = 512
```
Alevsk, line 45: can this be configured by the user via env variables?

restapi/admin_trace.go
cesnietor: I'll explain:
- The goroutines are fixed for this (`trace`) functionality. A goroutine [1] for listen on client heartbeat is needed. Second goroutine is needed for Writing to the socket (send the trace info), and a Third one is needed for listening checking if it crashed on reading, for instance, the client disconnected or the ping Deadline happend.

restapi/admin_trace.go
cesnietor: No, I tried, here I rely that shortTrace parses it correctly always cause it doesn't return any error, and it almost ensures that the type would always be correct cause is a struct.

restapi/admin_trace.go
cesnietor: yes, I think it is possible, let me check.

restapi/ws_handle.go
cesnietor: yes, we can definitely  but I don't see in what scenarios this would need to change. If you see any, I'll add it. I want also the not add that many "knobs" .

restapi/ws_handle.go
cesnietor: yes it can but not sure if needed, this variable is just for reading a heartbeat, regardless of the message, so far we pretty much ignore whatever the clients sends as long as it is within the pingWait. Do you see it necessary?

restapi/admin_trace.go
dvaldivia: I think you could instead wait on their respective channels to close up instead of a wait group to ensure there's no deadlock due to a go routine never calling `wg.Done`, since each goroutine already returns a channel, I think essentially you aonly need `sendTraceInfo` routine to close to mark the process as finished, and you could signal the other 2 routines to close via context, elimitating the need for a wait group

cesnietor: @dvaldivia  No, because I need any of (wsReadCheck or sendTraceInfo) to trigger a cancel context, not just sendTraceInfo.  If ReadCheck fails that should cancel sendTraceInfo and viceversa. Else either the Read or the Trace would remain open.  There is no deadlock since all return if context is cancelled.

restapi/admin_trace_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// Test-5: error happens while reading, expected Close Error CloseGoingAway
+	// 		   expected Close Error should not return an error, just return.
+	connReadMessageMock = func() (messageType int, p []byte, err error) {
+		return 0, []byte{}, &websocket.CloseError{Code: websocket.CloseGoingAway, Text: ""}
+	}
+	if err := startTraceInfo(mockWSConn, wsClientMock.madmin); err != nil {
+		t.Errorf("Failed on %s:, error occurred: %s", function, err.Error())
+	}
+
+	// Test-5: error happens while reading, non Close Error Type should be returned as
```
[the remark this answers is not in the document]
cesnietor: done

restapi/admin_trace_test.go
```diff
@@ -0,0 +1,169 @@
...
+	//         error
+	connReadMessageMock = func() (messageType int, p []byte, err error) {
+		return 0, []byte{}, fmt.Errorf("error on read")
+	}
+	if err := startTraceInfo(mockWSConn, wsClientMock.madmin); assert.Error(err) {
+		assert.Equal("error on read", err.Error())
+	}
+
+	// Test-6: error happens on serviceTrace Minio, trace should stop
```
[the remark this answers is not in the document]
cesnietor: done

PR #84, opened 2020-04-30
Admin Trace UI

dvaldivia:
Screen Shot 2020-04-29 at 8 11 48 PM


cesnietor:
the connection needs to be closed else it triggers an abnormal closure

PR #102, opened 2020-05-08
Connect MCS with Minio TLS/Custom CAs

Alevsk:
This PR adds support to connect MCS to minio instances running TLS with
self-signed certificates or  certificates signed by custom
Certificate Authorities

```
export MCS_MINIO_SERVER_TLS_ROOT_CAS=file1,file2,file3
```

Note: TLS Skip Verification is not supported unless there's a clear need
for it

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,15 @@ export MCS_MINIO_SERVER=http://localhost:9000
 ./mcs server
 ```
 
+## Connect MCS to a Minio using TLS and a self-signed certificate
+
+```
+...
+export MCS_MINIO_SERVER_TLS_SKIP_VERIFICATION=on
+export MCS_MINIO_SERVER=https://localhost:9000
+./mcs server
+```
+
 You can verify that the apis work by doing the request on `localhost:9090/api/v1/...`
 
 # Contribute to mcs Project
--- a/go.sum
+++ b/go.sum
@@ -12,6 +12,7 @@ github.com/Azure/azure-storage-blob-go v0.8.0 h1:53qhf0Oxa0nOjgbDeeYPUeyiNmafAFE
 github.com/Azure/azure-storage-blob-go v0.8.0/go.mod h1:lPI3aLPpuLTeUwh1sViKXFxwl2B6teiRqI0deQUvsw0=
 github.com/Azure/go-autorest v11.7.1+incompatible h1:M2YZIajBBVekV86x0rr1443Lc1F/Ylxb9w+5EtSyX3Q=
 github.com/Azure/go-autorest v11.7.1+incompatible/go.mod h1:r+4oMnoxhatjLLJ6zxSWATqVooLgysK6ZNox3g/xq24=
+github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/DataDog/datadog-go v2.2.0+incompatible/go.mod h1:LButxg5PwREeZtORoXG3tL4fMGNddJ+vMq1mwgfaqoQ=
 github.com/PuerkitoBio/purell v1.1.0/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
--- a/restapi/client.go
+++ b/restapi/client.go
@@ -18,6 +18,7 @@ package restapi
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -68,6 +68,15 @@ export MCS_MINIO_SERVER=http://localhost:9000
 ./mcs server
 ```
 
+## Connect MCS to a Minio using TLS and a self-signed certificate
+
+```
+...
+export MCS_MINIO_SERVER_TLS_SKIP_VERIFICATION=on
```
aead, line 75: Why is it necessary to have both concepts - custom CAs and disabling TLS X.509 certificate validation?
Is it not possible to set a self-signed certificate as custom CA / trusted cert?

This way self-signed certificates are no different from a custom CA setup

Alevsk: @aead thanks for the feedback, in the scenario of using only self-signed certificates with minio the mcs admin can use only the `export MCS_MINIO_SERVER_TLS_SKIP_VERIFICATION=on` configuration, in the scenario of using a certificate generated by a custom CA the user can just provide `export MCS_MINIO_SERVER_TLS_ROOT_CAS=rootCA.pem`

Alevsk: suggested changes applied :)

go.sum
```diff
@@ -404,10 +406,17 @@ github.com/minio/mc v0.0.0-20200415193718-68b638f2f96c/go.mod h1:l9PuOY62zT7AQJq
 github.com/minio/minio v0.0.0-20200415191640-bde0f444dbab/go.mod h1:v8oQPMMaTkjDwp5cOz1WCElA4Ik+X+0y4On+VMk0fis=
 github.com/minio/minio v0.0.0-20200501193630-d1c8e9f31ba0 h1:QxIz36O01LbKqJiz6HKeKCOC2afgydspkpahQ807msY=
 github.com/minio/minio v0.0.0-20200501193630-d1c8e9f31ba0/go.mod h1:Vhlqz7Se0EgpgFiVxpvzF4Zz/h2LMx+EPKH96Aera5U=
+github.com/minio/minio v0.0.0-20200506212505-4c9de098b0fb h1:okO1FO5iR5G9ZX8vElpeAgdqYue98UzQ3HQ5X5/pB2g=
```
cesnietor, line 409: did you run `go mod tidy?`

restapi/client.go
```diff
@@ -133,13 +134,45 @@ func (c mcsCredentials) Expire() {
...
+	return s.stsAssumeRole.IsExpired()
+}
+
+// STSClient contains http.client configuration need it by STSAssumeRole
+var STSClient = PrepareSTSClient()
+
 func newMcsCredentials(accessKey, secretKey, location string) (*credentials.Credentials, error) {
-	return credentials.NewSTSAssumeRole(getMinIOServer(), credentials.STSAssumeRoleOptions{
+	stsEndpoint := getMinIOServer()
+	if stsEndpoint == "" {
```
cesnietor, line 157:
```suggestion
	if string.TrimSpace(stsEndpoint) == "" {
```
else `" "` will not enter here, same for the ones below

restapi/config.go
```diff
@@ -51,6 +51,16 @@ func getMinIOServer() string {
 	return env.Get(McsMinIOServer, "http://localhost:9000")
 }
 
+// If MCS_MINIO_SERVER_TLS_ROOT_CAS is true mcs will load a list of certificates into the
+// http.client rootCAs store, this is useful for testing or when working with self-signed certificates
+func getMinioServerTLSRootCAs() []string {
+	caCertFileNames := env.Get(McsMinIOServerTLSRootCAs, "")
+	if caCertFileNames == "" {
```
cesnietor, line 58:
```suggestion
	if string.TrimSpace(caCertFileNames) == "" {
```

restapi/config.go
```diff
@@ -67,7 +77,7 @@ func getMinIOEndpointIsSecure() bool {
 	if strings.Contains(server, "://") {
 		parts := strings.Split(server, "://")
 		if len(parts) > 1 {
```
cesnietor, line 79: here it will return false if server is just like `https://` is that ok?  else it should be `> 0`

restapi/utils.go
```diff
@@ -54,3 +56,12 @@ func UniqueKeys(a []string) []string {
 	}
 	return list
 }
+
+// Verify if a file exist on the desired location and its not a folder
```
cesnietor, line 60: Suggestion: comment on exported function FileExists should be of the form "FileExists ..." (with optional leading article)

restapi/config.go
Alevsk: this is fine, `https://` is not a valid address

restapi/client.go
cesnietor: how is this solved?

restapi/config.go
cesnietor: how is this solved?

PR #107, opened 2020-05-08
wss for websockets on tls and single page application behavior

dvaldivia:
Detects `https` and sets the Websocket protocol to `wss`
on asset not found we return `index.html` so the react application can handle 404

### DIFF
--- a/portal-ui/src/screens/Console/Logs/Logs.tsx
+++ b/portal-ui/src/screens/Console/Logs/Logs.tsx
@@ -24,6 +24,7 @@ import { createStyles, Theme, withStyles } from "@material-ui/core/styles";
 import { niceBytes } from "../../../common/utils";
 import Ansi from "ansi-to-react";
 import { isNull, isNullOrUndefined } from "util";
+import { wsProtocol } from "../../../utils/wsUtils";
 
 const styles = (theme: Theme) =>
   createStyles({
@@ -79,7 +80,11 @@ const Logs = ({
     const isDev = process.env.NODE_ENV === "development";
     const port = isDev ? "9090" : url.port;
 
-    const c = new W3CWebSocket(`ws://${url.hostname}:${port}/ws/console`);
+    const wsProt = wsProtocol(url.protocol);
+
+    const c = new W3CWebSocket(
+      `${wsProt}://${url.hostname}:${port}/ws/console`
+    );
 
     let interval: any | null = null;
     if (c !== null) {
--- a/portal-ui/src/screens/Console/Trace/Trace.tsx
+++ b/portal-ui/src/screens/Console/Trace/Trace.tsx
@@ -22,6 +22,7 @@ import { traceMessageReceived, traceResetMessages } from "./actions";
 import { TraceMessage } from "./types";
 import { createStyles, Theme, withStyles } from "@material-ui/core/styles";
 import { niceBytes } from "../../../common/utils";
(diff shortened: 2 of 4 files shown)


portal-ui/src/screens/Console/Logs/Logs.tsx
```diff
@@ -79,7 +79,13 @@ const Logs = ({
...
     const port = isDev ? "9090" : url.port;
 
-    const c = new W3CWebSocket(`ws://${url.hostname}:${port}/ws/console`);
+    let wsProtocol = "ws";
+    if (url.protocol == "https:") {
+      wsProtocol = "wss";
+    }
+    const c = new W3CWebSocket(
+      `${wsProtocol}://${url.hostname}:${port}/ws/console`
+    );
```
Alevsk, line 88: This is almost duplicated code with `Trace.tsx`, can we put this on the `utils.ts` function?

portal-ui/src/screens/Console/Trace/Trace.tsx
```diff
@@ -61,7 +61,14 @@ const Trace = ({
     const isDev = process.env.NODE_ENV === "development";
     const port = isDev ? "9090" : url.port;
 
-    const c = new W3CWebSocket(`ws://${url.hostname}:${port}/ws/trace`);
+    let wsProtocol = "ws";
+    if (url.protocol == "https:") {
+      wsProtocol = "wss";
+    }
```
cesnietor, line 67: we can probably put this in a function if all ws will be using it

restapi/configure_mcs.go
```diff
@@ -167,11 +167,44 @@ func FileServerMiddleware(next http.Handler) http.Handler {
...
 				AssetInfo: portalUI.AssetInfo,
-				Prefix:    "build"}).ServeHTTP(w, r)
+				Prefix:    "build"}
+			//assets.Open("index.html")
+			wrapHandlerSinglePageApplication(http.FileServer(&assets)).ServeHTTP(w, r)
+
 		}
 	})
 }
+
```
cesnietor, line 181: does it need to be an exported struct? if so, all exported need to have comments on top of it.
```
comment on exported type NotFoundRedirectRespWr should be of the form "NotFoundRedirectRespWr ..." (with optional leading article)
```

restapi/configure_mcs.go
```diff
@@ -167,11 +167,44 @@ func FileServerMiddleware(next http.Handler) http.Handler {
...
+	if status != http.StatusNotFound {
+		w.ResponseWriter.WriteHeader(status)
+	}
+}
+
+func (w *NotFoundRedirectRespWr) Write(p []byte) (int, error) {
+	if w.status != http.StatusNotFound {
+		return w.ResponseWriter.Write(p)
+	}
+	return len(p), nil // Lie that we successfully written it
```
cesnietor, line 198:
```suggestion
	return len(p), nil // Lie that we successfully wrote it
```

restapi/configure_mcs.go
```diff
@@ -167,11 +167,44 @@ func FileServerMiddleware(next http.Handler) http.Handler {
...
+}
+
```
cesnietor, line 200: please add comments

restapi/configure_mcs.go
cesnietor: how is this addressed?

dvaldivia: unexported

PR #113, opened 2020-05-12
Changed modal forms to contain scrollbars

### LINKED ISSUE #110: Modal box content needs to be in scrollable container
All modal boxes containers need to have a scroll bar, this way the title and the buttons will not disappear when long forms are set in the modals.

bexsoft:
fixes #110 

## What does this do?
Changed modal forms to contain scrollbars where the inputs are, so this way the titles and buttons get always visible.

## How to test
Go into long forms and check that all of them contain a scrollbar in the inputs section. Title & buttons must be always visible
You can shrink the window to check that the behavior also happens with small screens

Screen Shot 2020-05-11 at 11 57 18 PM


cesnietor:
tested, LGTM.

PR #118, opened 2020-05-12
remove color dependency and any non ascii characters

cesnietor:
Since we shouldn't rely on color being sent from the server, colors and non ascii characters coming from server will be trimmed.

Only red color will be set on errors.

Screen Shot 2020-05-12 at 1 34 06 PM


harshavardhana:
As a correct behavior, MinIO will remove the coloring from console messages from `mc admin console`

PR #120, opened 2020-05-13
Add Watch api and UI integration

cesnietor:
Uses a similar approach to Trace and Console Logs by using
websockets. It also includes the integration with the UI which
needs 3 input fields that are sent as query parameters.

Screen Shot 2020-05-14 at 10 13 22 AM

It includes: temporary dependencies
> Wait until a new release of mc comes which needs https://github.com/minio/mc/commit/a6d82862cd0206dd3574a24b072fc2995c027d28

To Test:
- Go into the new window for `Watch`.
- A bucket list should be rendered on the `Select` button.
- Prefix and suffix are optional, if not defined, watch will be done on all prefixes and/or suffixes.
- Once Start is clicked, it should start listening.
- Copy a file into the bucket that you selected (using mc) and only the events that match your choices
should appear on mcs.
- If you go to other window, watch should stop

### DIFF
--- a/portal-ui/src/common/utils.ts
+++ b/portal-ui/src/common/utils.ts
@@ -38,3 +38,12 @@ export const setCookie = (name: string, val: string) => {
   document.cookie =
     name + "=" + value + "; expires=" + date.toUTCString() + "; path=/";
 };
+
+// timeFromdate gets time string from date input
+export const timeFromDate = (d: Date) => {
+  let h = d.getHours() < 10 ? `0${d.getHours()}` : `${d.getHours()}`;
+  let m = d.getMinutes() < 10 ? `0${d.getMinutes()}` : `${d.getMinutes()}`;
+  let s = d.getSeconds() < 10 ? `0${d.getSeconds()}` : `${d.getSeconds()}`;
+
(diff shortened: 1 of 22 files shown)


cesnietor:
position of Watch button changed to a non admin section since it is not required to be an admin to perform watch operation.

portal-ui/src/common/utils.ts
```diff
@@ -38,3 +38,12 @@ export const setCookie = (name: string, val: string) => {
...
 };
+
+// timeFromdate gets time string from date input
+export const timeFromDate = (d: Date) => {
+  let h = d.getHours() < 10 ? `0${d.getHours()}` : `${d.getHours()}`;
+  let m = d.getMinutes() < 10 ? `0${d.getMinutes()}` : `${d.getMinutes()}`;
+  let s = d.getSeconds() < 10 ? `0${d.getSeconds()}` : `${d.getSeconds()}`;
+
+  return `${h}:${m}:${s}:${d.getMilliseconds()}`;
+};
```
bexsoft, line 49: Please add an space at the end of the line
```suggestion
};

```

cesnietor: done

PR #122, opened 2020-05-14
Add size info to bucket list api

cesnietor:
Using madmin.AccountUsageInfo since that api
includes size already.
Also includes integration with UI.

Screen Shot 2020-05-14 at 2 13 34 PM

Solves: https://github.com/minio/mcs/issues/89


dvaldivia:
Can you attach the information to the Bucket Details screen?

Screen Shot 2020-05-14 at 2 27 39 PM
Screen Shot 2020-05-14 at 2 27 32 PM

cesnietor:
> Have you looked at the AccountingUsageInfo API here ?

@harshavardhana  I just checked it but it is pretty much giving the same stuff that I need, what's the added value?

cesnietor:
> Add bucket usage to bucket details screen
@dvaldivia  that's a different api, I'll send that in a separate pr for sake of simplicity

harshavardhana:
> > Have you looked at the AccountingUsageInfo API here ?
> 
> @harshavardhana  I just checked it but it is pretty much giving the same stuff that I need, what's the added value? 

It already has bucket information, it also shows if the bucket has read, write or read write access

harshavardhana:
> > > Have you looked at the AccountingUsageInfo API here ?
> > 
> > @harshavardhana  I just checked it but it is pretty much giving the same stuff that I need, what's the added value? 
> 
> It already has bucket information, it also shows if the bucket has read, write or read write access 

But it's per account basis ..

cesnietor:
> > > Have you looked at the AccountingUsageInfo API here ?
> > 
> > 
> > @harshavardhana  I just checked it but it is pretty much giving the same stuff that I need, what's the added value?
> 
> It already has bucket information, it also shows if the bucket has read, write or read write access

Ok, I'll take a look, I just want to make sure that if the crawl hasn't happened, it is still able to list the buckets, if this, provides that then I can make only one call.

cesnietor:
> > > > Have you looked at the AccountingUsageInfo API here ?
> > > 
> > > 
> > > @harshavardhana  I just checked it but it is pretty much giving the same stuff that I need, what's the added value?
> > 
> > 
> > It already has bucket information, it also shows if the bucket has read, write or read write access
> 
> Ok, I'll take a look, I just want to make sure that if the crawl hasn't happened, it is still able to list the buckets, if this provides that then I can make only one call.

@harshavardhana  So based on our discussions what do you think about starting with this and continue with the enhancements as we go?

cesnietor:
There are some changes coming in that we can leverage from: https://github.com/minio/minio/pull/9607

cesnietor:
Setting it as WIP.

cesnietor:
Using now madmin.AccountUsageInfo

cesnietor:
> Have you looked at the AccountingUsageInfo API here ?

I've changed the code to make use of it.

restapi/user_buckets.go
```diff
@@ -82,43 +82,39 @@ func registerBucketsHandlers(api *operations.McsAPI) {
...
+	info, err := client.accountUsageInfo(ctx)
 	if err != nil {
 		return []*models.Bucket{}, err
 	}
-
 	var bucketInfos []*models.Bucket
-	for _, bucket := range buckets {
-		bucketElem := &models.Bucket{Name: swag.String(bucket.Name), CreationDate: bucket.CreationDate.String()}
+	for _, bucket := range info.Buckets {
+		bucketElem := &models.Bucket{Name: swag.String(bucket.Name), CreationDate: bucket.Created.String(), Size: int64(bucket.Size)}
```
harshavardhana, line 93: do you want to just use Go's default for creation date? isn't a more humanized form preferred?

cesnietor: we handle it on UI, we just return it as it is for now.

PR #123, opened 2020-05-15
ACL for mcs

Alevsk:
This PR sets the initial version of the ACL for mcs, the idea behind
this is to start using the principle of least privileges when assigning
policies to users, currently mcsAdmin policy uses admin:* and s3:*, we
are assuming users have access to everything which is not good.

We need to start validating explicitly if users has acccess to an
specific endpoint based on IAM policy actions.

In this first version every endpoint (you can see it as a page to),
defines a set of well defined admin/s3 actions to work properly, ie:

```
// corresponds to /groups endpoint used by the groups page
var groupsActionSet = iampolicy.NewActionSet(
    iampolicy.ListGroupsAdminAction,
    iampolicy.AddUserToGroupAdminAction,
    //iampolicy.GetGroupAdminAction,
    iampolicy.EnableGroupAdminAction,
    iampolicy.DisableGroupAdminAction,
)

// corresponds to /policies endpoint used by the policies page
var iamPoliciesActionSet = iampolicy.NewActionSet(
    iampolicy.GetPolicyAdminAction,
    iampolicy.DeletePolicyAdminAction,
    iampolicy.CreatePolicyAdminAction,
    iampolicy.AttachPolicyAdminAction,
    iampolicy.ListUserPoliciesAdminAction,
)
```
With that said, for this initial version, now the sessions endpoint will
return a list of authorized pages to be render on the UI, on subsequent
prs we will add this verification of authorization via a server
middleware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,31 +14,30 @@ $ mc admin user add myminio mcs YOURMCSSECRET
 $ set -o history
 ```
 
-2. Create a policy for `mcs`
+2. Create a policy for `mcs` with access to everything (for testing and debugging)
 
 ```
 $ cat > mcsAdmin.json << EOF
 {
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Action": [
-        "admin:*"
-      ],
-      "Effect": "Allow",
-      "Sid": ""
-    },
-    {
-      "Action": [
-        "s3:*"
-      ],
-      "Effect": "Allow",
-      "Resource": [
-        "arn:aws:s3:::*"
-      ],
-      "Sid": ""
-    }
-  ]
+	"Version": "2012-10-17",
+	"Statement": [{
+			"Action": [
+				"admin:*"
+			],
+			"Effect": "Allow",
+			"Sid": ""
+		},
+		{
+			"Action": [
+                "s3:*"
+			],
+			"Effect": "Allow",
+			"Resource": [
+				"arn:aws:s3:::*"
+			],
+			"Sid": ""
+		}
+	]
 }
 EOF
 $ mc admin policy add myminio mcsAdmin mcsAdmin.json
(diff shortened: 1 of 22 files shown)


bexsoft:
Please run `make assets`

restapi/user_login.go
```diff
@@ -113,7 +116,22 @@ func getLoginResponse(lr *models.LoginRequest) (*models.LoginResponse, error) {
 		return nil, errInvalidCredentials
 	}
 	credentials := mcsCredentials{minioCredentials: creds}
-	sessionID, err := login(credentials)
+	// obtain the current policy assigned to this user
+	// necessary for generating the list of allowed endpoints
+	userInfo, err := adminClient.getUserInfo(ctx, "alevsk")
```
dvaldivia, line 121: Please remove hardcoded user name

pkg/acl/endpoints.go
```diff
@@ -0,0 +1,268 @@
...
+//				],
+//				...
+//			}
+//		]
+//	}
+// Will produce an array like: ["admin:ServerInfo", "admin:CreatePolicy", "admin:GetUser", "s3:ListenBucketNotification", "s3:PutBucketNotification"]\
+func GetActionsStringFromPolicy(policy *iampolicy.Policy) []string {
+	var actions []string
+	for _, statement := range policy.Statements {
+		// We only care about allowed actions
```
cesnietor, line 223: Regarding this, I think you can set a policy with two almost same statements and one with allow and other with deny and deny should win over it. Here I'm not sure what it will return.

cesnietor:
I'm getting this error.
Screen Shot 2020-05-15 at 2 12 38 PM

dvaldivia:
This policy causes the reducer to panic on null returned by the server, avoid NPE

```json
{
    "name": "mcsTestUserAddOnly",
    "Statement": [
        {
            "Action": [
                "admin:CreateUser"
            ],
            "Effect": "Allow",
            "Resource": [
                "arn:aws:s3:::*"
            ]
        }
    ],
    "version": "2012-10-17"
}
```

Screen Shot 2020-05-15 at 2 49 48 PM

pkg/acl/endpoints.go
Alevsk: The parser will honor everything that appears under the allow statement, meaning if you policy has all the actions required by a particular endpoint mcs will tell you its find for your UI to render that page, mcs will not compute any allow vs deny rules. If a user creates the same action in both allow and deny statement then minio will have the last word and decides if the incoming request from mcs should be approved or denied.

Alevsk:
> This policy causes the reducer to panic on null returned by the server, avoid NPE
> 
> ```json
> {
>     "name": "mcsTestUserAddOnly",
>     "Statement": [
>         {
>             "Action": [
>                 "admin:CreateUser"
>             ],
>             "Effect": "Allow",
>             "Resource": [
>                 "arn:aws:s3:::*"
>             ]
>         }
>     ],
>     "version": "2012-10-17"
> }
> ```
> 
> Screen Shot 2020-05-15 at 2 49 48 PM

fixed

dvaldivia:
Can you remove menu groups if not a single options is available? 

Screen Shot 2020-05-15 at 4 05 57 PM

cesnietor:
Just some comments, probably here we shouldn't show that Dashboard window also.
Or remove that `Access denied`
Screen Shot 2020-05-15 at 5 21 39 PM

cesnietor:
Can we remove this logs? or are they needed?
Screen Shot 2020-05-15 at 5 22 33 PM

Alevsk:
Screen Shot 2020-05-18 at 1 25 35 PM
@cesnietor @dvaldivia

README.md
```diff
@@ -50,6 +49,50 @@ $ mc admin policy add myminio mcsAdmin mcsAdmin.json
 $ mc admin policy set myminio mcsAdmin user=mcs
 ```
 
+
+### Note
+If you want to create a policy for `mcs` with limited privileges, ie `with access only to dashboard, buckets, notifications
```
dvaldivia, line 54: I think this note is nice but might confuse the users among the difference of the `mcsUser` and a regular `user`

bexsoft: I think we can change it for:

```suggestion
Additionally, you can create policies to limit the privileges for `mcs` users, for example, if you want the user to only have access to dashboard, buckets, notifications and watch page, the policy should look like this:
```

Alevsk:
> Please rebase

i don't see any new changes

PR #127, opened 2020-05-19
Fix websocket apis

cesnietor:
Remove ping check and instead use a context that will be canceled
if it the client sends a close message or an error occurs on reading.
The context will be used to cancel all functions using it.

### DIFF
--- a/restapi/admin_console.go
+++ b/restapi/admin_console.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"log"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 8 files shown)


restapi/admin_console.go
```diff
@@ -32,97 +31,38 @@ import (
...
+	// name of node, default = "" (all)
+	node := ""
+	// number of log lines
+	lineCount := 100
+	// type of logs "minio"|"application"|"all" default = "all"
+	logKind := "all"
+	// Start listening on all Console Log activity.
+	logCh := client.getLogs(ctx, node, lineCount, logKind)
+
+	for logInfo := range logCh {
```
harshavardhana, line 45: the expected pattern here is 

```
for {
select {
case <-ctx.Done():
case logInfo := <-logCh:
}
```

restapi/admin_trace.go
```diff
@@ -50,93 +49,33 @@ type callStats struct {
...
-			mError = err
+func startTraceInfo(ctx context.Context, conn WSConn, client MinioAdmin) error {
+	// trace all traffic
+	allTraffic := true
+	// Trace failed requests only
+	errOnly := false
+	// Start listening on all trace activity.
+	traceCh := client.serviceTrace(ctx, allTraffic, errOnly)
+
+	for traceInfo := range traceCh {
```
harshavardhana, line 60: for {
select {
case <-ctx.Done():
case traceInfo := <-traceCh:
}

restapi/user_watch.go
```diff
@@ -35,106 +34,48 @@ type watchOptions struct {
...
-	}(&wg)
-
-	if err := <-ch; err != nil {
-		mError = err
+func startWatch(ctx context.Context, conn WSConn, wsc MCS3Client, options watchOptions) error {
+	wo, pErr := wsc.watch(options.WatchOptions)
+	if pErr != nil {
+		fmt.Println("error initializing watch:", pErr.Cause)
+		return pErr.Cause
 	}
```
harshavardhana, line 42: you can add `defer wo.Close()` here

restapi/user_watch.go
```diff
@@ -35,106 +34,48 @@ type watchOptions struct {
...
-					err = conn.writeMessage(websocket.TextMessage, bytes)
-					if err != nil {
-						log.Println("error writeMessage:", err)
-						ch <- err
-						return
-					}
+	for {
+		select {
+		case <-ctx.Done():
+			close(wo.DoneChan)
```
harshavardhana, line 46: don't do this, instead do `defer wo.Close()`

restapi/user_watch.go
```diff
@@ -35,106 +34,48 @@ type watchOptions struct {
...
+		case events, ok := <-wo.Events():
+			// zero value returned because the channel is closed and empty
+			if !ok {
+				return nil
+			}
+			for _, event := range events {
+				// Serialize message to be sent
+				bytes, err := json.Marshal(event)
+				if err != nil {
+					fmt.Println("error on json.Marshal:", err)
```
harshavardhana, line 57: do you prefer log.Println or `fmt.Println?

restapi/admin_console.go
cesnietor: done

restapi/admin_trace.go
cesnietor: done

restapi/user_watch.go
cesnietor: keeping it like that.

restapi/user_watch.go
cesnietor: keeping it like that.

restapi/user_watch.go
cesnietor: used log instead

PR #141, opened 2020-05-22
Added loaders to bucket information block

bexsoft:
## What does this do?

Added loaders to bucket information block

Screen Shot 2020-05-22 at 4 56 40 PM


Alevsk:
LGTM

PR #156, opened 2020-06-03
Integrate mkube storageclass api with UI

cesnietor:
Open mcs with flag `MCS_OPERATOR_ONLY=on` then when adding a new Tenant, the storage classes listed should be the same as if you were doing `kubectl get storageclasses.storage.k8s.io `


restapi/mkube.go
```diff
@@ -30,6 +30,7 @@ import (
 // serverMkube handles calls for mkube
 func serverMkube(client *http.Client, w http.ResponseWriter, req *http.Request) {
 	// destination of the request, the mkube server
+	req.URL.Path = fmt.Sprintf("/api/v1/%s", strings.TrimPrefix(req.URL.Path, "/api/v1/mkube"))
```
dvaldivia, line 33: perhaps, you only need to remove the `/mkube/` with `/` that way you don't have the `/v1/` in multiple places

cesnietor: yes but they can potentially be different one is hitting mkube apis and the other one is mcs api.

cesnietor: done

PR #162, opened 2020-06-04
Tenant Details

dvaldivia:
Screen Shot 2020-06-04 at 11 40 49 AM

### DIFF
--- a/portal-ui/src/screens/Console/Tenants/TenantDetails/AddZoneModal.tsx
+++ b/portal-ui/src/screens/Console/Tenants/TenantDetails/AddZoneModal.tsx
@@ -5,13 +5,19 @@ import { modalBasic } from "../../Common/FormComponents/common/styleLibrary";
 import InputBoxWrapper from "../../Common/FormComponents/InputBoxWrapper/InputBoxWrapper";
 import SelectWrapper from "../../Common/FormComponents/SelectWrapper/SelectWrapper";
 import Grid from "@material-ui/core/Grid";
-import { factorForDropdown, getTotalSize } from "../../../../common/utils";
+import {
+  factorForDropdown,
+  getTotalSize,
+  niceBytes,
+} from "../../../../common/utils";
 import { Button, LinearProgress } from "@material-ui/core";
 
 interface IAddZoneProps {
   classes: any;
   open: boolean;
   onCloseZoneAndReload: (shouldReload: boolean) => void;
+  volumesPerInstance: number;
+  volumeSize: number;
 }
 
 const styles = (theme: Theme) =>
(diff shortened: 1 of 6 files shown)


dvaldivia:
@bexsoft it should open o_O I actually updated the Add Zones component and tested it

portal-ui/src/screens/Console/Tenants/TenantDetails/TenantDetails.tsx
```diff
@@ -106,12 +109,16 @@ const TenantDetails = ({ classes, match }: ITenantDetailsProps) => {
...
+  const [zones, setZones] = useState<IZone[]>([]);
   const [instances, setInstances] = useState<number>(0);
-  const [drives, setDrives] = useState<number>(0);
+  const [volumes, setVolumes] = useState<number>(0);
   const [addZoneOpen, setAddZone] = useState<boolean>(false);
   const [addBucketOpen, setAddBucketOpen] = useState<boolean>(false);
   const [addReplicationOpen, setAddReplicationOpen] = useState<boolean>(false);
+  const [loading, setLoading] = useState<boolean>(false);
+  const [error, setError] = useState<string>("");
+  const [tenant, setTenant] = useState<ITenant | null>(null);
```
cesnietor, line 121: after some number of useState I think we should use redux cause else it gets really nasty. What are your comments here @bexsoft ?

dvaldivia: I think it's fine since we dont want to persist this states when the user leaves the screen, this states are meant to store the result of the API call

bexsoft: @cesnietor I think it's ok to continue using local states for this unless we want to use these states in other parts of the application, what we may do is to have only the data lists, loaders ans errors in redux, this could help us to improve the performance as we may do less calls to the APIs

cesnietor: got it, cause yeah too many `useState`.

portal-ui/src/screens/Console/Tenants/TenantDetails/TenantDetails.tsx
```diff
@@ -133,12 +140,50 @@ const TenantDetails = ({ classes, match }: ITenantDetailsProps) => {
...
+    api
+      .invoke("GET", `/api/v1/mkube/tenants/${tenantName}`)
+      .then((res: ITenant) => {
+        const total = res.volume_count * res.volume_size;
+
+        setCapacity(total);
+        setZoneCount(res.zone_count);
+        setVolumes(res.volume_count);
+        setInstances(res.instance_count);
+        const resZones = res.zones !== null ? res.zones : [];
```
bexsoft, line 157: When there are no zones, res.zones is `undefined` in res variable, hence the object is not iterable and the zones doesn't get applied. Then with the validation of zones as null the modal box doesn't open

```suggestion
        const resZones = !res.zones ? [] : res.zones;
```

portal-ui/src/screens/Console/Tenants/TenantDetails/AddZoneModal.tsx
```diff
@@ -60,17 +66,15 @@ const AddZoneModal = ({
...
   const [zoneName, setZoneName] = useState<string>("");
   const [numberOfInstances, setNumberOfInstances] = useState<number>(0);
-  const [volumesPerInstance, setVolumesPerInstance] = useState<number>(0);
-  const [sizeFactor, setSizeFactor] = useState<string>("GiB");
-  const [volumeConfiguration, setVolumeConfig] = useState<string>("");
-  const [storageClass, setStorageClass] = useState<string>("");
 
   const instanceCapacity: number =
-    parseFloat(volumeConfiguration) * volumesPerInstance;
+    volumesPerInstance * volumeSize * volumesPerInstance;
```
bexsoft, line 77: Is it ok to duplicate `volumesPerInstance`? 
```suggestion
    volumesPerInstance * volumeSize;
```

PR #174, opened 2020-06-16
Added Namespace field to create tenant form.

bexsoft:
## What does this do?

Added Namespace field to create tenant form, also refactored a function to filter zones


bexsoft:
integrated in #179

PR #185, opened 2020-06-30
Fix login and logout flow for MCS

Alevsk:
fixes: https://github.com/minio/mcs/issues/184

There was a bug in Safari in related to the browser not setting the session token
correctly in localstorage, this was because we were using
window.location.href for redirect instead of history.push after login, the redirect execution was faster
was faster that the promise function getting the response after the login request
and it seems to be that Safari will kill all current request of a
window when the page is getting redirected.

Test this:

Try to sign-in using Safari browser (latest version is recommended)

### DIFF
--- a/portal-ui/src/screens/LoginPage/LoginPage.tsx
+++ b/portal-ui/src/screens/LoginPage/LoginPage.tsx
@@ -28,7 +28,8 @@ import { SystemState } from "../../types";
 import { userLoggedIn } from "../../actions";
 import api from "../../common/api";
 import { ILoginDetails, loginStrategyType } from "./types";
-import { setCookie } from "../../common/utils";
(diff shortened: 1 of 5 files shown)


portal-ui/src/screens/LoginPage/LoginPage.tsx
```diff
@@ -164,9 +165,7 @@ const Login = ({ classes, userLoggedIn }: ILoginProps) => {
       .then(() => {
         // We set the state in redux
         userLoggedIn(true);
-        // There is a browser cache issue if we change the policy associated to an account and then logout and history.push("/") after login
```
dvaldivia, line 167: What about this dire warning about `history.push()`?

PR #187, opened 2020-07-01
Merge M3 into MCS

dvaldivia:
Merges `github.com/minio/m3` into the console server


harshavardhana:
also can m3 become read-only now? @dvaldivia

dvaldivia:
> also can m3 become read-only now? @dvaldivia 

Sure!

PR #191, opened 2020-07-08
Fix Session validation for MCS Operator Mode

dvaldivia:
Fixes the validation on the Session token for when we are running in `MCS_OPERATOR_MODE`

also `MCS_MKUBE_OPERATOR_ONLY` was renamed to `MCS_OPERATOR_MODE`


restapi/client.go
```diff
@@ -168,7 +168,7 @@ func newMcsCredentials(accessKey, secretKey, location string) (*credentials.Cred
 	// Future authentication methods can be added under this switch statement
 	switch {
 	// MKUBE authentication for MCS
```
cesnietor, line 170: please change comments

pkg/acl/config.go
```diff
@@ -22,8 +22,8 @@ import (
 	"github.com/minio/minio/pkg/env"
 )
 
-// GetOperatorOnly gets MCS mkube admin mode status set on env variable
+// GetOperatorMode gets MCS mkube admin mode status set on env variable
```
cesnietor, line 25: please remove mkube from comments

dvaldivia: done

cesnietor:
please update k8s/base env variable

cesnietor:
also this env variable needs to be documented

dvaldivia:
@cesnietor this is a small change that unblocks @bexsoft , I'll update the env but since we no longer have dependencies on mkube I'll send a separate PR removing those dependencies, let's not add more changes to this PR

restapi/client.go
dvaldivia: all mkube dependencies will be removed on separare PR

dvaldivia:
@cesnietor it's trying to perform a reuqest to m3, which I wanted to address on separate PR, but then I just send the change, bare in mind if you are testing the login using `kubectl proxy` all requests will succeed, so test the login while running this inside the cluster

cluster/config.go
```diff
@@ -34,7 +34,7 @@ var (
 	errCantDetermineMCImage    = errors.New("can't determine MC Image")
 )
 
-func getK8sAPIServer() string {
+func GetK8sAPIServer() string {
 	// if m3 is running inside a k8s pod KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT will contain the k8s api server apiServerAddress
 	// if m3 is not running inside k8s by default will look for the k8s api server on localhost:8001 (kubectl proxy)
```
Alevsk, line 39:
```suggestion
	// if mcs is running inside a k8s pod KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT will contain the k8s api server apiServerAddress
	// if mcs is not running inside k8s by default will look for the k8s api server on localhost:8001 (kubectl proxy)
```

cluster/config.go
```diff
@@ -54,7 +54,7 @@ func getK8sAPIServerInsecure() bool {
 	return strings.ToLower(env.Get(m3k8SAPIServerInsecure, "off")) == "on"
 }
```
Alevsk, line 55: TODO for me:

instead of using insecure flag in `rest.TLSClientConfig{Insecure: true}` its better to load the self-signed cert in the rootCA pool of the http client used to contact to k8s apiserver

cesnietor:
Screen Shot 2020-07-08 at 1 13 17 PM
this is happening:

```
Request URL: http://localhost:9090/api/v1/login/mkube
Request Method: POST
Status Code: 404 Not Found
Remote Address: [::1]:9090
Referrer Policy: no-referrer-when-downgrade
```

cesnietor:
please make one single commit

dvaldivia:
@cesnietor done

cesnietor:
please make sure to edit the comments before squashing.

PR #193, opened 2020-07-09
Refactor for session management

Alevsk:
Previously every Handler function was receiving the session token in the
form of a jwt string, in consequence every time we want to access the
encrypted claims of the jwt we needed to run a decryption process,
additionally we were decrypting the jwt twice, first at the session
validation then inside each handler function, this was also causing a
lot of using related to the merge between m3 and mcs

What changed:

Now we validate and decrypt the jwt once in `configure_mcs.go`, this
works for both, mcs (console) and operator sessions, and then pass the
decrypted claims to all the functions that need it, so no further token
validation or decryption is need it.


docs/mcs_operator_mode.md
```diff
@@ -0,0 +1,39 @@
...
+more info here: https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#directly-accessing-the-rest-api
+you can override this using `MCS_K8S_API_SERVER`, ie use the k8s cluster from `kubectl config view`
+
+## Extract the Service account token and use it with MCS
+
+For local development you can use the jwt associated to the `m3-sa` service account, you can get the token running
+the following command in your terminal:
+
+```
+kubectl get secret $(kubectl get serviceaccount m3-sa -o jsonpath="{.secrets[0].name}") -o jsonpath="{.data.token}" | base64 --decode
```
bexsoft, line 32: I think this needs to change to mcs-sa:

```suggestion
kubectl get secret $(kubectl get serviceaccount mcs-sa -o jsonpath="{.secrets[0].name}") -o jsonpath="{.data.token}" | base64 --decode
```

docs/mcs_operator_mode.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+Then run the mcs server
+
+```
+MCS_MKUBE_ADMIN_ONLY=on ./mcs server
```
bexsoft, line 38: This needs to be MCS_OPERATOR_MODE=on

```suggestion
MCS_OPERATOR_MODE=on ./mcs server
```

docs/mcs_operator_mode.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+# Development
+
+If console is running inside a k8s pod `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` will contain the k8s api server apiServerAddress
+if console is not running inside k8s by default will look for the k8s api server on `localhost:8001` (kubectl proxy)
+
+If you are running mcs in your local environment and wish to make request to `Kubernetes` you can set `MCS_K8S_API_SERVER`, if
+the environment variable is not present by default `MCS` will use `"http://localhost:8001"`, additionally you will need to set the
+`MCS_MKUBE_ADMIN_ONLY=on` variable to make MCS display the Operator UI.
```
cesnietor, line 20: MCS_MKUBE_ADMIN_ONLY I think is no longer used

PR #196, opened 2020-07-15
Add storageClass as part of tenant info api response

cesnietor:


restapi/admin_tenants_test.go
```diff
@@ -233,18 +233,96 @@ func Test_TenantInfo(t *testing.T) {
 		k8sclientGetSecretMock = tt.mockGetSecret
 		k8sclientGetServiceMock = tt.mockGetService
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := getTenantAdminClient(tt.args.ctx, tt.args.client, tt.args.namespace, tt.args.tenantName, tt.args.serviceName, tt.args.scheme)
+			_, err := getTenantAdminClient(tt.args.ctx, tt.args.client, tt.args.namespace, tt.args.tenantName, tt.args.serviceName, tt.args.scheme)
```
Alevsk, line 236: why we stop comparing the actual result?

cesnietor: this is just to ensure that an adminClient is created, we don't need to check the elements of it since that is not the purpose of that function but, I can add an assertion that the type is the one we expect.

cesnietor: also because all contents on AdminClient are Private XD but I've added the validation

PR #207, opened 2020-07-27
Validate Add Zone Paramters

dvaldivia:


restapi/admin_tenants.go
```diff
@@ -600,10 +601,67 @@ func addTenantZone(ctx context.Context, operatorClient OperatorClient, params ad
...
+		},
+	})
+
+	zone := operator.Zone{
 		Name:    params.Body.Name,
 		Servers: int32(*params.Body.Servers),
-	})
+	}
+
+	minInst.Spec.Zones = append(minInst.Spec.Zones, zone)
```
cesnietor, line 664: won't this duplicate the zones? we are doing append twice

cesnietor: .

dvaldivia: indeed

PR #213, opened 2020-07-28
TLS with user provided certificates and KES support for MinIO

Alevsk:
This PR adds the following features:

- Allow user to provide its own keypair certificates for enable TLS in
  MinIO
- Allow user to configure data encryption at rest in MinIO with KES
- Removes JWT schema for login and instead Console authentication will use
  encrypted session tokens

## Enable TLS between client and MinIO with user provided certificates

Instead of using AutoCert feature now the user can provide `cert` and
`key` via `tls` object, values must be valid `x509.Certificate`
formatted files encoded in `base64`

## Enable encryption at rest configuring KES

User can deploy KES via Console/Operator by defining the encryption
object, AutoCert must be enabled or custom certificates for KES must be
provided, KES support 3 KMS backends: `Vault`, `AWS KMS` and `Gemalto`,
previous configuration of the KMS is necessary.

eg of body request for create-tenant

```
{
    "name": "honeywell",
    "access_key": "minio",
    "secret_key": "minio123",
    "enable_mcs": false,
    "enable_ssl": false,
    "service_name": "honeywell",
    "zones": [
        {
            "name": "honeywell-zone-1",
            "servers": 1,
            "volumes_per_server": 4,
            "volume_configuration": {
                "size": 256000000,
                "storage_class": "vsan-default-storage-policy"
            }
        }
    ],
    "namespace": "default",
    "tls": {
      "tls.crt": "",
      "tls.key": ""
    },
    "encryption": {
        "server": {
          "tls.crt": "",
          "tls.key": ""
        },
        "client": {
          "tls.crt": "",
          "tls.key": ""
        },
      "vault": {
        "endpoint": "http://vault:8200",
        "prefix": "",
        "approle": {
          "id": "",
          "secret": ""
        }
      }
    }
}
```


restapi/admin_tenants.go
```diff
@@ -385,10 +388,280 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
...
 		},
 	}
+
+	// operator request AutoCert feature
+	encryption := false
+	if params.Body.EnableSsl != nil {
+		encryption = true
+		minInst.Spec.RequestAutoCert = *params.Body.EnableSsl
+	}
+
```
cesnietor, line 401: can we abstract this in smaller functions? `getTenantCreatedResponse()` is already large.

PR #214, opened 2020-07-28
remove remaining variable references for mcs

harshavardhana:


cesnietor:
Lol this I also thought was necessary but the MCS3Client there reflects the `mc`  S3 client definition is just to call mc functions.
So not needed XD.

harshavardhana:
> Lol this I also thought was necessary but the MCS3Client there reflects the `mc` S3 client definition, is just to call mc functions.
> So not needed XD.

is it necessary to call it `mc` S3 client?

cesnietor:
> is it necessary to call it mc S3 client?
mmh well it is that, any suggestions? probably minioClient?

harshavardhana:
To avoid the confusion moved to `mcClient` instead of `mcS3Client`

PR #219, opened 2020-07-30
Bumps the version of Console when using Operator APIs

dvaldivia:


Dockerfile.release
```diff
@@ -1,6 +1,12 @@
+FROM ubuntu:18.04
+
+RUN apt-get update -y && apt-get install -y ca-certificates
+
 FROM scratch
 MAINTAINER MinIO Development "dev@min.io"
 EXPOSE 9090
 COPY console /console
 
+COPY --from=0 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```
harshavardhana, line 10:
```suggestion
COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```

Dockerfile.release
```diff
@@ -1,6 +1,12 @@
+FROM ubuntu:18.04
```
harshavardhana, line 1:
```suggestion
FROM ubuntu:18.04 as certs
```

dvaldivia:
@harshavardhana done

PR #221, opened 2020-07-31
Support for deploying minio/console with IDP integration

Alevsk:


restapi/admin_tenants.go
```diff
@@ -391,6 +391,56 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
...
+			if tlsSkipVerify {
+				minInst.Spec.Env = append(minInst.Spec.Env, corev1.EnvVar{
+					Name:  "MINIO_IDENTITY_LDAP_TLS_SKIP_VERIFY",
+					Value: "on",
+				})
+			}
+			if serverInsecure {
+				minInst.Spec.Env = append(minInst.Spec.Env, corev1.EnvVar{
+					Name:  "MINIO_IDENTITY_LDAP_SERVER_INSECURE",
+					Value: "on",
```
harshavardhana, line 439: While this is okay for now, will make changes on the server codebase so that this is easily referenced and importable, such that any of these ENV changes console is not broken.

PR #224, opened 2020-08-02
Add Insecure parameter to NewAdminClient function

dvaldivia:
When using the madmin client, for some operations such as health checks against a MinIO instnace with TLS we need a client with insecure turned on.

### DIFF
--- a/restapi/client-admin.go
+++ b/restapi/client-admin.go
@@ -33,8 +33,13 @@ import (
 
 const globalAppName = "console"
 
-// NewAdminClient gives a new client interface
+// NewAdminClient gives a new madmin client interface
 func NewAdminClient(url, accessKey, secretKey string) (*madmin.AdminClient, *probe.Error) {
+	return NewAdminClientWithInsecure(url, accessKey, secretKey, false)
+}
+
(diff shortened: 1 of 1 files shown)


restapi/client-admin.go
```diff
@@ -44,7 +49,7 @@ func NewAdminClient(url, accessKey, secretKey string) (*madmin.AdminClient, *pro
 		AppName:     appName,
 		AppVersion:  ConsoleVersion,
 		AppComments: []string{appName, runtime.GOOS, runtime.GOARCH},
-		Insecure:    false,
+		Insecure:    insecure,
```
harshavardhana, line 52: Looks like this is because the correct transport is not passed in which contains the correct root CAs @dvaldivia ?

dvaldivia: yep, we have the need to query the MinIO's ignoring their TLS configs, because we want to reach their `admin info` API, since each tenant would have a different configuration this would be a easy solution, what do you think?

harshavardhana: Yes it is @dvaldivia my only thing is that as long as we are not doing this for write operations. Reads/Query is fine..

PR #225, opened 2020-08-03
Limit Console RAM to 64Mi. Increase Logging for Tenant APIs.

dvaldivia:


kerneltime:
I guess console memory never goes more than 64 megs ? When listing buckets and keys how do we ensure that the memory footprint is within the min requested.

dvaldivia:
@kerneltime we don't, we have not tested with a large enough number of buckets that would crash the system, and on regular k8s this request means we want at least 64Mi guaranteed, it's not a limit, so if exceeded it would not kill the pod.

if we had to list 500,000 buckets, then we would start taking 64MB with only the bucket names in string format in memory, but that's a pretty insane number of buckets, and a nice problem to have if someone comes with such problem.

kerneltime:
We  might get killed if the pod cannot take the spike from our minimum requested.. I understand that we have not specified a limit but just the floor..

PR #229, opened 2020-08-05
Set default pod management policy, k8s defaults to OrderedReady

kerneltime:


kerneltime:
This does not get set for the tenant in the CRD but the operator defaults it.

PR #233, opened 2020-08-05
Pass Annotations to PVC

dvaldivia:
This PR pushes down the Annotations for the Tenant to the underlying PVCs via their Claim Template

### DIFF
--- a/restapi/admin_tenants.go
+++ b/restapi/admin_tenants.go
@@ -815,9 +815,18 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
 	if tenantReq.ServiceName != "" {
 		minInst.Spec.ServiceName = tenantReq.ServiceName
 	}
+	// add annotations
+	var annotations map[string]string
+	if len(tenantReq.Annotations) > 0 {
+		if minInst.Spec.Metadata == nil {
+			minInst.Spec.Metadata = &metav1.ObjectMeta{}
+		}
+		annotations = tenantReq.Annotations
+		minInst.Spec.Metadata.Annotations = annotations
+	}
 	// set the zones if they are provided
 	for _, zone := range tenantReq.Zones {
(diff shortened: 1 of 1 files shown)


restapi/admin_tenants.go
```diff
@@ -815,9 +815,18 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
...
 		minInst.Spec.ServiceName = tenantReq.ServiceName
 	}
+	// add annotations
+	var annotations map[string]string
+	if len(tenantReq.Annotations) > 0 {
+		if minInst.Spec.Metadata == nil {
+			minInst.Spec.Metadata = &metav1.ObjectMeta{}
+		}
+		annotations = tenantReq.Annotations
+		minInst.Spec.Metadata.Annotations = annotations
```
kerneltime, line 825: Applying the annotation to both PVC and the Tenant?

dvaldivia: Yes, the requests Annotations for the Tenant are pushed down to the PVC

kerneltime: We can split them later I guess..

restapi/admin_tenants.go
```diff
@@ -1241,16 +1243,22 @@ func parseTenantZoneRequest(zoneParams *models.Zone) (*operator.Zone, error) {
...
+			Name: "data",
 		},
+		Spec: volTemp,
+	}
+	if len(annotations) > 0 {
+		vct.ObjectMeta.Annotations = annotations
+	}
+
+	zone := &operator.Zone{
+		Name:                zoneParams.Name,
```
Alevsk, line 1258: Can `zoneParams` be null at some point? I observed its being received as a pointer

dvaldivia: it's validated before calling this function so it can't be null

PR #238, opened 2020-08-08
Allow tolerationSeconds to be empty on Zone tolerations Requests

cesnietor:
Since toleration seconds can be empty, we were forcing it to be an integer defaulting to 0 which
was creating a toleration with value 0 when value should have been nil.
Request now changes to:
```
...
"tolerations": [
                {
                    "effect": "NoExecute",
                    "key": "node.kubernetes.io/not-ready",
                    "operator": "Exists",
                    "tolerationSeconds": {
                        "seconds": 300
                    }
                },
...
```
If tolerationSeconds exists it should be assigned to the Tenant instance with the desired value

If tolerationSeconds is not defined it should not assign any value on the Tenant instance and leave it empty.


swagger.yml
```diff
@@ -2113,21 +2113,29 @@ definitions:
...
+  zoneTolerationSeconds:
+    description: TolerationSeconds represents the period of
+            time the toleration (which must be of effect NoExecute,
+            otherwise this field is ignored) tolerates the taint.
+            By default, it is not set, which means tolerate the taint
+            forever (do not evict). Zero and negative values will
+            be treated as 0 (evict immediately) by the system.
+    type: object
+    required:
+      - seconds
```
dvaldivia, line 2134: I thought this was allowed to be nil :S

restapi/admin_tenants.go
```diff
@@ -1233,12 +1226,18 @@ func parseTenantZoneRequest(zoneParams *models.Zone, annotations map[string]stri
 	// parse tolerations
 	tolerations := []corev1.Toleration{}
 	for _, elem := range zoneParams.Tolerations {
+		var tolerationSeconds *int64
+		if elem.TolerationSeconds != nil {
```
dvaldivia, line 1230: won't the swagger API reject this?

cesnietor: no because TolerationSeconds it is not required, the required one is `seconds` on TolerationSeconds, kind of weird because both are pointers but swagger doesn't validate this.

swagger.yml
cesnietor: it is, that's the main purpose of this PR, zoneTolerationSeconds is not required, but if zoneTolerationSeconds is defined, then seconds is required.

PR #239, opened 2020-08-08
Enable user provided certificates for Console

Alevsk:


dvaldivia:
Fix conflicts

restapi/admin_tenants.go
```diff
@@ -470,26 +470,26 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
 		}
 	}
 
-	// operator request AutoCert feature
-	encryption := false
-	if tenantReq.EnableSsl != nil {
-		encryption = true
+	isEncryptionAvailable := false
+	if *tenantReq.EnableSsl {
```
harshavardhana, line 474: Ssl variable should be changed to TLS everywhere @Alevsk - SSL word, in general, is deprecated and not used anymore. 

Keeps it consistent with all other variables

Alevsk: done 👍

PR #244, opened 2020-08-10
prepareSTSClientTransport tls function refactor

Alevsk:
- Reading root ca certificates operation will run only once after Console
starts, reduce the chance of panics happening during runtime
- Fixed bug in which tls.config insecureSkipVerification configuration
  could get overrided after variable reasignation

### DIFF
--- a/restapi/tls.go
+++ b/restapi/tls.go
@@ -19,16 +19,30 @@ package restapi
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
 	"io/ioutil"
+	"log"
 	"net"
 	"net/http"
 	"time"
 )
 
-var (
-	certDontExists = "File certificate doesn't exists: %s"
-)
+func getCertPool() *x509.CertPool {
+	caCertFileNames := getMinioServerTLSRootCAs()
+	// If CAs certificates are configured we save them to the http.Client RootCAs store
+	certs := x509.NewCertPool()
+	for _, caCert := range caCertFileNames {
+		pemData, err := ioutil.ReadFile(caCert)
+		if err != nil {
+			// logging this error
+			log.Println(err)
+			continue
+		}
+		certs.AppendCertsFromPEM(pemData)
+	}
+	return certs
+}
+
+var certPool = getCertPool()
 
 func prepareSTSClientTransport(insecure bool) *http.Transport {
 	// This takes github.com/minio/minio/pkg/madmin/transport.go as an example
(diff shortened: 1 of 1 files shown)


restapi/tls.go
```diff
@@ -30,24 +30,43 @@ var (
...
-	// This takes github.com/minio/minio/pkg/madmin/transport.go as an example
-	//
-	// DefaultTransport - this default transport is similar to
-	// http.DefaultTransport but with additional param  DisableCompression
-	// is set to true to avoid decompressing content with 'gzip' encoding.
+func getCertPool() *x509.CertPool {
+	caCertFileNames := getMinioServerTLSRootCAs()
+	// If CAs certificates are configured we save them to the http.Client RootCAs store
+	certs := x509.NewCertPool()
+	if len(caCertFileNames) > 0 {
```
harshavardhana, line 37: this is not needed since `for range` supports niladic lists..  simplify this as just 

```
for _, caCert := range caCertFileNames {
```

restapi/tls.go
```diff
@@ -30,24 +30,43 @@ var (
...
+		for _, caCert := range caCertFileNames {
+			// Validate certificate exists
+			if FileExists(caCert) {
```
harshavardhana, line 40: you don't need to validate if the file exists because `ioutil.ReadFile` will dot that anyways.

harshavardhana: any panicking function upon error is generally called `mustFunction()` to indicate to the caller that this can panic. So we should call this 

`mustGetCertPool()`

restapi/tls.go
```diff
@@ -66,33 +85,11 @@ func prepareSTSClientTransport(insecure bool) *http.Transport {
...
-				} else {
-					// if provided cert filename doesn't exists stop console
-					panic(fmt.Sprintf(certDontExists, caCert))
-				}
-			}
-			tlsConfig.RootCAs = certs
+		if insecure {
+			tlsConfig.InsecureSkipVerify = true
+		} else {
+			tlsConfig.RootCAs = certPool
```
harshavardhana, line 91: this entire stuff can be simpler

```
tlsConfig.RootCAs = certPool
tlsConfig.InsecureSkipVerify = getMinIOEndpointIsSecure()
```

restapi/tls.go
Alevsk: fixed, thanks Harsha

restapi/tls.go
Alevsk: good advice, fixed, thanks

PR #248, opened 2020-08-18
encrypt token session using aes-gcm if cpu support it or ChaCha20

Alevsk:


pkg/auth/token.go
```diff
@@ -136,34 +143,111 @@ func decryptClaims(ciphertext string) (*DecryptedClaims, error) {
...
+		algorithm = "AES-256-GCM-HMAC-SHA-256"
+	} else {
+		algorithm = "ChaCha20Poly1305"
 	}
-	cipherText := gcm.Seal(nonce, nonce, plaintext, nil)
-	return cipherText, nil
-}
+	var aead cipher.AEAD
+	switch algorithm {
+	case "AES-256-GCM-HMAC-SHA-256":
```
harshavardhana, line 170: You can use binary constants to keep smaller payload https://github.com/minio/minio/blob/master/pkg/madmin/encrypt.go#L133 @Alevsk

pkg/auth/token.go
```diff
@@ -136,34 +143,111 @@ func decryptClaims(ciphertext string) (*DecryptedClaims, error) {
 	}, nil
 }
 
-// Encrypt a blob of data using AEAD (AES-GCM) with a pbkdf2 derived key
-func encrypt(plaintext []byte) ([]byte, error) {
-	block, _ := aes.NewCipher(derivedKey)
-	gcm, err := cipher.NewGCM(block)
+type SealedSecret struct {
+	Algorithm string `json:"aead"`
+	IV        []byte `json:"iv"`
```
harshavardhana, line 148: Use []byte with json will be super expensive in terms of base64 encoding it needs to do are you okay with that?

PR #252, opened 2020-08-21
add labels to tenant secrets for easy deletion

Alevsk:


restapi/admin_tenants.go
```diff
@@ -769,6 +782,7 @@ func setImageRegistry(ctx context.Context, tenantName string, req *models.ImageR
 		},
 		Data: map[string][]byte{
 			corev1.DockerConfigJsonKey: []byte(string(imRegistryJSON)),
+			operator.TenantLabel:       []byte(tenantName),
```
cesnietor, line 785: shuldn't this be as label?
```
Labels: map[string]string{
						operator.TenantLabel: tenantName,
					},
```

cesnietor: because if it doesnt exist we create it wih this content

dvaldivia: does this needs to be in the data of the secret??

Alevsk: good catch, yes this should be in labels field

PR #257, opened 2020-08-25
Add Tenant Update Zones api

cesnietor:
Since the Tenant's zones is an array, a PUT operation was done where
all zone elements on the Tenant are replaced by the defined ones on the request.

Api to update zones is done on:
PUT `/namespaces/{namespace}/tenants/{tenant}/zones`

Payload can be E.g.:
```
{
    "zones": [
        {
            "name": "zone-0",
            "servers": 2,
            "volumes_per_server": 2,
            "volume_configuration": {
                "storage_class_name": "standard",
                "size": 1073741824
            },
            "resources": {
                "requests": {
                    "memory": 3221225472
                }
            },
            "tolerations": [
                                {
                    "effect": "NoExecute",
                    "key": "node.kubernetes.io/not-ready",
                    "operator": "Exists",
                    "tolerationSeconds": {
                        "seconds": 300
                    }
                },
                {
                    "effect": "NoExecute",
                    "key": "node.kubernetes.io/unreachable",
                    "operator": "Exists",
                    "tolerationSeconds": {
                        "seconds": 300
                    }
                },
            ],
            "affinity": {
                "podAntiAffinity": {
                    "requiredDuringSchedulingIgnoredDuringExecution": [
                        {
                            "topologyKey": "kubernetes.io/hostname",
                            "labelSelector": {
                                    "matchExpressions": [
                                        {
                                            "key": "v1.min.io/tenant",
                                            "operator": "In",
                                            "values": ["minio-tenant-1"]
                                        },
                                        {
                                            "key": "v1.min.io/zone",
                                            "operator": "In",
                                            "values": ["zone-0"]
                                        }
                                    ]
                            }
                        }
                    ]
                }   
            }
        }
    ]
}
```

Response will be in the format of `models.Tenant` (the one gotten on `/namespaces/{namespace}/tenants/{tenant}`)


swagger.yml
```diff
@@ -2534,3 +2563,13 @@ definitions:
...
+  
+  zoneUpdateRequest:
+    type: object
+    required:
+      - zones
+    properties:
+      zones:
+        type: array
+        items:
+          $ref: "#/definitions/zone"
```
Alevsk, line 2575: not a blocker but its a good practice to have an empty line at the end of the file because of file processing https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

cesnietor: done

restapi/admin_tenants.go
```diff
@@ -1751,3 +1760,71 @@ func getKESConfiguration(ctx context.Context, clientSet *kubernetes.Clientset, n
...
+	if minInst.Spec.Metadata == nil {
+		minInst.Spec.Metadata = &metav1.ObjectMeta{
+			Annotations: map[string]string{},
+		}
+	}
+
+	// set the zones if they are provided
+	var newZoneArray []operator.Zone
+	for _, zone := range zonesReq {
+		zone, err := parseTenantZoneRequest(zone, minInst.Spec.Metadata.Annotations)
```
dvaldivia, line 1811: `parseTenantZoneRequest` would allow to change fields that shouldn't be changed after zone creation, such as PVC storage Class or size, number of servers, etc, we should be careful not to change them since that could break havok on the tenant, can we just grab the fields that are safe to change? like memory, node selector,  affinity and tolerations?

cesnietor: since it is possible to have two zones with the same name, it would be difficult to modify the correct one, also it would make it messy in the sense that we would need to loop to find which to edit.

restapi/admin_tenants.go
```diff
@@ -1751,3 +1760,71 @@ func getKESConfiguration(ctx context.Context, clientSet *kubernetes.Clientset, n
...
+	}
+
+	if minInst.Spec.Metadata == nil {
+		minInst.Spec.Metadata = &metav1.ObjectMeta{
+			Annotations: map[string]string{},
+		}
+	}
+
+	// set the zones if they are provided
+	var newZoneArray []operator.Zone
```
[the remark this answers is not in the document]
cesnietor: no, there can be multiple zones with the same name

PR #263, opened 2020-09-01
fix tls certPool client regression

Alevsk:


restapi/tls.go
```diff
@@ -28,6 +28,9 @@ import (
 
 func getCertPool() *x509.CertPool {
 	caCertFileNames := getMinioServerTLSRootCAs()
+	if len(caCertFileNames) == 0 {
```
harshavardhana, line 31: You can return instead, avoids caller being worried about the returned value being `nil` or not. 

```
x509.SystemCertPool()
```

restapi/tls.go
```diff
@@ -28,6 +28,9 @@ import (
...
+		return nil
+	}
```
harshavardhana, line 33: A better implementation is to use as below

```
// getCertPool - adds all the root CAs into certPool
// at the certPaths list
func getCertPool(certPaths  ...string) (*x509.CertPool) {
        rootCAs, _ := x509.SystemCertPool()
        if rootCAs == nil {
                // In some systems (like Windows) system cert pool is
                // not supported or no certificates are present on the
                // system - so we create a new cert pool.
                rootCAs = x509.NewCertPool()
        }

        // Load all custom CA files.
        for _, certPath := range certPaths {
                caCert, err := ioutil.ReadFile(certPath)
                if err != nil {
                        // ignore files which are not readable.
                        continue
                }
                rootCAs.AppendCertsFromPEM(caCert)
        }

        return rootCAs
}
```

PR #264, opened 2020-09-02
Replace resources api to return the max allocatable memory

cesnietor:
Resouces api changed to GET on `/api/v1/cluster/max-allocatable-memory`
which returns the max allocatable memory for a given number of nodes, 
nodes with taints NoExecute or NoSchedule shall not be considered.
request needs to be e.g.
```
/api/v1/cluster/max-allocatable-memory?num_nodes=4
```
Response payload:
```
{
    "max_memory": 4131528704
}
```
If not enough nodes or no available memory response will be empty `{}`

- [X] Tests included


swagger.yml
```diff
@@ -1227,15 +1227,21 @@ paths:
       tags:
         - AdminAPI
   
-  /cluster/resources:
```
bexsoft, line 1230: Is this is no longer needed?

cesnietor: not anymore, this was meant to be for the same feature of calculating the max ram but we'll do it in this new api

Alevsk (approved):

swagger.yml
dvaldivia: I think resources could also express storage capacity if available right? I think nesting this API under resources made sense

cesnietor: that api is no longer needed, was designed for the main purpose

cesnietor:
PR modified to use queryparameter instead of body parameter

dvaldivia (approved):

Alevsk (approved):

PR #271, opened 2020-09-06
fix bug for tenant image pull credentials

Alevsk:


restapi/admin_tenants.go
```diff
@@ -792,24 +792,23 @@ func setImageRegistry(ctx context.Context, tenantName string, req *models.ImageR
...
+			instanceSecret := corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: pullSecretName,
+					Labels: map[string]string{
+						operator.TenantLabel: tenantName,
+					},
+				},
+				Data: secretCredentials,
+				Type: corev1.SecretTypeDockerConfigJson,
+			}
```
cesnietor, line 811: what is the difference? or what was the issue? I see that almost the same was happening in the previous behavior

dvaldivia: we may be replacing some content the previous secret already had, for example owner references, so we are only updating the data.

Alevsk: in the previous code we were updating not only the Data field of the secret, we was also updating the time, labels and names, in case someone added a custom label to the secret or something this will override that, with this change we are only updating the Data of the secret

dvaldivia (approved):

restapi/admin_tenants.go
cesnietor: got it, we could also think on doing a patch if we think the object might mutate too often.

cesnietor (approved):

PR #275, opened 2020-09-10
React tests

Alevsk:


dvaldivia:
can these test be run by the github workflow as well?

Alevsk:
> can these test be run by the github workflow as well?

yes, ill added, im just finishing writing some test cases for testing react components and then will test with the github flow

PR #281, opened 2020-09-18
fix regression when calculating token using chacha20

Alevsk:


harshavardhana (approved):

aead:
Guys, wait - things change is not correct. Now, tokens are malleable. The correct fix is to generate a `16` byte IV, not a `32` byte IV an then ignoring half of it...

PR #283, opened 2020-09-22
IV generation for ChaCha20 poly auth scheme

Alevsk:
Generate 16 bytes IV instead of an IV of 32 bytes (and then use half of it) when using ChaCha20 to
encrypt tokens, this is to prevent tokens to become malleable.

## How to Test

- Login into console/operator UI
- Session should work correctly 

Screen Shot 2020-09-21 at 6 14 18 PM

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -155,10 +155,10 @@ const (
 // or data key provided as plaintext.
 //
 // The returned ciphertext data consists of:
-//    iv | AEAD ID | nonce | encrypted data
-//     32      1        12      ~ len(data)
+//    AEAD ID | iv | nonce | encrypted data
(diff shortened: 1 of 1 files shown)


harshavardhana (dismissed):

aead (commented):
The solution of this PR would work. However, it may be slightly simpler to just use `16` byte IVs for all algorithms:

```
AEAD ID | IV | nonce | encrypted data
     1   16      12           len(data) + 16  
```

It is still perfectly fine w.r.t. nonce reuse prevention. A `16` byte / `128` bit IV + `96` bit nonce will  still repeat with negligible probability even when calling `encrypt` about `~ 2^64` times.

aead (approved):

harshavardhana (approved):

PR #285, opened 2020-09-22
Added Annotations, Labels and NodeSelector fields

Alevsk:
For Console/Encryption objects in the  CreateTenant Api


dvaldivia:
you are not going to read the `image` from the `console` attribute? I see `consoleCOnfiguration` has the field in there

dvaldivia (dismissed):

cesnietor (dismissed):

cesnietor (commented):
how is the console image property being used?

Alevsk:
> how is the console image property being used?

will be used in the future, in the meantime we can use the existing `console_image` so we dont break the frontend

Alevsk:
> you are not going to read the `image` from the `console` attribute? I see `consoleCOnfiguration` has the field in there

will be used in the future, in the meantime we can use the existing `console_image` so we dont break the frontend

cesnietor (approved):

dvaldivia (approved):

Issue #290, opened 2020-09-28
Add list objects


cesnietor:
Show the objects within a bucket on the UI.
- [x] Create API to get objects of a bucket
- [x] Create UI to list bucket's objects

PR #310, opened 2020-10-02
Add list objects UI and integrate with listing and delete api

cesnietor:
fixes https://github.com/minio/console/issues/290
Integrates apis for objects.
When you click on a bucket it will redirect to the objects of that bucket, right know it only does a single redirect.
Also if the item is a folder or a single object, it should be deleted with no issues. (the api calls either recursive or not)
Screen Shot 2020-10-01 at 5 17 00 PM
Screen Shot 2020-10-01 at 5 17 06 PM

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/Buckets.tsx
+++ b/portal-ui/src/screens/Console/Buckets/Buckets.tsx
@@ -1,4 +1,4 @@
-// This file is part of MinIO Buckets Server
+// This file is part of MinIO Console Server
 // Copyright (c) 2020 MinIO, Inc.
 //
 // This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


portal-ui/src/screens/Console/Buckets/Buckets.tsx
```diff
@@ -1,4 +1,4 @@
-// This file is part of MinIO Buckets Server
```
dvaldivia, line 1: hahaha

dvaldivia (dismissed):

Alevsk (approved):

dvaldivia (approved):

Alevsk (approved):

PR #314, opened 2020-10-05
Handle Invalid User error coming from madmin

dvaldivia:
Right now we display `Internal Server Error` when invalid credentials are presneted, this makes it so we only present `Unauthorized`

### DIFF
--- a/restapi/error.go
+++ b/restapi/error.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/go-openapi/swag"
 	"github.com/minio/console/models"
+	"github.com/minio/minio/pkg/madmin"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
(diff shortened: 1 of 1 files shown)


cesnietor (commented):
If minio changes the message this will no longer happen, I suggest using ErrAdminNoSuchUser from minio.

dvaldivia:
@cesnietor i tried but that error is an integer so it's hard to compare against as it gets transformed into a api response which makes the comparison harder

restapi/error.go
```diff
@@ -86,6 +87,11 @@ func prepareError(err ...error) *models.Error {
 			errorCode = 401
 			errorMessage = errorGenericInvalidSession.Error()
 		}
+		// console invalid session error
+		if strings.Contains(err[0].Error(), "The specified user does not exist.") {
+			errorCode = 401
+			errorMessage = errorGenericInvalidSession.Error()
+		}
 		// if we received a second error take that as friendly message but dont override the code
```
harshavardhana, line 95: you can use `madmin.ToErrorResponse(err).Code == "XMinioAdminNoSuchUser"`

harshavardhana (changes requested):

restapi/error.go
dvaldivia: done!

bexsoft (approved):

harshavardhana (approved):

PR #317, opened 2020-10-07
 Consolidate Remote Buckets and Replication modals

### LINKED ISSUE #301: Integrate Remote Buckets into a single screen
Right now we have support for setting remove buckets by creating an ARN with the `Remote Buckets` screen and then going into the bucket details and setting a `Replication Rule`.

We need to collapse these two screens into one, ideally, from bucket details we should be able to set the ARN for replication that uses the current bucket as the source bucket and defines the target bucket, and then use that ARN to enable the UI to set `Replication Rules` and then create use use that "implied ARN" for the bucket

kaankabalak:
This commit consolidates Remote Buckets and Replication modals into a
single modal in the Add Replication Rule modal located in the Buckets
section. This commit also removes the Remote Buckets section, as the
operations previously done from there can be done directly from the
Add Replication Rule modal.

Fixes #301


dvaldivia:
Can we align the tabs and the button

Screen Shot 2020-10-06 at 9 59 22 PM

kaankabalak:
> Can we align the tabs and the button
> 
> Screen Shot 2020-10-06 at 9 59 22 PM

Fixed, would appreciate if you could review again 👍

dvaldivia (approved):

Alevsk (approved):
Lgtm

PR #323, opened 2020-10-09
Get Tenant Secret From Tenant CR

dvaldivia:
We were assuming the Tenant Credentials Secret instead of reading it from it's .spec.credsSecret this commit addresses that


Alevsk (approved):

bexsoft (approved):
LGTM

Issue #306, opened 2020-10-01
Logout users when the session is no longer valid


dvaldivia:
Right now when the Console session expires we are not logging out the users.

The session endpoint returns 200, it should returns a status that indicates the session has expired.

Doing calls with an expired token returns 500

<img width="1402" alt="Screen Shot 2020-10-01 at 9 36 26 AM" src="https://user-images.githubusercontent.com/18384552/94837975-c39ba780-03c9-11eb-86cd-c5a497c66814.png">

PR #332, opened 2020-10-19
Invalidate console session when minio user doesn't exists

Alevsk:
## Testing

- Create new user and assign policy:

```
./mc admin user add orcadmin alevsk alevsk12356
./mc admin policy set orcadmin mcsAdmin user=alevsk
```
- Login to console and list buckets/users/policies etc
- Delete the user using `mc`
```
./mc admin user remove orcadmin alevsk
```
- Try to navigate some more in Console, you should be redirected to the login screen since your session will be invalidated

Fixes https://github.com/minio/console/issues/306

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,6 @@ Additionally, you can create policies to limit the privileges for `console` user
 To run the server:
 
 ```
-export CONSOLE_HMAC_JWT_SECRET=YOURJWTSIGNINGSECRET
-
 #required to encrypt jwet payload
 export CONSOLE_PBKDF_PASSPHRASE=SECRET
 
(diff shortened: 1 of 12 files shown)


dvaldivia:
Will this log out the user on the UI?

dvaldivia (approved):

cesnietor (changes requested):
testing.

cesnietor (dismissed):
Tested, just if the user is disabled it won't redirect to login page but as discussed previously  this will be investigated.
LGTM.

Alevsk:
> Will this log out the user on the UI?

yes, in the UI user will be redirected to the login screen, we already have some logic to do that

dvaldivia (approved):

cesnietor (approved):

Issue #234, opened 2020-08-05
SNI support for Console


Alevsk:
Similar to https://github.com/minio/minio/pull/10207 Console needs to have SNI support for the scenario in which is being deployed inside a `kubernetes` cluster and it's directly exposed to the internet via Service LoadBalancer/NodePort

PR #352, opened 2020-10-28
SNI support for Console

Alevsk:
Fixes https://github.com/minio/console/issues/234


pkg/certs/certs.go
```diff
@@ -0,0 +1,224 @@
...
+	return filepath.Join(GlobalCertsDir.Get(), PublicCertFile)
+}
+
+func getPrivateKeyFile() string {
+	return filepath.Join(GlobalCertsDir.Get(), PrivateKeyFile)
+}
+
+func GetTLSConfig() (x509Certs []*x509.Certificate, manager *certs.Manager, err error) {
+
+	GlobalContext = context.Background()
```
aead, line 148: Why is `GlobalContext` a package variable if we (re)set it here to `context.Background()`

harshavardhana: yeah we can't do it like this, this looks rather scary.

pkg/certs/certs.go
```diff
@@ -0,0 +1,224 @@
...
+	"strings"
+
+	"github.com/minio/cli"
+	"github.com/minio/minio/cmd/config"
+	"github.com/minio/minio/cmd/logger"
+	"github.com/minio/minio/pkg/certs"
+	"github.com/mitchellh/go-homedir"
+)
+
+var GlobalContext context.Context
```
harshavardhana, line 35:
```suggestion
var GlobalContext context.Context
```

This should be initialized

```
// GlobalContext context that is canceled when the server is requested to shut down.
var GlobalContext context.Context

// cancelGlobalContext can be used to indicate server shutdown.
var cancelGlobalContext context.CancelFunc

// Initialize service mutex once.
func init() {
        initGlobalContext()
}

func initGlobalContext() {
        GlobalContext, cancelGlobalContext = context.WithCancel(context.Background())
        GlobalServiceDoneCh = GlobalContext.Done()
}
```

harshavardhana (changes requested):

pkg/certs/certs.go
Alevsk: right, talked with harsha, removed globalContext since console doesnt support that yet, im using just `context.Background()` instead

pkg/certs/certs.go
Alevsk: removed globalContext since console doesnt support that yet, im using just context.Background() instead

README.md
```diff
@@ -113,11 +113,41 @@ export CONSOLE_MINIO_SERVER=http://localhost:9000
 ./console server
 ```
 
+## Run Console with TLS enable
+
+Copy your `public.crt` and `private.key` to `~/.console/certs`, then:
+
+```$xslt
```
harshavardhana, line 120:
```suggestion
```bash
```

README.md
```diff
@@ -113,11 +113,41 @@ export CONSOLE_MINIO_SERVER=http://localhost:9000
...
+./console server
+```
+
+Additionally, `Console` has support for multiple certificates, clients can request them using `SNI`. It expects the following structure:
+
+```$xslt
```
harshavardhana, line 126:
```suggestion
```
```

File: restapi/consts.go [hunk 5]
```diff
@@ -18,16 +18,15 @@ package restapi
...
-	ConsoleAccessKey             = "CONSOLE_ACCESS_KEY"
-	ConsoleSecretKey             = "CONSOLE_SECRET_KEY"
-	ConsoleMinIOServer           = "CONSOLE_MINIO_SERVER"
-	ConsoleMinIOServerTLSRootCAs = "CONSOLE_MINIO_SERVER_TLS_ROOT_CAS"
-	ConsoleProductionMode        = "CONSOLE_PRODUCTION_MODE"
-	ConsoleHostname              = "CONSOLE_HOSTNAME"
-	ConsolePort                  = "CONSOLE_PORT"
-	ConsoleTLSHostname           = "CONSOLE_TLS_HOSTNAME"
-	ConsoleTLSPort               = "CONSOLE_TLS_PORT"
+	ConsoleVersion        = `0.1.0`
```
harshavardhana, line 21: A breaking change so perhaps 

```suggestion
	ConsoleVersion        = `0.2.0`
```

harshavardhana (dismissed):

aead (approved):

dvaldivia (approved):

PR #356, opened 2020-10-28
Add put object api and add list object improvements

cesnietor:
To add object tags we can now do PUT on `/api/v1/buckets/<bucketName>/objects/tags?prefix=<prefix>&version_id=<object_version>`
 Request payload example:
```
{
    "tags": {
        "tag1": "val1",
        "tag2": "val2"
    }
}
```
When doing GET it will list the tags only for single object defined on prefix like:
`/api/v1/buckets/<bucketName>/objects?prefix=folder/file.txt&with_versions=true`
Response Payload:
```
{
    "objects": [
        {
            "is_latest": true,
            "last_modified": "2020-10-14 21:54:00.842 +0000 UTC",
            "legal_hold_status": "OFF",
            "name": "file.txt",
            "retention_mode": "COMPLIANCE",
            "retention_until_date": "2020-10-23 15:04:05 +0000 UTC",
            "size": 19822527,
            "tags": {
                "tag1": "val1",
                "tag2": "val2"
            },
            "version_id": "be2e946b-fa54-4a7a-9d79-f2392e221585"
        },
        {
            "last_modified": "2020-10-14 21:31:39.231 +0000 UTC",
            "name": "file.txt",
            "version_id": "f3eb5d80-9bfb-4eea-a554-236ea8dafff0"
        }
    ],
    "total": 2
}
```

- [X]  Tests included


swagger.yml
```diff
@@ -437,6 +437,38 @@ paths:
...
+          required: true
+          type: string
+        - name: prefix
+          in: query
+          required: true
+          type: string
+        - name: version_id
+          in: query
+          required: true
+          type: string
```
Alevsk, line 456: is there any special reason about why `prefix` and `query` cannot be part of the `putObjectTagsRequest` object in the request body?, everything else looks good

Alevsk (approved):

dvaldivia (approved):

PR #382, opened 2020-11-10
Add set policy to multiple user/groups

cesnietor:
Added api that allow setting policy to multiple users and/or groups.
PUT api on `/api/v1/set-policy-multi/<policy_name>`

Request payload response e.g.:
```
{
    "users": ["user1"],
    "groups": ["newgroup1"]
}
```

users and groups can be empty.

- [X] Tests included


restapi/admin_policies.go
```diff
@@ -248,6 +255,37 @@ func getSetPolicyResponse(session *models.Principal, name string, params *models
...
+	if err := setPolicyMultipleEntities(ctx, adminClient, name, params.Users, params.Groups); err != nil {
+		return prepareError(err)
+	}
+	return nil
+}
+
+// setPolicyMultipleEntities sets a policy to multiple users/groups
+func setPolicyMultipleEntities(ctx context.Context, client MinioAdmin, policyName string, users, groups []string) error {
+	for _, user := range users {
+		if err := client.setPolicy(ctx, policyName, user, false); err != nil {
```
dvaldivia, line 277: if there's an error because a single user is wrong we'll go into a bad state wouldn't we, we some users were applied the policy and the rest was not due to one incorrect user in the middle of the list, why dont we add a validation for users and groups before we apply the policies?

dvaldivia (changes requested):

restapi/admin_policies.go
cesnietor: yes, there's currently no way to make this 100% atomic.

dvaldivia: it's ok if it's not atomic, but we should at least make sure all the users and groups are valid before we start applying the policy

dvaldivia: @cesnietor

cesnietor: I've updated to validate on request the user/group string based on a regex defined on aws restrictions.
```
User names can be a combination of up to 64 letters, digits, and these characters: plus (+), equal (=), comma (,), period (.), at sign (@), underscore (_), and hyphen (-). 
```

Alevsk (approved):

bexsoft (approved):

PR #390, opened 2020-11-12
Support for Cookie authentication

Alevsk:
- Added support for cookie authentication (authorization header authentication still exists and will have priority)
- Removed local storage token management from UI
- Cookie hardening (sameSite, httpOnly, secure), leverage on browser existing protections
- Login endpoint sets cookie via header, logout endpoint expires cookie

Future improvements

- Look for all places in backend that returns 401 unauthorized, and destroy session there (not a priority since cookie its invalid anyway)
- Downloading objects in object browser can be simplified since is just a GET request and users will be authenticated via Cookies, no need to craft additional requests

### DIFF
--- a/restapi/configure_console.go
+++ b/restapi/configure_console.go
@@ -21,6 +21,7 @@ package restapi
 import (
 	"bytes"
 	"crypto/tls"
+	"fmt"
 	"log"
 	"net/http"
 	"strings"
@@ -168,8 +169,10 @@ func setupMiddlewares(handler http.Handler) http.Handler {
 // The middleware configuration happens before anything, this middleware also applies to serving the swagger.json document.
 // So this is a good place to plug in a panic handling middleware, logging and metrics
 func setupGlobalMiddleware(handler http.Handler) http.Handler {
+	// handle cookie or authorization header for session
+	next := AuthenticationMiddleware(handler)
 	// serve static files
-	next := FileServerMiddleware(handler)
+	next = FileServerMiddleware(next)
 	// Secure middleware, this middleware wrap all the previous handlers and add
 	// HTTP security headers
 	secureOptions := secure.Options{
@@ -198,6 +201,31 @@ func setupGlobalMiddleware(handler http.Handler) http.Handler {
 	secureMiddleware := secure.New(secureOptions)
 	app := secureMiddleware.Handler(next)
 	return app
+}
+
+func AuthenticationMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// prioritize authorization header and skip
+		if r.Header.Get("Authorization") != "" {
+			next.ServeHTTP(w, r)
+			return
+		}
+		tokenCookie, err := r.Cookie("token")
+		if err != nil {
+			next.ServeHTTP(w, r)
+			return
+		}
+		currentTime := time.Now()
+		if tokenCookie.Expires.After(currentTime) {
+			next.ServeHTTP(w, r)
+			return
+		}
+		token := tokenCookie.Value
+		if token != "" {
+			r.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
+		}
+		next.ServeHTTP(w, r)
+	})
 }
 
 // FileServerMiddleware serves files from the static folder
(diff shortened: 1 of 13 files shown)


harshavardhana:
why are we doing cookie based?

dvaldivia (approved):

Alevsk:
> why are we doing cookie based?

Console is a browser application, there's no need to the use localStorage (we still maintain the support for clients that doesn't support cookies), instead we can use browser cookies and leverage on the existing security protections, right now the session token it's too expose and any injected malicious JS code would be able to read that information

Alevsk:
> > why are we doing cookie based?
> 
> Console is a browser application, there's no need to the use localStorage (we still maintain the support for clients that doesn't support cookies), instead we can use browser cookies and leverage on the existing security protections, right now the session token it's too expose and any injected malicious JS code would be able to read that information

cookie based auth expose us to other series of threats such as CSRF but we can also fix those if ever happen, we need to add layers of security

dvaldivia (dismissed):

restapi/configure_console.go
```diff
@@ -200,6 +203,26 @@ func setupGlobalMiddleware(handler http.Handler) http.Handler {
 	return app
 }
 
+func AuthenticationMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// prioritize authorization header and skip
```
aead, line 208: What do we `skip`?

restapi/configure_console.go
```diff
@@ -200,6 +203,26 @@ func setupGlobalMiddleware(handler http.Handler) http.Handler {
...
+		if r.Header.Get("Authorization") != "" {
+			next.ServeHTTP(w, r)
+			return
+		}
+		tokenCookie, err := r.Cookie("token")
+		if err != nil {
+			next.ServeHTTP(w, r)
+			return
+		}
+		token := tokenCookie.Value
```
aead, line 218: Are we not checking for `MaxAge`, `HttpOnly` etc. https://golang.org/pkg/net/http/#Cookie

Alevsk: @aead MaxAge was meant mostly for the browser, but of course we can validate here too and skip if the cookie is not valid anymore

restapi/configure_console.go
Alevsk: We skip copying the token from the cookie into the Authorization header

harshavardhana:
> Console is a browser application, there's no need to the use localStorage (we still maintain the support for clients that doesn't support cookies), instead we can use browser cookies and leverage on the existing security protections, right now the session token it's too expose and any injected malicious JS code would be able to read that information

Almost every browser supports localStorage - Cookies are unnecessarily complicated - unless we are fixing something real.

harshavardhana:
> Almost every browser supports localStorage - Cookies are unnecessarily complicated - unless we are fixing something real.

Also any injected JS cannot access localStorage because browsers make sure that new JS code is never able to read your application's localStorage.

harshavardhana:
Ideally, we should be using https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage if we want to be paranoid about it.

Alevsk:
> Ideally, we should be using https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage if we want to be paranoid about it.

I'll take a look harsha

Alevsk:
> Ideally, we should be using https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage if we want to be paranoid about it.

But it's not only that, the fact that JS (our own code or a third-party library) can read our token its pretty scary, right now we are using cookies already because of websocket communication (websockets authentication works only via cookies as far as I know, @cesnietor can confirm that), so this PR will unify the session handling mechanism to use only cookies instead of cookies and authorization header.

harshavardhana:
> But it's not only that, the fact that JS (our own code or a third-party library) can read our token its pretty scary, right now we are using cookies already because of websocket communication (websockets authentication works only via cookies as far as I know, @cesnietor can confirm that), so this PR will unify the session handling mechanism to use only cookies instead of cookies and authorization header.

Understood

harshavardhana (approved):

restapi/configure_console.go
Alevsk:
```
		if tokenCookie.Expires.After(currentTime) {
			next.ServeHTTP(w, r)
			return
		}
```

Added After validation, so if the cookie its expired we will skip the logic below, in general there's not much to validate here, since the cookie is user provided value and he can manipulate it, what matters its the encrypted sts token inside the cookie

dvaldivia (approved):

PR #399, opened 2020-11-18
Connected delete & download buttons for object details

bexsoft:
## What does this do?

Connects download & delete buttons from object details view


dvaldivia (approved):

kaankabalak (approved):
Looking good to me, maybe in the future we might want to add a Snackbar to notify the user that the object has been deleted similar to what we do when we upload an object.

bexsoft:
> Looking good to me, maybe in the future we might want to add a Snackbar to notify the user that the object has been deleted similar to what we do when we upload an object.

Will create a ticket for it

PR #421, opened 2020-11-19
Connected object tags API

bexsoft:
## What does this do?

Connects object tags API to object details screen.

**NOTE:** There is a small delay between tag changes after saving any modification, will fix this after adding loaders to object details page.

## How does it look?

Screen Shot 2020-11-19 at 13 16 26
Screen Shot 2020-11-19 at 14 58 20
Screen Shot 2020-11-19 at 13 16 07
Screen Shot 2020-11-19 at 13 15 58

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ObjectDetails/AddTagModal.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ObjectDetails/AddTagModal.tsx
@@ -1,0 +1,138 @@
+import React, { useState } from "react";
+import { Button, Grid } from "@material-ui/core";
+import InputBoxWrapper from "../../../../Common/FormComponents/InputBoxWrapper/InputBoxWrapper";
+import ModalWrapper from "../../../../Common/ModalWrapper/ModalWrapper";
+import { createStyles, Theme, withStyles } from "@material-ui/core/styles";
+import { modalBasic } from "../../../../Common/FormComponents/common/styleLibrary";
+import api from "../../../../../../common/api";
+
+interface ITagModal {
+  modalOpen: boolean;
+  currentTags: any;
+  bucketName: string;
+  versionId: string;
+  onCloseAndUpdate: (refresh: boolean) => void;
+  selectedObject: string;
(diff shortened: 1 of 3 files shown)


dvaldivia (dismissed):

cesnietor (changes requested):
I saw that we set the same value on `key` and `value` but minio allows them to be different, so if the object has tags already set by mc we would show only one value:
```
mc tag list s3/testbucket/testobject
Name                :    testobject
editable            :    only-by-owner-and-authenticated
confidentiality     :    open-to-authenticated-only
```
we might need to change to show/set both key and values.

dvaldivia:
@cesnietor would showing that need some changes to the UI?

cesnietor:
> @cesnietor would showing that need some changes to the UI?

yes probably a redesign or do a `key:value` on tag itself.

bexsoft:
But in this case what would be the purpose of showing the key too?, wouldn't it be better to make the labels as string arrays? I think people would be used to look for complete phrases rather than key values

cesnietor:
> But in this case what would be the purpose of showing the key too?, wouldn't it be better to make the labels as string arrays? I think people would be used to look for complete phrases rather than key values

yes, I agree, but since we mirror mc's behavior we need to do an equivalent thing.  Also it is an AWS thing https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html

bexsoft:
> I saw that we set the same value on `key` and `value` but minio allows them to be different, so if the object has tags already set by mc we would show only one value:
> 
> ```
> mc tag list s3/testbucket/testobject
> Name                :    testobject
> editable            :    only-by-owner-and-authenticated
> confidentiality     :    open-to-authenticated-only
> ```
> 
> we might need to change to show/set both key and values.

Added changed to be S3 Compliant:

Screen Shot 2020-11-19 at 19 00 11
Screen Shot 2020-11-19 at 19 00 03
Screen Shot 2020-11-19 at 18 59 56
Screen Shot 2020-11-19 at 18 59 43

cesnietor (approved):
tested LGTM

kaankabalak (approved):

PR #432, opened 2020-11-20
Adding context timeout to login and get location operations

Alevsk:


restapi/user_login.go
```diff
@@ -113,17 +114,18 @@ func getConfiguredRegionForLogin(client MinioAdmin) (string, error) {
...
 		return nil, prepareError(err, errorGeneric)
 	}
 	adminClient := adminClient{client: mAdmin}
 	// obtain the configured MinIO region
 	// need it for user authentication
-	location, err := getConfiguredRegionForLogin(adminClient)
+	location, err := getConfiguredRegionForLogin(ctx, adminClient)
 	if err != nil {
-		return nil, prepareError(err, errorGeneric)
+		return nil, prepareError(err, errConnectingToMinio)
```
cesnietor, line 128: is this the only error that minio can return? or are there more reasons why there could be an error? cause for all errors we would use the same message `errConnectingToMinio`.

Alevsk: there could be many different errors, but this is the error we will show the user in the UI (similar to something went wrong when connecting to minio, a generic error), for debugging purposes the actual `err` will be shown in console logs with more details of what failed

cesnietor (approved):

bexsoft (approved):

PR #445, opened 2020-11-25
Fixed object lock

### LINKED ISSUE #443: Make bucket with versioning enabled is not working
When creating the bucket the ObjectLocking flag should be in the same  makeBucket api but the current implementation is doing it after creating the bucket, hence it doesn't enable the versioning correctly.
Solution is passing objectLocking as input:
```
// implements minio.MakeBucketWithContext(ctx, bucketName, location)
func (c minioClient) makeBucketWithContext(ctx context.Context, bucketName, location string, objectLocking bool) error {
	return c.client.MakeBucket(ctx, bucketName, minio.MakeBucketOptions{
		Region:        location,
		ObjectLocking: objectLocking,
	})
}
```

### LINKED ISSUE #444: Fix an issue with bucket creation & versioning enabling
Versioning enable needs to be done during bucket creation not after it, need to fix logic to avoid issues during this creation

bexsoft:
fixes #443 
fixes #444 

## What does this do?

Fixes an issue with object lock during bucket creation

### DIFF
--- a/restapi/user_buckets.go
+++ b/restapi/user_buckets.go
@@ -285,9 +285,9 @@ func getListBucketsResponse(session *models.Principal) (*models.ListBucketsRespo
 }
 
 // makeBucket creates a bucket for an specific minio client
-func makeBucket(ctx context.Context, client MinioClient, bucketName string) error {
+func makeBucket(ctx context.Context, client MinioClient, bucketName string, objectLocking bool) error {
 	// creates a new bucket with bucketName with a context to control cancellations and timeouts.
-	if err := client.makeBucketWithContext(ctx, bucketName, "us-east-1"); err != nil {
+	if err := client.makeBucketWithContext(ctx, bucketName, "us-east-1", objectLocking); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


restapi/user_buckets.go
```diff
@@ -309,16 +309,16 @@ func getMakeBucketResponse(session *models.Principal, br *models.MakeBucketReque
...
 		return prepareError(err)
 	}
 	// if versioned
-	if br.Versioning {
+	/*if br.Versioning {
 		// we will tolerate this call failing
 		if err := minioClient.enableVersioning(ctx, *br.Name); err != nil {
 			log.Println("error versioning bucket:", err)
 		}
-	}
```
cesnietor: could you please delete this? other than that LGTM

bexsoft: deleted

cesnietor (approved):

cesnietor (approved):

Alevsk (approved):
LGTM

PR #450, opened 2020-11-25
Create Tenant TLS refactor

Alevsk:
- fixed small bug in which RequestAutoCert was not setting properly
- support AutoCert and external certificates on Tenant creation

### DIFF
--- a/restapi/admin_tenants.go
+++ b/restapi/admin_tenants.go
@@ -433,6 +433,7 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
 	tenantReq := params.Body
 	minioImage := tenantReq.Image
 	ctx := context.Background()
+	consoleHasTLS := false
 
 	if minioImage == "" {
 		minImg, err := cluster.GetMinioImage()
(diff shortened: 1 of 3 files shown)


bexsoft (approved):

restapi/admin_tenants.go
```diff
@@ -600,15 +602,10 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
 		}
 		minInst.Spec.ExternalCertSecret = externalCertSecret
 	}
-
+	// If encryption configuration is present and TLS will be enabled (using AutoCert or External certificates)
```
cesnietor, line 605: I didn't understand this comment XD

restapi/admin_tenants_helper.go
```diff
@@ -196,15 +196,15 @@ func getKESConfiguration(ctx context.Context, clientSet K8sClientI, ns string, e
...
 		Replicas: 1,
 	}
 	// Using custom image for KES
 	if encryptionCfg.Image != "" {
 		kesConfiguration.Image = encryptionCfg.Image
 	}
-	// Generate server certificates for KES only if autoCert is disabled
-	if autoCert == nil || (autoCert != nil && !*autoCert) {
+	// Generate server certificates for KES
+	if encryptionCfg.Server != nil {
```
cesnietor, line 207: if this is nil should we return any error or nah?

restapi/admin_tenants.go
Alevsk: We are checking if user provided encryption configuration and also if its possible to enable encryption (because Tenants will be deployed using TLS via AutoCert or External certificates)

restapi/admin_tenants_helper.go
Alevsk: `encryptionCfg.Server` are External certificates for KES, in case this is `nil` KES will still be deployed using `AutoCert` certificates

cesnietor (approved):

PR #453, opened 2020-11-27
Add Dockerfile based on Red Hat UBI

nitisht:


harshavardhana (approved):

Dockerfile.release.ubi
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:18.04 as certs
+
+RUN apt-get update -y && apt-get install -y ca-certificates
+
```
harshavardhana, line 4: We don't need this @nitisht `microdnf install ca-certificates` is enough at single layer

harshavardhana (changes requested):

Dockerfile.release.ubi
nitisht: done

cesnietor (approved):

bexsoft (approved):

nitisht:
Can we merge this pls?

harshavardhana (approved):

Issue #388, opened 2020-11-12
Files in object browser - subfolders are downloaded with full path in name


bexsoft:
Files in object browser - subfolders are downloaded with incorrect names, they contain the full path before the correct name

klauspost:
This could be a likely solution: https://github.com/minio/minio/pull/10823

PR #466, opened 2020-12-03
Fix object download

cesnietor:
fixes https://github.com/minio/console/issues/388
If an object is within a folder the object downloaded now only has the object's name.
Also, it now supports object version downloading.


bexsoft (approved):
LGTM, Test OK

Alevsk (approved):

PR #473, opened 2020-12-04
import minio/pkg/certs as xcerts

harshavardhana:
update dependency to fix windows cert issues,
now handled properly by the pkg/certs package
in MinIO.


Alevsk (approved):
here the issue was the alias in the package name?

cesnietor (approved):

PR #480, opened 2020-12-07
Remove alpine base and use UBI as the default base image

nitisht:


Dockerfile.release
```diff
@@ -1,12 +1,21 @@
-FROM ubuntu:18.04 as certs
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.3
 
-RUN apt-get update -y && apt-get install -y ca-certificates
+COPY CREDITS /licenses/CREDITS
+COPY LICENSE /licenses/LICENSE
 
-FROM scratch
-MAINTAINER MinIO Development "dev@min.io"
```
Alevsk, line 6: @nitisht my understanding is scratch image will have only the console binary and nothing else while `registry.access.redhat.com/ubi8/ubi-minimal:8.3` will have some more binaries, including package managers, "increasing" the attack surface, @dvaldivia do we want to move from scratch to RedHat images?

Alevsk (dismissed):

Dockerfile.release
harshavardhana: @abperiasamy we decided that we should move all our family containers to redhat supported images.

The reason is we need to certify all our containers to run in OpenShift environment. Using scratch container is not acceptable and having ways to build container is also not acceptable.

We are also going to add multiarch build automatically making console containers multiarch by default.

harshavardhana (approved):

dvaldivia:
Do we really need any of the libs that `ubi-minimal` brings? scratch seems like the ideal bare minimum for our needs cc @harshavardhana @nitisht

harshavardhana:
> Do we really need any of the libs that `ubi-minimal` brings? scratch seems like the ideal bare minimum for our needs cc @harshavardhana @nitisht

@dvaldivia as already indicated we need this for certification of our containers. Scratch is not supported and we don't want multiple different releases of console container. 

UBI is the hardened REDHAT base image with required support from them for our enterprise customers.

UBI base image is standardized across all our projects now and console is nothing special from that point of view.

Alevsk (approved):

Alevsk:
> > Do we really need any of the libs that `ubi-minimal` brings? scratch seems like the ideal bare minimum for our needs cc @harshavardhana @nitisht
> 
> @dvaldivia as already indicated we need this for certification of our containers. Scratch is not supported and we don't want multiple different releases of console container.
> 
> UBI is the hardened REDHAT base image with required support from them for our enterprise customers.
> 
> UBI base image is standardized across all our projects now and console is nothing special from that point of view.

If this is mandatory because of the certification then I think its find

dvaldivia (approved):

harshavardhana:
With this approach, we are also able to now ship console as multi-arch manifest container i.e `docker pull minio/console` on ARM pulls the ARM binary instead of amd64

PR #507, opened 2020-12-11
Fix all react warnings

dvaldivia:
Fixes all the outstanding warnings that react complains about.

Now `yarn build` completes without a complaint

Screen Shot 2020-12-11 at 12 56 36 PM

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ObjectDetails/ObjectDetails.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ObjectDetails/ObjectDetails.tsx
@@ -194,7 +194,7 @@ const ObjectDetails = ({
           setLoadObjectData(false);
         });
     }
-  }, [loadObjectData]);
+  }, [loadObjectData, bucketName, pathInBucket]);
 
   let tagKeys: string[] = [];
 
@@ -341,6 +341,11 @@ const ObjectDetails = ({
       )}
       <Grid container>
         <Grid item xs={12} className={classes.container}>
+          {error !== "" && (
+            <Grid item xs={12}>
+              {error}
+            </Grid>
+          )}
           <Grid item xs={12} className={classes.obTitleSection}>
             <div>
               <BrowserBreadcrumbs />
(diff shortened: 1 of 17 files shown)


portal-ui/src/screens/Console/Users/AddUser.tsx
```diff
@@ -67,14 +67,35 @@ const AddUserContent = ({
   const [currentGroups, setCurrentGroups] = useState<string[]>([]);
 
   useEffect(() => {
+    const getUserInformation = () => {
```
bexsoft, line 70: We can remove this function (It will be re-defined on each `selectedUser` change). we can just place the api call inside the else block on line 95.

bexsoft (changes requested):

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ObjectDetails/ObjectDetails.tsx
```diff
@@ -341,6 +341,11 @@ const ObjectDetails = ({
       )}
       <Grid container>
         <Grid item xs={12} className={classes.container}>
+          {error !== "" && (
+            <Grid item xs={12}>
```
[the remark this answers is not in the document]
dvaldivia: This is how we display errors in so many places XD do we have a component for errors?

portal-ui/src/screens/Console/Users/AddUser.tsx
dvaldivia: it's a constant so it should be fine, it's never re-used so this is the recommended way

dvaldivia: I'll make it a callback anyway

bexsoft:
I found another issue that seems not to be related to this change but must be fixed (#512)

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ObjectDetails/ObjectDetails.tsx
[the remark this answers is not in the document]
bexsoft: Not yet, but right now we are using a class called `errorBlock`

bexsoft (approved):

cesnietor (approved):
tested LGTM

PR #517, opened 2020-12-15
Added EC Parity correct values to add tenant modal

### LINKED ISSUE #411: When sizing tenant ask for more detailed information
We should ask the columns on the left of this image, and offer a preview of the calculated values on the right

Screen Shot 2020-11-18 at 6 54 53 PM

### LINKED ISSUE #500: Fix sizing step when adding a tenant
We have a problem where adding a tenant in the sizing step we don't make a lot of sense, not even showing EC, so we need to rework this screen

bexsoft:
fixes #500 
fixes #411 

## What does this do?

Adds correct EC information & calculation in create tenant screen

## How does it look?

Screen Shot 2020-12-14 at 22 28 44
Screen Shot 2020-12-14 at 22 28 36


dvaldivia (approved):

portal-ui/src/common/utils.ts
```diff
@@ -324,6 +352,67 @@ const structureCalc = (
...
+
+  const totalStorage = totalDisks * pvSize;
+  const maxEC = parityValidValues[0];
+  const maxParityNumber = parseInt(maxEC.split(":")[1], 10);
+
+  const erasureStripeSet = maxParityNumber * 2; // ESS is calculated by multiplying maximum parity by two.
+
+  const storageFactors: IStorageFactors[] = parityValidValues.map(
+    (currentParity) => {
+      const parityNumber = parseInt(currentParity.split(":")[1], 10);
```
Alevsk, line 382: `currentParity.split(":")[1]`

here we are always sure `currentParity` will be in the format "item:item" ? otherwise we may end with a npe

Alevsk (approved):

portal-ui/src/common/utils.ts
bexsoft: Yes, we have total control of the value returned from this API:

Screen Shot 2020-12-15 at 11 52 36

PR #524, opened 2020-12-17
add warning when active directory has insecure connection

Alevsk:
Screen Shot 2020-12-16 at 6 12 42 PM


bexsoft (approved):
LGTM

harshavardhana:
Perhaps better to say "security: all traffic to Active Directory will be plain-text"

dvaldivia (approved):

Issue #529, opened 2020-12-18
Cannot create users


kevinlul:
I'm running the last tag from [Docker Hub](https://hub.docker.com/r/minio/console/tags?page=1&ordering=last_updated) (v0.4.6).

After setting up and signing in through the browser, I can add and remove policies, add and remove groups, set policies on users and groups, and remove users. However, I cannot add users. After clicking on "Create User" and filling in a set of valid credentials that would be able to be added through `mc`, I get "An error occurred, please try again." Looking in the logs, the original error is "The account key is not eligible for this operation"

I understand that this is still under development but I can't find a list of implemented features so I'm opening this issue.

cesnietor:
might be related to https://github.com/minio/minio/pull/10978

harshavardhana:
> After setting up and signing in through the browser, I can add and remove policies, add and remove groups, set policies on users and groups, and remove users. However, I cannot add users. After clicking on "Create User" and filling in a set of valid credentials that would be able to be added through `mc`, I get "An error occurred, please try again." Looking in the logs, the original error is "The account key is not eligible for this operation"

@kevinlul upgrade your MinIO server.

kevinlul:
@harshavardhana I am on the latest from Docker Hub. I just spun this up today.
```
minio version RELEASE.2020-12-18T03-27-42Z
console version 0.4.6 - d573007
```

kevinlul:
How come using the same account in `mc` works for adding a new user but not console? Are they not using the same API on MinIO?

dvaldivia:
@kevinlul yes, they use the same library as well,`mc`. We need to debug where this bug is happening

harshavardhana:
> How come using the same account in `mc` works for adding a new user but not console? Are they not using the same API on MinIO?

Console uses a different mechanism @kevinlul

harshavardhana:
New release of MinIO should fix this now RELEASE.2020-12-23T02-24-12Z

kevinlul:
It works fantastic! Thank you 🚀

PR #530, opened 2020-12-19
Update to latest minio version and fix updates

cesnietor:
This fixes issue were users couldn't be created (https://github.com/minio/console/issues/529) as well as updates Config
related apis to return whether server needs to restart after setting a config or not.

To test this, you can test changing user password and adding new users.


dvaldivia (dismissed):

harshavardhana (approved):
NOTE: notification settings might also become dynamic and non-restart friendly in future.

harshavardhana (approved):

dvaldivia (approved):

PR #543, opened 2021-01-07
Add Diagnostic Page on Console

cesnietor:
Adds  HealthInfo Api and integrates it with Console UI.
Since The diagnostic can take a while, the websocket connection is not closed until the diagnosis is finished, some messages are being shown to show status. Also once the diagnosis is done the download button should be enabled.
Disclaimer: if the page is refreshed the previous diagnosis will be lost and you will have to run it again, you can still navigate during diagnosis through other pages.
Default deadline for Diagnosis is `1hr`.
- This also adds prettier to Makefile

Reqs: if prettier is not installed please do `cd portal-ui/; yarn add --dev --exact prettier;`

Screen Shot 2021-01-08 at 7 46 46 PM
Screen Shot 2021-01-07 at 8 52 01 PM
Screen Shot 2021-01-08 at 7 46 38 PM

- [X] Tests included.


portal-ui/src/actions.ts
```diff
@@ -72,3 +73,10 @@ export const setSnackBarMessage = (message: string) => {
     snackBarMessage: message,
   };
 };
+
+export const setServerDiagStat = (status: string) => {
+  return {
+    type: SET_SERVER_DIAG_STAT,
+    serverDiagnosticStatus: status,
+  }
+}
```
bexsoft, line 82: Please run prettier on files (version 2.1.1)

bexsoft (changes requested):

restapi/admin_health_info.go
```diff
@@ -0,0 +1,94 @@
...
+		// return if context ends
+		case <-ctx.Done():
+			return nil
+		case adminHealthInfo, ok := <-healthChan:
+			// zero value returned because the channel is closed and empty
+			if !ok {
+				return nil
+			}
+			if adminHealthInfo.Error != "" {
+				return errors.New(adminHealthInfo.Error)
```
Alevsk, line 68: Since this errors will be returned directly to the client via `sendWsCloseMessage(wsc.conn, err)` should we return a more generic error using the prepareError function? to avoid any kind of sensitive information exposure via an error stacktrace or something

Alevsk (changes requested):

restapi/admin_health_info.go
cesnietor: Agree, done. Solved on ws_handle.go - sendWsCloseMessage()
Generic error is returned and the original error is being logged, I didn't use prepareError since we check other type of errors in there.

dvaldivia (approved):

Alevsk (approved):
LGTM

bexsoft (approved):
LGTM, Please merge after #549

bexsoft (approved):
LGTM

Alevsk (approved):

PR #554, opened 2021-01-15
Allow arbitrary number of file uploads

cesnietor:
Now you can select multiple files on Console UI. 
Note: Folder upload is still not supported also a timeout may occur (with large files) which will be addressed in future prs.

Parameter definition for file upload on swagger.yaml was removed
since go-swagger doesn't support multiple upload of files. Implementation
was done instead on  user_objects.go file.

To test: Select multiple files on Console UI and upload should happen without any errors (on bucket, on subfolders, etc.)


Alevsk (approved):
LGTM

dvaldivia (approved):

PR #574, opened 2021-01-24
fix: turn-off TLS redirects if configured

### LINKED ISSUE #428: API to Set object locking for an object
https://docs.minio.io/docs/minio-bucket-object-lock-guide.html

### LINKED ISSUE #573: Console redirects to TLS port/host even if MCS_SECURE_SSL_REDIRECT is set to "off"
In our environment we are using HaProxy for load balancing and https connections to Kubernetes where incoming requests are handled via Istio ingress.

Due to this we have an Istio ingress pointing to the minio console service on port 9090. Our expectation is that we can terminate Https traffic on HaProxy and all other traffic uses HTTP.

Unfortunately, requests to the console are always redirected to 0.0.0.0:9443 although MCS_SECURE_SSL_REDIRECT is set to "off" in the ENV variables of the console's PODs.

Hence, we cannot access the console...

Our versions used:
krew v0.4.0
minio operator v3.0.29
minio console v0.5.2 (changed manually)

harshavardhana:
fixes #573

### DIFF
--- a/cmd/console/server.go
+++ b/cmd/console/server.go
@@ -58,6 +58,11 @@ var serverCmd = cli.Command{
 			Usage: "HTTPS server port",
 		},
 		cli.StringFlag{
+			Name:  "tls-redirect",
+			Value: restapi.GetTLSRedirect(),
+			Usage: "HTTPS redirect by default",
+		},
+		cli.StringFlag{
 			Name:  "certs-dir",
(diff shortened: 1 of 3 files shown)


Alevsk (dismissed):
LGTM

bexsoft (approved):
LGTM

dvaldivia (approved):

PR #576, opened 2021-01-26
Support for adding ExternalCaCert secrets

Alevsk:
Screen Shot 2021-01-25 at 18 28 48

### DIFF
--- a/portal-ui/src/screens/Console/Tenants/ListTenants/AddTenant.tsx
+++ b/portal-ui/src/screens/Console/Tenants/ListTenants/AddTenant.tsx
@@ -253,6 +253,15 @@ const AddTenant = ({
       encoded_cert: "",
     },
   ]);
+  const [caCertificates, setCaCertificates] = useState<KeyPair[]>([
+    {
+      id: Date.now().toString(),
+      key: "",
+      cert: "",
+      encoded_key: "",
+      encoded_cert: "",
+    },
+  ]);
   const [consoleKeyVal, setConsoleKeyVal] = useState<string>("");
   const [consoleCertVal, setConsoleCertVal] = useState<string>("");
   const [serverKeyVal, setServerKeyVal] = useState<string>("");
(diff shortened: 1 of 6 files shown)


bexsoft (dismissed):
LGTM, there is a small alignment issue with MinIO certificates selector that is not related to this change. will file a new ticket to fix it.

bexsoft:
Ticket created: https://github.com/minio/console/issues/577

portal-ui/src/screens/Console/Tenants/ListTenants/AddTenant.tsx
```diff
@@ -941,6 +992,16 @@ const AddTenant = ({
 
       let tenantCerts: any = null;
       let consoleCerts: any = null;
+      let caCerts: any = null;
+
+      if (caCertificates.length > 0) {
```
cesnietor, line 997: is this always true? I see that it's being always initialized with one element, also I see that for delete certificates the length is evaluated as > 1, just checking.
```
 const deleteCaCertificate = (id: string) => {
    if (caCertificates.length > 1) {
```

restapi/admin_tenants.go
```diff
@@ -701,7 +701,29 @@ func getTenantCreatedResponse(session *models.Principal, params admin_api.Create
...
+				Name: fmt.Sprintf("ca-certificate-%d", i),
+				Content: map[string][]byte{
+					"public.crt": certificateContent,
+				},
+			})
+		}
+		if len(caCertificates) > 0 {
+			certificateSecrets, err := createOrReplaceSecrets(ctx, &k8sClient, ns, caCertificates, tenantName)
+			if err != nil {
+				return nil, prepareError(err)
```
cesnietor, line 722: do we want here to return the error or do `prepareError(errorGeneric, nil, err)` as above?

portal-ui/src/screens/Console/Tenants/ListTenants/AddTenant.tsx
Alevsk: yeah, `caCertificates` is an array that will contain always 1 element (so we render 1 certificate entry), user can append more certificates to the array, the user then can delete elements from the start, the middle or the end of the array until the array reaches 1 element, if that's the case we prevent deleting more elements (because we always need to have 1)

cesnietor (approved):

bexsoft (approved):

PR #579, opened 2021-01-27
Changed default TLS redirect to off when no configuration is provided

bexsoft:
## What does this do?

Fixes an infinite reload issue when no TLS configuration is set 

**NOTE:** You may need to clear your cache in order to test this PR


harshavardhana (approved):

bexsoft:
Will close as TlS must be enabled by default. This will be handled in #580

PR #614, opened 2021-02-25
Disable Users and Groups Menu options when LDAP is enabled on MinIO

Alevsk:
## Without LDAP enabled

Screen Shot 2021-02-25 at 10 51 57

## With LDAP enabled

Screen Shot 2021-02-25 at 10 51 11

### DIFF
--- a/pkg/acl/const.go
+++ b/pkg/acl/const.go
@@ -18,4 +18,6 @@ package acl
 
 const (
 	consoleOperatorMode = "CONSOLE_OPERATOR_MODE"
+	// const for ldap configuration
+	ConsoleLDAPEnabled = "CONSOLE_LDAP_ENABLED"
 )
--- a/restapi/user_service_accounts.go
+++ b/restapi/user_service_accounts.go
@@ -58,7 +58,9 @@ func registerServiceAccountsHandlers(api *operations.ConsoleAPI) {
 
 // createServiceAccount adds a service account to the userClient and assigns a policy to him if defined.
 func createServiceAccount(ctx context.Context, userClient MinioAdmin, policy string) (*models.ServiceAccountCreds, error) {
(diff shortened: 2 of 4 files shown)


pkg/acl/const.go
```diff
@@ -18,4 +18,6 @@ package acl
 
 const (
 	consoleOperatorMode = "CONSOLE_OPERATOR_MODE"
+	// const for ldap configuration
+	ConsoleLDAPEnabled = "CONSOLE_LDAP_ENABLED"
```
cesnietor, line 22: seems this constant is already declared on `pkg/auth/ldap` on `const.go` should we use that instead?

Alevsk: the variables are on different namespaces, i don't want to have dependency from that namespace to this one because this can quickly end on circular dependencies x.x

restapi/user_service_accounts.go
```diff
@@ -58,7 +58,7 @@ func registerServiceAccountsHandlers(api *operations.ConsoleAPI) {
 
 // createServiceAccount adds a service account to the userClient and assigns a policy to him if defined.
 func createServiceAccount(ctx context.Context, userClient MinioAdmin, policy string) (*models.ServiceAccountCreds, error) {
-	iamPolicy := &iampolicy.Policy{}
+	var iamPolicy *iampolicy.Policy
```
harshavardhana, line 61: Add a comment on why it is kept this way? to ensure that we do not initialize `iampolicy.Policy` struct since that is a valid policy?

harshavardhana (changes requested):

harshavardhana (approved):

bexsoft (approved):
LGTM

cesnietor (approved):

Issue #594, opened 2021-02-05
UI hide version box on non versioned buckets


hijak:
Hi

when clicking the file in the ui it will open a window with the versioning information. this appears also on non versioned buckets and produces an "error occurred" box at the top.

Can we hide the versioning box so that does not query the api that isnt active on the bucket?

bexsoft:
Hi @hijak! Would you please provide us which version of Console are you using? Thank you

hijak:
hey that would be version:

```

PR #637, opened 2021-03-09
Show object versions on UI only when bucket has versioning enabled

cesnietor:
fixes https://github.com/minio/console/issues/594
If bucket doesn't have versioning enabled, objects shouldn't show component that lists versions.

Screen Shot 2021-03-09 at 2 43 02 PM
Screen Shot 2021-03-09 at 2 45 35 PM

To test:
Create a bucket without versioning, add an object and on the object details the list shouldn't appear.
Create a bucket with versioning, add an object and on the object details the list should appear.


dvaldivia (approved):

dvaldivia (approved):

bexsoft (approved):
I can't access Buckets Browser because of an issue with match var from route (issue is in master), If I go to the bucket path this PR works as expected. Will fix the other issue in another ticket

cesnietor:
> I can't access Buckets Browser because of an issue with match var from route (issue is in master), If I go to the bucket path this PR works as expected. Will fix the other issue in another ticket

got it thanks.

PR #647, opened 2021-03-17
License page and License verification 

Alevsk:
## License Page

- Add `ref=op` to all `<a></a>` elements in the License Page
- If License is activated in Operator UI, change the label from Subscription to Upgrade, when the user click it, it will be redirected to the `subnet subscription management` page
- Add Refresh License button to the License Page, when user click the button we will pull a new license from `https://subnet.min.io/api/auth/subscription/license-key` based on the existing one, and the license will be updated on all Console tenants (investigate if we need to restart all this pods)
- Apply new design

## License verification

- When console starts verify current license passed via `CONSOLE_SUBNET_LICENSE` env variable is still valid against subnet `https://subnet.min.io/api/auth/subscription/license-key` endpoint
- Launch a go routine that will verify every X amount of time current license is valid

## How it looks (Operator screens)

Screen Shot 2021-03-18 at 15 29 04

Screen Shot 2021-03-18 at 14 59 02

Screen Shot 2021-03-18 at 14 59 09

## How it looks (Console Admin screens)

Screen Shot 2021-03-18 at 15 24 24

Screen Shot 2021-03-18 at 15 26 12


nagashreem:
Some comments on the UI:
> If License is activated in Operator UI, change the label from Subscription to Upgrade, 

There is no Subscription Button on the first screenshot
image

In the second screenshot, the exclamation mark seems to have moved to the second line
image

Also showing the button label as Subscribe might be confusing here (unless we show Subscribe button in the previous state - 1st screenshot)
image

pkg/subnet/subnet.go
```diff
@@ -171,3 +206,18 @@ func ValidateLicense(client cluster.HTTPClientI, licenseKey, email, password str
...
 	return licInfo, license, nil
 }
+
+func RefreshLicense(client cluster.HTTPClientI, licenseKey string) (licInfo *licverifier.LicenseInfo, license string, err error) {
+	if licenseKey != "" {
+		license, err = getNewLicenseFromExistingLicense(client, licenseKey)
+		if err != nil {
+			return nil, "", err
+		}
+		licenseInfo, rawLicense, err := ValidateLicense(client, license, "", "")
```
krisis, line 216: Do we need to validate the license received from Subnet?

krisis (approved):
The license refresh part looks good to me.

dvaldivia (approved):

krisis (approved):

pkg/subnet/subnet.go
Alevsk: @krisis hi kp, this is used to parse the jwt token string into the license structure

bexsoft (approved):

PR #650, opened 2021-03-17
[WIP] Hash user passwords using argon2

Alevsk:
Added support for hashing user secret key with argon2 and then store it inside the
encrypted session token, necessary for current password comparison (prevent account take over vulnerabilities) during the
change-password process

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -21,8 +21,10 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/hmac"
+	"crypto/rand"
 	"crypto/sha1"
 	"crypto/sha256"
+	"crypto/subtle"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -32,6 +34,8 @@ import (
 	"log"
 	"net/http"
 	"strings"
+
+	"golang.org/x/crypto/argon2"
 
 	"github.com/go-openapi/swag"
 	"github.com/minio/console/models"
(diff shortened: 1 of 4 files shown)


pkg/auth/token.go
```diff
@@ -333,3 +337,55 @@ func GetClaimsFromTokenInRequest(req *http.Request) (*models.Principal, error) {
...
+	format := "$argon2id$v=%d$m=%d,t=%d,p=%d$%s$%s"
+	full := fmt.Sprintf(format, argon2.Version, argon2Config.memory, argon2Config.time, argon2Config.threads, b64Salt, b64Hash)
+	return full, nil
+}
+
+// ComparePassword is used to compare a user-inputted password to a hash to see if the password matches or not.
+func ComparePassword(password, hash string) (bool, error) {
+	parts := strings.Split(hash, "$")
+	c := &PasswordConfig{}
+	_, err := fmt.Sscanf(parts[3], "m=%d,t=%d,p=%d", &c.memory, &c.time, &c.threads)
```
aead, line 376: We must check that are actually 3 parts. Ideally use `stringsSplitN` and check that returned `[]string` has exactly `4` elements

restapi/user_account_test.go
```diff
@@ -85,7 +85,7 @@ func Test_changePassword(t *testing.T) {
 				ctx:    context.Background(),
 				session: &models.Principal{
 					AccountAccessKey: "TESTTEST",
-					AccountSecretKey: "TESTTEST123",
+					AccountSecretKey: "$argon2id$v=19$m=65536,t=1,p=4$npAQ6LeqDsPMEQEIl7GnDA$cNasELqD07M78DWgmtVS2ilDu1GZioPkstpDdi0cFPA",
 				},
```
aead, line 89: Maybe also add one or two tests with invalid / malformed password hashes?

pkg/auth/token.go
```diff
@@ -333,3 +337,55 @@ func GetClaimsFromTokenInRequest(req *http.Request) (*models.Principal, error) {
...
+// argon2Config configuration for the argon2 hashing function
+var argon2Config = &PasswordConfig{
+	time:    1,
+	memory:  64 * 1024,
+	threads: 4,
+	keyLen:  32,
+}
+
+// NewPasswordHash will receive a password and hash it using argon2
+func NewPasswordHash(password string) (string, error) {
```
aead, line 357:
```suggestion
func HashPassword(password string) (string, error) {
```

Alevsk:
@aead thanks for the review, at the end this change was not need it, sometimes the solution is not to add more code but instead to remove it :D

PR #656, opened 2021-03-18
Added bucket-policy

adfost:
Added a function to list policies that apply to a bucket


admin.json
```diff
@@ -0,0 +1,21 @@
+{
```
dvaldivia, line 1: don't check in this file

go.mod
```diff
@@ -3,18 +3,20 @@ module github.com/minio/console
 go 1.16
 
 require (
+	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d // indirect
 	github.com/blang/semver/v4 v4.0.0
 	github.com/cheggaaa/pb/v3 v3.0.6
 	github.com/coreos/go-oidc v2.2.1+incompatible
-	github.com/go-openapi/errors v0.19.6
+	github.com/go-openapi/errors v0.20.0
```
dvaldivia, line 10: perhaps these dependency changes are not needed since this PR is not about upgradring the swagger version

portal-ui/src/screens/Console/Buckets/ViewBucket/ViewBucket.tsx
```diff
@@ -261,6 +264,22 @@ const ViewBucket = ({
     }
   }, [loadingReplication, setErrorSnackMessage, bucketName]);
 
+  /*useEffect(() => {
```
dvaldivia, line 267: if the UI integration is incomplete, don't send those file in your PR until it's done

dvaldivia (changes requested):

PR #661, opened 2021-03-20
Add Bucket Locking option and fix versioning configuration.

dvaldivia:
We used to turn locking for any bucket with versioning features, not we only turn it on if there's retention configured or explicitly on, this means that we can have buckets with versioning enabled only

Screen Shot 2021-03-19 at 9 39 33 PM


portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket.tsx
```diff
@@ -1,4 +1,5 @@
 // This file is part of MinIO Console Server
+// This file is part of MinIO Console Server
```
bexsoft, line 2: Please remove duplicated line

bexsoft (changes requested):
Please fix the above, also it would be nice to show the object locking status in the bucket details screen
Screen Shot 2021-03-22 at 12 49 58

belvid:
Hi!

The 3 lines below features should have a lighter gray color, since those are dividers and not actionable (as the input line above). I would suggest to use ##eaeaea for the lines as dividers (and apply the same to all similar cases)

portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket.tsx
dvaldivia: what a weird accident

bexsoft:
> Hi!
> 
> The 3 lines below features should have a lighter gray color, since those are dividers and not actionable (as the input line above). I would suggest to use ##eaeaea for the lines as dividers (and apply the same to all similar cases)

Sure! I will file a new ticket for it

adfost:
Are the json changes necessary?

bexsoft:
> Are the json changes necessary?

In this case, yes, as these files are auto generated by `make assets`

dvaldivia:
@bexsoft done

Screen Shot 2021-03-22 at 1 03 29 PM

dvaldivia:
Object Locking Enabled
Screen Shot 2021-03-22 at 1 43 09 PM


Object Locking Disabled

Screen Shot 2021-03-22 at 1 43 20 PM

dvaldivia:
Screen Shot 2021-03-22 at 2 23 15 PM

Screen Shot 2021-03-22 at 2 23 20 PM

bexsoft (approved):
LGTM

Alevsk (approved):
LGTM

PR #665, opened 2021-03-24
Added replication bulk modal to buckets list

bexsoft:
## What does this do?
Implements replication modal for bulk edit

## How does it look?
Screen Shot 2021-03-31 at 22 42 06
Screen Shot 2021-03-31 at 22 41 59
Screen Shot 2021-03-31 at 22 41 52
Screen Shot 2021-03-31 at 22 41 45


dvaldivia (approved):

Alevsk:
There's a weird bug in which every time I type a letter the input field lost focus
Screen Shot 2021-04-02 at 15 40 21

I could not pass this step, even if the buckets exist on the remote cluster im still getting the same error
Screen Shot 2021-04-02 at 15 41 47

Alevsk (changes requested):
There's a weird bug in which every time I type a letter the input field lost focus
Screen Shot 2021-04-02 at 15 40 21

I could not pass this step, even if the buckets exist on the remote cluster im still getting the same error
Screen Shot 2021-04-02 at 15 41 47

Alevsk:
Also root credentials dont seem to work for bucket replication, maybe we should put a message indicating that to the user

dvaldivia:
@Alevsk I think we forbid using root credentials for a lot of things, even STS, perhaps we should add that in the readme of console as well

dvaldivia:
only two things:

Still see the double scrollbar:

Screen Shot 2021-04-05 at 8 58 32 AM

and I don't see any loading indicating that blocks the UI while I wait on this other step:

Screen Shot 2021-04-05 at 9 07 27 AM

Alevsk (dismissed):

dvaldivia (approved):

Alevsk (approved):

PR #668, opened 2021-03-25
UI for Bucket Policy and API for Users with Access to Bucket

adfost:


admin.json
```diff
@@ -0,0 +1,21 @@
+{
```
dvaldivia, line 1: remove this file

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
dvaldivia, line 1: remove this file

dvaldivia (changes requested):

PR #671, opened 2021-03-26
Adding API for Users with Access to Bucket

adfost:


restapi/admin_users.go
```diff
@@ -467,3 +474,31 @@ func getAddUsersListToGroupsResponse(session *models.Principal, params admin_api
...
+	ctx := context.Background()
+	mAdmin, err := newMAdminClient(session)
+	if err != nil {
+		return nil, prepareError(err)
+	}
+	// create a minioClient interface implementation
+	// defining the client to be used
+	adminClient := adminClient{client: mAdmin}
+
+	users, err := listUsers(ctx, adminClient)
```
dvaldivia, line 488: Think this doesn't cover the groups a user belongs to which might also include some policies

PR #708, opened 2021-04-27
Improvements to Built-In users

dvaldivia:
Signed-off-by: Daniel Valdivia <hola@danielvaldivia.com>

This PR makes a few tweaks to the built-in users list to have a more dense list.

Also this makes it so the storage class default for logs and prometheus is the same as the tenant storage class by default

Screen Shot 2021-04-27 at 9 29 44 AM


bexsoft (approved):

adfost:
The changes seem ok, but I am getting the error that it's not letting me select a storage class again ...

adfost:
Ok I got it to work, this seems fine, it's ok having multiple add buttons right?

adfost (approved):

PR #732, opened 2021-05-11
Fix NodeSelector Screen

dvaldivia:
This makes it so the node selector logic is different from podAffinity

Signed-off-by: Daniel Valdivia <hola@danielvaldivia.com>


Alevsk (approved):
Tested, works fine

PR #744, opened 2021-05-14
User policies

adfost:


dvaldivia:
you should probably squash into a single commit

dvaldivia:
Can you also attach screenshots of the resulting change?

dvaldivia (changes requested):
While trying to run the server I get the error
```
2021-05-15 18:42:26.514561 I | 2021/05/15 18:42:26 server.go:97: invalid character '<' looking for beginning of object key string
```
you need to regen swagger and commit the files that are generated

admin.json
```diff
@@ -0,0 +1,22 @@
+{
```
dvaldivia, line 1: Don't commit this file

portal-ui/src/screens/Console/Buckets/ListBuckets/ListBuckets.tsx
```diff
@@ -97,8 +97,9 @@ const ListBuckets = ({
   const [loadingPerms, setLoadingPerms] = useState<boolean>(true);
   const [canCreateBucket, setCanCreateBucket] = useState<boolean>(false);
   const [selectedBuckets, setSelectedBuckets] = useState<string[]>([]);
-  const [replicationModalOpen, setReplicationModalOpen] =
-    useState<boolean>(false);
+  const [replicationModalOpen, setReplicationModalOpen] = useState<boolean>(
```
dvaldivia, line 100: update your prettier to `2.3.0` to avoid having this phantom changes

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,380 @@
...
+            value={selectedTab}
+            indicatorColor="primary"
+            textColor="primary"
+            onChange={(_, newValue: number) => {
+              setSelectedTab(newValue);
+            }}
+            aria-label="policy-tabs"
+          >
+            <Tab label="Users" />
+            <Tab label="Edit Policy" />
```
dvaldivia, line 280: I think it's better to move `Edit Policy` to the first place of the tabs, and rename it to `Details` as it's the actual policy

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,380 @@
...
+            <form
+              noValidate
+              autoComplete="off"
+              onSubmit={(e: React.FormEvent<HTMLFormElement>) => {
+                addRecord(e);
+              }}
+            >
+              <Grid container>
+                <Grid item xs={12} className={classes.formScrollable}>
+                  <Grid item xs={12}>
```
dvaldivia, line 315: I don't think this is needed anymore, as the header of the page already container the name of the policy, remove this hole grid item

restapi/admin_policies.go
```diff
@@ -82,6 +82,13 @@ func registersPoliciesHandler(api *operations.ConsoleAPI) {
 		}
 		return admin_api.NewListPoliciesWithBucketOK().WithPayload(policyResponse)
 	})
+	api.AdminAPIListUsersForPolicyHandler = admin_api.ListUsersForPolicyHandlerFunc(func(params admin_api.ListUsersForPolicyParams, session *models.Principal) middleware.Responder {
+		policyResponse, err := getListUsersForPolicyResponse(session, params.Policy)
```
dvaldivia, line 86: perhaps this variable should be called `policyUsersResponse` as it's more of a list of users for the policy

dvaldivia (changes requested):
There's a react rendering bug which causes the list of users to dance around like crazy 

ezgif-2-186fbdf46587

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,380 @@
...
+        </Grid>
+        <Grid item xs={12}>
+          <br />
+        </Grid>
+        <Grid item xs={12}>
+          {selectedTab === 0 && (
+            <TableWrapper
+              itemActions={[
+                {
+                  type: "delete",
```
dvaldivia, line 291: remove this action, we don't need to delete users from  this screen

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,380 @@
...
+            <Tab label="Edit Policy" />
+          </Tabs>
+        </Grid>
+        <Grid item xs={12}>
+          <br />
+        </Grid>
+        <Grid item xs={12}>
+          {selectedTab === 0 && (
+            <TableWrapper
+              itemActions={[
```
dvaldivia, line 289: Can you add the default action so clicking on any of this users takes us to the route `/users/:userName`

dvaldivia (changes requested):

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,380 @@
...
+            <Link to={"/policies"} className={classes.breadcrumLink}>
+              Policy
+            </Link>
+            {` > ${match.params["policyName"]}`}
+          </Fragment>
+        }
+      />
+      <Grid item xs={12} className={classes.container} />
+      <Grid container>
+        <Grid item xs={12}>
```
dvaldivia, line 263: it's ok to remove one of these grid items with <br />

dvaldivia:
Can you put these details inside a `<Paper>` container, so they sit over white background
Screen Shot 2021-05-17 at 2 19 48 PM

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,363 @@
...
+  useEffect(() => {
+    loadUsersForPolicy();
+    loadPolicyDetails();
+    if (policy) {
+      setPolicyName(policy.name);
+      setPolicyDefinition(
+        policy ? JSON.stringify(JSON.parse(policy.policy), null, 4) : ""
+      );
+    }
+  }, [policy]);
```
bexsoft, line 226: 5.- Process completes, as the policy value has changed, this effect is executed again (Go back to step 1)

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,363 @@
...
+      .catch((err) => {
+        setAddLoading(false);
+        setModalErrorSnackMessage(err);
+      });
+  };
+
+  useEffect(() => {
+    loadUsersForPolicy();
+    loadPolicyDetails();
+    if (policy) {
```
bexsoft, line 220: 4.- Policy here contains the same value as you have not waiting to all the process to complete

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,363 @@
...
+      })
+      .catch((err) => {
+        setAddLoading(false);
+        setModalErrorSnackMessage(err);
+      });
+  };
+
+  useEffect(() => {
+    loadUsersForPolicy();
+    loadPolicyDetails();
```
bexsoft, line 219: 1.- You load policy details

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,363 @@
...
+      })
+      .catch((err) => {
+        console.log("Error in loading users");
+      });
+  };
+  const loadPolicyDetails = () => {
+    api
+      .invoke("GET", `/api/v1/policy?name=${policyName}`)
+      .then((result: any) => {
+        setPolicy(result);
```
bexsoft, line 189: 3.- You set the values of the policies that you received (I think these values are not sorted), Also values will not be reflected during this render run

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,363 @@
...
+    api
+      .invoke("GET", `/api/v1/policies/${policyName}/users`)
+      .then((result: any) => {
+        setUserList(result);
+      })
+      .catch((err) => {
+        console.log("Error in loading users");
+      });
+  };
+  const loadPolicyDetails = () => {
```
bexsoft, line 185: 2.- Policy details is being requested

bexsoft (changes requested):

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
dvaldivia: +1

restapi/admin_policies.go
dvaldivia: +1

dvaldivia (changes requested):

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,355 @@
...
+}
+
+const styles = (theme: Theme) =>
+  createStyles({
+    buttonContainer: {
+      textAlign: "right",
+    },
+    multiContainer: {
+      display: "flex",
+      alignItems: "center" as const,
```
bexsoft, line 58: I think we can remove the `as const` conversion as this is only needed for external linked files. Please test if it is not required, if not we can remove

dvaldivia (dismissed):

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
dvaldivia: should be harmless to keep

bexsoft (approved):

bexsoft (changes requested):
.

portal-ui/src/screens/Console/Policies/ListPolicies.tsx
```diff
@@ -0,0 +1,232 @@
...
+            if (pa.name > pb.name) {
+              return 1;
+            }
+
+            if (pa.name < pb.name) {
+              return -1;
+            }
+
+            return 0;
+          });
```
Alevsk, line 108: Sort logic can be moved to the backend

Alevsk (approved):

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,349 @@
...
+  useEffect(() => {
+    loadUsersForPolicy();
+    loadPolicyDetails();
+    if (policy) {
+      setPolicyName(policy.name);
+      setPolicyDefinition(
+        policy ? JSON.stringify(JSON.parse(policy.policy), null, 4) : ""
+      );
+    }
+  }, [policy]);
```
bexsoft, line 235: Please split this effect in two parts because there is still a couple of re-renders present:
```suggestion
useEffect(() => {
    if(loadingUsers) {
api
        .invoke("GET", `/api/v1/policies/${policyName}/users`)
        .then((result: any) => {
          setUserList(result);
          setLoadingUsers(false);
        })
        .catch((err) => {
          console.log("Error in loading users");
        });
    }
    
    if(loadingPolicy) {
      api
        .invoke("GET", `/api/v1/policy?name=${policyName}`)
        .then((result: any) => {
          setPolicy(result);
          setLoadingPolicy(false);
        })
        .catch((err) => {
          console.log("Error in loading policy");
        });
    }
}, [loadingUsers, loadingPolicy, policyName]);

  useEffect(() => {
    if (policy) {
      setPolicyName(policy.name);
      setPolicyDefinition(
        policy ? JSON.stringify(JSON.parse(policy.policy), null, 4) : ""
      );
    }
  }, [policy]);
```

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,349 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import React, { useEffect, useState } from "react";
+import { Policy } from "./types";
+import { connect } from "react-redux";
+import get from "lodash/get";
```
bexsoft, line 19: You can remove this import as it is not being used
```suggestion

```

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,349 @@
...
+import { createStyles, Theme, withStyles } from "@material-ui/core/styles";
+import {
+  containerForHeader,
+  modalBasic,
+} from "../Common/FormComponents/common/styleLibrary";
+import Paper from "@material-ui/core/Paper";
+import Grid from "@material-ui/core/Grid";
+import { Button, LinearProgress, Typography } from "@material-ui/core";
```
bexsoft, line 27: You can remove Typography ad it is not being used
```suggestion
import { Button, LinearProgress } from "@material-ui/core";
```

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,349 @@
...
+import { User } from "../Users/types";
+import api from "../../../common/api";
+import PageHeader from "../Common/PageHeader/PageHeader";
+import { Link } from "react-router-dom";
+import {
+  setErrorSnackMessage,
+  setModalErrorSnackMessage,
+} from "../../../actions";
+import { Fragment } from "react";
+import InputBoxWrapper from "../Common/FormComponents/InputBoxWrapper/InputBoxWrapper";
```
bexsoft, line 40: You can remove this import as it not being used
```suggestion

```

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,349 @@
...
+        .invoke("GET", `/api/v1/policy?name=${policyName}`)
+        .then((result: any) => {
+          setPolicy(result);
+          setLoadingPolicy(false);
+        })
+        .catch((err) => {
+          console.log("Error in loading policy");
+        });
+    }
+  };
```
bexsoft, line 203: These two functions can be removed as queries are now part of the effect that is going to invoke them, an alternative method is to convert these two functions info a [useCallback](https://es.reactjs.org/docs/hooks-reference.html#usecallback) and add them to the dependency array of the effect

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -0,0 +1,349 @@
...
+      })
+      .catch((err) => {
+        setAddLoading(false);
+        setModalErrorSnackMessage(err);
+      });
+  };
+
+  useEffect(() => {
+    loadUsersForPolicy();
+    loadPolicyDetails();
```
bexsoft, line 228: These two can be removed ad the effect if going to perform these queries

bexsoft (changes requested):
You can see the differences of these changes by enabling the `Highlight updates when components render` checkbox in react tools inspector. 

Screen Shot 2021-05-19 at 13 23 23

You'll notice the amount of re-renders in the application (the longest time the boxes appear, more re-renders are happening)

Screen Shot 2021-05-19 at 13 25 20

dvaldivia (approved):

bexsoft (approved):

PR #772, opened 2021-05-27
Update WriteTimeout in server.go

peshoangelov:
increase the default of WriteTimeout in server.go in order to allow long downloads of huge files (>700M size)


dvaldivia (approved):

harshavardhana (changes requested):
This applies to all upload and downloads for objects @dvaldivia 

If we have an upper limit it won't work. We need to decouple this top level deadline.. 

Or tell users to use `mc` to download files that are bigger.

Any arbitrary timeout is not going to help..

restapi/server.go
```diff
@@ -98,7 +98,7 @@ type Server struct {
 	ListenLimit  int           `long:"listen-limit" description:"limit the number of outstanding requests"`
 	KeepAlive    time.Duration `long:"keep-alive" description:"sets the TCP keep-alive timeouts on accepted connections. It prunes dead TCP connections ( e.g. closing laptop mid-download)" default:"3m"`
 	ReadTimeout  time.Duration `long:"read-timeout" description:"maximum duration before timing out read of the request" default:"30s"`
-	WriteTimeout time.Duration `long:"write-timeout" description:"maximum duration before timing out write of the response" default:"60s"`
+	WriteTimeout time.Duration `long:"write-timeout" description:"maximum duration before timing out write of the response" default:"1200s"`
```
bexsoft, line 101: In this case, a modification in a swagger auto-generated field is being attempt, I think it would be better to look for a way to modify this value using swagger.yml

bexsoft (changes requested):

dvaldivia:
@peshoangelov I was testing/integrating your PR and it turns out, swagger supports you passing `--write-timeout 1200` when you start the server and that technically should do it

PR #781, opened 2021-06-02
allow console to listen on ipv6

harshavardhana:
also converge tls-host and host, because hostnames
have nothing to do with HTTP or HTTPs they are the
same for both HTTP and HTTPs. Deprecating the
older flag `--tls-host` but it will still be honored
as hidden flag.


harshavardhana:
Merging this since this blocks other work on minio/minio repo, please test out the master when you find time @dvaldivia @Alevsk

PR #782, opened 2021-06-03
remove various unexpected features in console

harshavardhana:
- Unix listeners are removed
- KeepAlive, IdleTimeout etc are removed
- Authorization logic is simplified


restapi/configure_console.go
```diff
@@ -215,30 +207,22 @@ func setupGlobalMiddleware(handler http.Handler) http.Handler {
...
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// prioritize authorization header and skip
-		if r.Header.Get("Authorization") != "" {
-			next.ServeHTTP(w, r)
+		token, err := auth.GetTokenFromRequest(r)
+		if err != nil && err != auth.ErrNoAuthToken {
+			http.Error(w, err.Error(), http.StatusUnauthorized)
 			return
 		}
-		tokenCookie, err := r.Cookie("token")
```
dvaldivia, line 229: we use to be header only authorization, but then we reverted to cookies as it was the only reliable way to authenticate connections being upgraded to web sockets, so we still need this

harshavardhana: We don't need it this entire code is moved into single function @dvaldivia

dvaldivia (approved):

dvaldivia (approved):

cmd/console/server.go
```diff
@@ -175,36 +175,37 @@ func StartServer(ctx *cli.Context) error {
...
-	go func() {
-		failedAttempts := 0
-		for {
-			if err := restapi.RefreshLicense(); err != nil {
-				log.Println(err)
-				failedAttempts++
-				// end license refresh after 3 consecutive failed attempts
-				if failedAttempts >= 3 {
-					return
+	if acl.GetOperatorMode() {
```
Alevsk, line 186: I was talking with Daniel, we need to discuss with Ab if we want to restrict this refresh license feature to only operator-mode, otherwise we need to remove this line

Alevsk (changes requested):

Alevsk (approved):

Alevsk (approved):

dvaldivia (approved):

PR #791, opened 2021-06-04
Move log coude out of auto-generated files

dvaldivia:
Need to move some of the new code into a separate location because `make swagger-gen` regenerates `restapi/server.go` every time 

### DIFF
--- a/restapi/server.go
+++ b/restapi/server.go
@@ -38,70 +38,42 @@ import (
 	"time"
 
 	"github.com/go-openapi/runtime/flagext"
+	"github.com/go-openapi/swag"
 	flags "github.com/jessevdk/go-flags"
+	"golang.org/x/net/netutil"
 
-	"github.com/minio/cli"
 	"github.com/minio/console/restapi/operations"
 )
 
 const (
 	schemeHTTP  = "http"
 	schemeHTTPS = "https"
+	schemeUnix  = "unix"
 )
 
-var defaultSchemes = []string{
-	schemeHTTP,
+var defaultSchemes []string
+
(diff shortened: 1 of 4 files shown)


restapi/server.go
```diff
@@ -38,70 +38,42 @@ import (
...
+	"golang.org/x/net/netutil"
 
-	"github.com/minio/cli"
 	"github.com/minio/console/restapi/operations"
 )
 
 const (
 	schemeHTTP  = "http"
 	schemeHTTPS = "https"
+	schemeUnix  = "unix"
```
harshavardhana, line 51: We should remove all of this..we can't add this back. We should find a way to turn this off in swagger.

harshavardhana (changes requested):

restapi/server.go
dvaldivia: let me look into it

restapi/server.go
```diff
@@ -38,70 +38,42 @@ import (
...
+var defaultSchemes []string
 
-func logInfo(msg string, data ...interface{}) {
-	infoLog.Printf(msg+"\n", data...)
-}
-
-func logError(msg string, data ...interface{}) {
-	errorLog.Printf(msg+"\n", data...)
+func init() {
+	defaultSchemes = []string{
```
dvaldivia, line 57: here you can see the `schemeUnix` is not activated @harshavardhana

harshavardhana (approved):

harshavardhana (approved):

bexsoft (approved):

PR #806, opened 2021-06-10
Move change user password modal

jinapurapu:
fixes https://github.com/miniohq/engineering/issues/158

**What does this do?**
Moved Change User Password feature to UserDetails page, located in an iconbutton accessed menu along with DeleteUser
 
**How does it look?**
Screenshot from 2021-06-10 14-10-37
Screenshot from 2021-06-10 14-10-52
Screenshot from 2021-06-10 14-11-08
Screenshot from 2021-06-10 14-11-27


portal-ui/src/screens/Console/Account/ChangeUserPasswordModal.tsx
```diff
@@ -67,9 +67,14 @@ const ChangeUserPassword = ({
       return;
     }
     setLoading(true);
+ 
+    if (newPassword.length < 8) {
+      setModalErrorSnackMessage("Passwords must be at least 8 characters long");
```
bexsoft, line 72: Please set loading to false after showing the error:
```suggestion
      setModalErrorSnackMessage("Passwords must be at least 8 characters long");
      setLoading(false);
```

portal-ui/src/screens/Console/Account/ChangePasswordModal.tsx
```diff
@@ -66,6 +66,11 @@ const ChangePassword = ({
       return;
     }
 
+    if (newPassword.length < 8) {
```
bexsoft, line 69: What is the difference between this component and `ChangeUserPasswordModal`?

bexsoft (changes requested):
Beside of that, LGTM

portal-ui/src/screens/Console/Account/ChangePasswordModal.tsx
dvaldivia: `ChangePasswordModal` is for the current user, which asks for `current password` and the other one is to change some other user's password

bexsoft: Ok, I think it would be better to merge these two components into a single one, and just use a flag to show / hide fields

bexsoft (approved):

dvaldivia (approved):

PR #810, opened 2021-06-11
fix: GetAllCertificatesAndCAs shouldn't fail internally

harshavardhana:
all libraries in pkg/* should never called Fatal()
internally, console is imported now and it is important
that the failure logging etc are all well controlled.

Bonus: update to latest minio/pkg v1.0.6 to get trial
customer license verification fixes.


Alevsk (approved):

harshavardhana:
Yo PTAL @Alevsk @dvaldivia @bexsoft

dvaldivia (approved):

Alevsk (approved):

bexsoft (approved):
LGTM

PR #813, opened 2021-06-16
Bring Tenant Metrics to Tenant Details

dvaldivia:
Screen Shot 2021-06-15 at 8 53 57 PM

Fixes https://github.com/miniohq/engineering/issues/173


portal-ui/src/screens/Console/Tenants/TenantDetails/TenantSummary.tsx
```diff
@@ -441,8 +438,6 @@ const mapState = (state: AppState) => ({
   ),
 });
 
-const connector = connect(mapState, {
-  setErrorSnackMessage,
-});
+const connector = connect(mapState, {});
```
bexsoft, line 441: We can pass null instead of an empty object
```suggestion
const connector = connect(mapState, null);
```

swagger.yml
```diff
@@ -2303,6 +2303,70 @@ paths:
             $ref: "#/definitions/error"
       tags:
         - AdminAPI
+  /namespaces/{namespace}/tenants/{tenant}/info:
+    get:
+      summary: Tenant Info
+      operationId: TenantInfo
```
bexsoft, line 2309: I think this should be called `TenantMetrics`

bexsoft (changes requested):

swagger.yml
dvaldivia: This one returns the equivalent of `mc admin info` for the tenant, that's why I renamed the other to TenantDetails

bexsoft (approved):

swagger.yml
bexsoft: ok

Alevsk (approved):

PR #818, opened 2021-06-17
fix: do not use var to define global sets

harshavardhana:
console when imported cannot honor certain environment
values, we need to make sure to read it directly from
environment.


harshavardhana:
Taking this in to continue my testing on MinIO server and updating the PR .

PR #830, opened 2021-06-23
Add Filtering to Users on PolicyDetails

adfost:
Screen Shot 2021-06-22 at 6 02 40 PM

Add filtering to list users with policy.


portal-ui/src/screens/Console/Tenants/TenantDetails/PodDetails.tsx
```diff
@@ -0,0 +1,351 @@
+// This file is part of MinIO Console Server
```
dvaldivia, line 1: please remove this file and send only the changes if any, this was moved, rebase to master to find out the new location

dvaldivia (changes requested):

portal-ui/src/screens/Console/Policies/PolicyDetails.tsx
```diff
@@ -191,14 +199,22 @@ const PolicyDetails = ({
...
-        });
+      if (loadingUsers) {
+        api
+          .invoke("GET", `/api/v1/policies/${policyName}/users`)
+          .then((result: any) => {
+            var resultUserArray = [];
+            for (var i = 0; i < result.length; i++) {
+              var added: UserName = { name: result[i] };
+              resultUserArray.push(added);
+            }
```
bexsoft, line 210: Please use a`.map` instead

portal-ui/src/screens/Console/Users/types.tsx
```diff
@@ -25,6 +25,9 @@ export interface User {
   policy?: string[];
 }
 
+export interface UserName {
+  name: string;
+}
```
bexsoft, line 30: There is no need to create an object to make filtering, you can apply it in any array of strings (as the one that users list returns)

bexsoft (changes requested):
Please add more information to this PR as it doesn't refer a task or tells us what is doing

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/RewindEnable.tsx
```diff
@@ -0,0 +1,153 @@
+// This file is part of MinIO Console Server
```
dvaldivia, line 1: why are you adding this file?

PR #831, opened 2021-06-23
Make ElementKey optional

dvaldivia:


bexsoft (changes requested):
Why are we trying to apply this change? TableWrapper is designed to receive an empty string when you are sending a list of string to render

bexsoft (dismissed):

bexsoft (approved):

Alevsk (approved):

PR #839, opened 2021-06-26
add build tags to separate operator code

harshavardhana:


harshavardhana:
I haven't changed the UI side of things here for separation, I am assuming that will be done by @bexsoft @dvaldivia ??

restapi/admin_subscription.go
```diff
@@ -15,381 +15,20 @@
...
-		if err != nil {
-			return admin_api.NewSubscriptionRefreshDefault(int(err.Code)).WithPayload(err)
-		}
-		return admin_api.NewSubscriptionRefreshOK().WithPayload(license)
-	})
-}
-
-// retrieveLicense returns license from K8S secrets (If console is deployed in operator mode) or from
-// the configured CONSOLE_SUBNET_LICENSE environment variable
+// retrieveLicense returns license from the configured CONSOLE_SUBNET_LICENSE environment variable
```
dvaldivia, line 29: we probably need this for license APIs to work,let me address that first on another PR

dvaldivia (changes requested):

restapi/admin_subscription.go
harshavardhana: oh okay - can you check with the handlers do get registered correctly? - because all I did was attempt curl and check.

dvaldivia:
I think this is not needed anymore is it?

PR #852, opened 2021-07-07
Integer input for NiceBytes and NiceDays

adfost:
Added the functions NiceBytesInt and NiceDaysInt in order to use integers instead of strings for input. In order to maintain backwards compatibility, I retained NiceBytes and NiceDays with string input, although I changed them to call the new functions to simplify the code. As this was utility change not sure what image to add.

### DIFF
--- a/portal-ui/src/common/utils.ts
+++ b/portal-ui/src/common/utils.ts
@@ -36,8 +36,13 @@ export const k8sUnits = ["Ki", "Mi", "Gi", "Ti", "Pi", "Ei"];
 export const k8sCalcUnits = ["B", ...k8sUnits];
 
 export const niceBytes = (x: string, showK8sUnits: boolean = false) => {
-  let l = 0,
-    n = parseInt(x, 10) || 0;
+  let n = parseInt(x, 10) || 0;
+
+  return niceBytesInt(n, showK8sUnits)
+};
+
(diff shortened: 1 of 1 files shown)


portal-ui/src/common/utils.ts
```diff
@@ -468,13 +478,13 @@ export const niceDays = (secondsValue: string, timeVariant: string = "s") => {
...
+          ? `${minutes} minute${minutes > 1 ? "s" : ""}`
+          : ""
   } ${
-    seconds >= 1 && minutes === 0 && hours === 0
-      ? `${seconds} second${seconds > 1 ? "s" : ""}`
-      : ""
+      seconds >= 1 && minutes === 0 && hours === 0
+          ? `${seconds} second${seconds > 1 ? "s" : ""}`
+          : ""
   }`;
```
harshavardhana, line 488: are there no javascript packages that provide humanize helpers? why are we doing this manually?

harshavardhana (changes requested):

portal-ui/src/common/utils.ts
adfost: This was the same code I got before.

adfost: just modified the inputs

dvaldivia (dismissed):

dvaldivia (approved):

harshavardhana (approved):

PR #856, opened 2021-07-08
Delete multiple objects

adfost:
Delete multiple objects using selectors on object list

Screen Shot 2021-07-08 at 12 44 02 PM


dvaldivia:
When the button for `Delete Selected` becomes enabled, the icon keeps it's original color so it can't be seen, it should flip to white, same for the font, 

Screen Shot 2021-07-09 at 9 51 40 AM


After deleting an object, the button doesn't reset to it's original state

Screen Shot 2021-07-09 at 9 52 44 AM

dvaldivia (approved):
testd. lgtm

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/DeleteMultipleObjects.tsx
```diff
@@ -0,0 +1,120 @@
...
+          color="primary"
+          disabled={deleteLoading}
+        >
+          Cancel
+        </Button>
+        <Button
+          onClick={() => {
+            for (var i = 0; i < selectedObjects.length; i++) {
+              removeRecord(i);
+            }
```
Alevsk, line 102: This will trigger multiple delete object calls, and the `removeRecord` inside has

```
        setDeleteLoading(false);
        closeDeleteModalAndRefresh(true);
```

that will be called multiple times as well, @adfost its better to implement a new endpoint called bulkDelete that receive a list of object IDs and then perform the delete operations on the backend, once all objects are deleted you return operation success to the frontend, that way you will have only 1 call instead of multiple ones

Alevsk (changes requested):

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/DeleteMultipleObjects.tsx
```diff
@@ -0,0 +1,122 @@
...
+        recursive.push(true);
+      } else {
+        recursive.push(false);
+      }
+    }
+    setDeleteLoading(true);
+    api
+      .invoke(
+        "DELETE",
+        `/api/v1/buckets/${selectedBucket}/delete-multiple?path=${selectedObjects}&recursive=${recursive}`
```
bexsoft, line 64: If we are sending multiple paths to delete, I think it's better so send them in the body of the request as a string array instead of sending them in the query. This may cause troubles in the future.

Please change API accordingly

```suggestion
        `/api/v1/buckets/${selectedBucket}/delete-multiple?recursive=${recursive}`, {path: selectedObjects}
```

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
```diff
@@ -837,6 +879,17 @@ const ListObjects = ({
...
+              variant="contained"
+              color="primary"
+              startIcon={<DeleteIcon />}
+              onClick={() => {
+                setDeleteMultipleOpen(true);
+              }}
+              disabled={selectedObjects.length === 0}
+            >
+              Delete Selected
+            </Button>
```
bexsoft, line 892: Please change the position of this button as this is not the main action that we want to promote. Please change it before the create folder button:

125111859-452c0e80-e09b-11eb-8f6d-8a6242dc2a02

bexsoft (changes requested):

bexsoft (dismissed):

restapi/user_objects.go
```diff
@@ -283,6 +290,29 @@ func getDeleteObjectResponse(session *models.Principal, params user_api.DeleteOb
...
 
+// getDeleteMultiplePathsResponse returns whether there was an error on deletion of any object
+func getDeleteMultiplePathsResponse(session *models.Principal, params user_api.DeleteMultipleObjectsParams) *models.Error {
+	ctx := context.Background()
+	var version string
+	for i := 0; i < len(params.PathName); i++ {
+		if params.VersionID != nil {
+			version = *params.VersionID
+		}
+		s3Client, err := newS3BucketClient(session, params.BucketName, params.PathName[i])
```
dvaldivia, line 301: This is inefficient, initialize the client just once.

swagger.yml
```diff
@@ -370,6 +370,41 @@ paths:
       tags:
         - UserAPI
 
+  /buckets/{bucket_name}/delete-multiple/{path_name}:
```
dvaldivia, line 373: `/buckets/{bucket_name}/delete-objects`

dvaldivia (changes requested):

dvaldivia (approved):
Tested. LGTM

Alevsk (approved):

dvaldivia (dismissed):

dvaldivia (approved):

bexsoft (approved):

PR #859, opened 2021-07-09
Fix escape characters in filename.

adfost:
Automatically convert escape characters in filenames, so there is no problem with deleting such files.

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
@@ -515,7 +515,7 @@ const ListObjects = ({
     const formData = new FormData();
 
     for (let file of files) {
-      const fileName = file.name;
+      const fileName = unescape(file.name);
       const blobFile = new Blob([file]);
       formData.append(fileName, blobFile);
     }


dvaldivia (changes requested):
This doesn't fix the issue, I have a file called `02%20-%20FLY%20ME%20TO%20THE%20MOON%20.mp3` which I copied to my bucket, that's the actual name of the file, I cannot delete it via browser

example on loading the file:

```
➜ ls 02%20-%20FLY%20ME%20TO%20THE%20MOON%20.mp3 
02%20-%20FLY%20ME%20TO%20THE%20MOON%20.mp3
➜ mc cp 02%20-%20FLY%20ME%20TO%20THE%20MOON%20.mp3 play/000test/
...E%20MOON%20.mp3:  5.90 MiB / 5.90 MiB  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  2.25 MiB/s 2s console git:(3b8bbce70) ✗ 
➜ mc ls play/000test                                            
[2021-07-13 17:00:09 PDT] 5.9MiB 02%20-%20FLY%20ME%20TO%20THE%20MOON%20.mp3
```

harshavardhana (changes requested):
You can't unescape in this style you generally get the object in URL encoded form in the list response.

This needs to be converted on the client side which - minio-go already does.

The way the conversion should happen is such that when the object name is selected to delete the server handler component in console should convert back the file component and send it to DeleteObject()

Note there are several nuances here that needs to be understood so just the UI side fix is not enough.

bexsoft:
What ticket is this PR fixing?

dvaldivia:
@bexsoft it's the problem when trying to delete files with weird characters

harshavardhana:
The fundamental problem we found here is related to swagger marshalling which is causing incompatibilities, swagger marshaller is wrongly detecting filenames and percentage encoded characters incorrectly which leads to completely incorrect values. 

So the correct fix here is we write a custom marshaller for specific structs that populates object names to avoid URL encoding of object key names.

dvaldivia:
@harshavardhana it's a simpler problem imo, from the client we need to html encode and then decode on the server to avoid the other decoders, this will solve the issue

harshavardhana:
> @harshavardhana it's a simpler problem imo, from the client we need to html encode and then decode on the server to avoid the other decoders, this will solve the issue

No it is not @dvaldivia there is more context missing first of all swagger should never actually encode and decode values. 

Doing that causes problems for the WebUI and you cannot safely assume escape sequences on the UI since you lack the context on the object names presented.

PR #861, opened 2021-07-09
fix: change password is implicit

### LINKED ISSUE #857: Allow users to change their password
Earlier WebUI of MinIO allowed users to change their password. As pointed out by @harshavardhana, Console restricts "Change password" only to users with `admin:CreateUser` permissions which is interpretation of Console of users privileges. As older WebUI is phased out, it is currently impossible for users to change passwords without admin privileges using available tools. At the same time, permission to change user password for her/himself is implicit. Unfortunately, it is not reflected currently in Console.

Please allow users to change their passwords to replicate older functionality of WebUI in Console.

Issue earlier discussed in https://github.com/minio/minio/discussions/12649

### LINKED ISSUE #858: Allow non-admin users to create new buckets
Continuation of https://github.com/minio/minio/issues/12651

With the non-admin user account, Console does not allow to create new buckets while user does have permission to do so. Filing as a bug since earlier WebUI allowed to do so.

User permissions:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::${aws:username}-*"
            ]
        }
    ]
}
```

Note that the permissions allow to create buckets with some specific names. I wonder whether it would make sense to show GUI buttons allowing to create buckets in "Object browser" and under "Buckets" without checking policies regarding it. If it is not allowed, API would report it and then just showing the error message would be sufficient.

MinIO release: RELEASE.2021-07-08T19-43-25Z

"Create bucket" button, as shown in the screenshot below (screenshot has bucket names and sizes removed, otherwise they are shown) is missing for such user. It is also missing under "Buckets" section. No error in Console UI either.

Issue moved from the main repository as it is expected to be here (reference https://github.com/minio/minio/issues/12664)

no_bucket_create

harshavardhana:
Also, only match actions do not need to match
dynamic values in resources.

fixes #857
fixes #858

### DIFF
--- a/pkg/acl/endpoints.go
+++ b/pkg/acl/endpoints.go
@@ -178,12 +178,8 @@ var serviceAccountsActionSet = ConfigurationActionSet{
 
 // changePasswordActionSet requires admin:CreateUser policy permission
 var changePasswordActionSet = ConfigurationActionSet{
-	actionTypes: iampolicy.NewActionSet(
-		iampolicy.AllAdminActions,
-	),
-	actions: iampolicy.NewActionSet(
-		iampolicy.CreateUserAdminAction,
-	),
+	actionTypes: iampolicy.NewActionSet(),
+	actions:     iampolicy.NewActionSet(),
 }
 
 // tenantsActionSet temporally no actions needed for tenants sections to work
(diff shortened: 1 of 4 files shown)


harshavardhana:
I see that no one is available to review, and we need this to be available for tomorrow's MinIO release.

harshavardhana:
> I see that no one is available to review, and we need this to be available for tomorrow's MinIO release.

In lieu of this will proceed to merge this.

PR #864, opened 2021-07-13
fix: remove response timeouts for console

harshavardhana:


cmd/console/server.go
```diff
@@ -194,6 +194,10 @@ func StartServer(ctx *cli.Context) error {
 
 	server.Host = rctx.Host
 	server.Port = rctx.HTTPPort
+	// set conservative timesout for uploads
```
dvaldivia, line 197: this is an auto-generated file we should see this values on `configure_console.go`

dvaldivia (changes requested):

cmd/console/server.go
harshavardhana: `cmd/console/server.go` is not an autogenerated file @dvaldivia

dvaldivia (approved):

Issue #723, opened 2021-05-07
Incorrect object behaviors with standalone/FS mode


Weilbyte:
I am running MinIO console with a MinIO standalone instance that is using a data folder. Console is `minio/console:latest` and MinIO is `2021.4.22`. I have so far ran into a few issues regarding buckets and interacting with objects within them. 

## The issues
The following assumes a bucket created with no versioning (not supported), no object locking (not supported) and no quota.

### Error when getting object details
Clicking on an object/file will make an error message pop up ("An error occurred, please try again"). This is happening because the object details request is being sent with parameter `with_versions` is set to `true`, despite versioning being turned off (and not supported at all) in our bucket - as a result, error with status 500 is returned.    

Container error message:

```
2021-05-07 15:01:13.495680 I | 2021/05/07 15:01:13 error.go:47: original error: A header you provided implies functionality that is not implemented
```

Request:

```
> https://console.[SNIP]/api/v1/buckets/testb/objects?prefix=example.mp4&with_versions=true

{"code":500,"message":"an error occurred, please try again"}
```

### Error sharing object
From the object screen, using the buttons presented below the object name and path, downloading works, deleting works but sharing *does not*, upon the modal opening you will be greeted with an error ("Prefix in query is required"). Pictured below.

![image](https://user-images.githubusercontent.com/43392677/117469696-50a57100-af56-11eb-970f-113b7953ae2a.png)

This is occurring because the request sent does not include the object path/prefix. 

Request:

```
> https://console.[SNIP]/api/v1/buckets/testb/objects/share?prefix=&version_id=null

{"code":602,"message":"prefix in query is required"}
```

This error can also be seen in #654 

### Unsupported options present during bucket creation
When creating a bucket in standalone/FS mode, the unsupported options should be grayed out (using them gives a generic error and leaves the user with no insight into why it doesn't work). I am unsure of how the console can detect it is being used with a MinIO server in that configuration but maybe this information could be given to the console via an environment variable. 

At the very least, a more descriptive error should be given to the user (instead of just "An error occurred, please try again", they would receive "{$OPTION} not supported in FS mode").

harshavardhana:
console is currently implemented for only distributed setups @Weilbyte

Weilbyte:
@harshavardhana would PRs that address non-distributed setups be accepted?

harshavardhana:
> @harshavardhana would PRs that address non-distributed setups be accepted?

[...]

PR #874, opened 2021-07-16
fix: properly support FS/Gateway mode

harshavardhana:
fixes https://github.com/minio/console/issues/723

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/minio/cli v1.22.0
 	github.com/minio/direct-csi v1.3.5-0.20210601185811-f7776f7961bf
 	github.com/minio/kes v0.11.0
-	github.com/minio/madmin-go v1.0.12
+	github.com/minio/madmin-go v1.0.17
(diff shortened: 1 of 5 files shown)


Alevsk (approved):

harshavardhana:
More changes this PR needs is UI Elements need to be turned off if certain backend is chosen.

I am not aware of how to do that @dvaldivia @bexsoft

bexsoft:
> More changes this PR needs is UI Elements need to be turned off if certain backend is chosen.
> 
> I am not aware of how to do that @dvaldivia @bexsoft

I will start working on that today @harshavardhana, in this case we will just need to return a flag during login and send it in the response so we can enable or disable items like we do with IAM

bexsoft (dismissed):

harshavardhana:
PTAL @Alevsk @dvaldivia

dvaldivia (approved):

dvaldivia:
Tested. LGTM

bexsoft (approved):

Issue #752, opened 2021-05-17
Console fails to handle some characters in access key


lkiesow:
Opening console and creating a user with an access key like `<b>xy</b>` will cause the user interface to misbehave, and you cannot access the user details:

https://user-images.githubusercontent.com/1008395/118556469-35134500-b764-11eb-9487-8d79d587f619.mp4

dvaldivia:
Thanks for the report! Looks like we need to URL the access key for the routing not to break!

lkiesow:
Something similar happens for the groups:

![Screenshot from 2021-05-17 23-06-09](https://user-images.githubusercontent.com/1008395/118556882-b5d24100-b764-11eb-8c38-f3a502247edb.png)

dvaldivia:
seems like the routing for both our restful APIs and on react router breaks with this, I'm testing encoding the values and it's not working either

PR #896, opened 2021-07-23
Fixed encoding & URL issues for users & groups pages

bexsoft:
fixes https://github.com/minio/console/issues/752

## What does this do?

Fixes an issue with encoding & users

## How does it look?

Screen Shot 2021-07-23 at 18 35 13
Screen Shot 2021-07-23 at 18 34 10


portal-ui/src/screens/Console/Users/ListUsers.tsx
```diff
@@ -155,7 +155,8 @@ const ListUsers = ({ classes, setErrorSnackMessage }: IUsersProps) => {
   };
 
   const viewAction = (selectionElement: any): void => {
-    history.push(`/users/${selectionElement.accessKey}`);
+    console.log(encodeURI(selectionElement.accessKey));
```
dvaldivia, line 158: remove

swagger-console.yml
```diff
@@ -1269,13 +1269,13 @@ paths:
       tags:
         - AdminAPI
 
-  /users/{name}/groups:
+  /user/groups:
```
dvaldivia, line 1272: I think we should keep `{name}` in the URL, as this is more RESTful, for the resource `{user}` put this or get that

dvaldivia (changes requested):

swagger-console.yml
bexsoft: This is causing issues with names that contain slashes, that is why I changed it to be in query instead. Also I changed to be singular to handle this

dvaldivia: ok, that makes sense then

dvaldivia (approved):

portal-ui/src/screens/Console/Users/ListUsers.tsx
bexsoft, line 158:
```suggestion
    
```

Alevsk (approved):

harshavardhana (approved):

PR #916, opened 2021-08-02
Pod events bug fix

adfost:
Pod events will not disappear anymore if you return to the pod events tab. Fixing If there are events for a pod, when switching tabs from Events to Logs in PodDetails, the events are lost #205.


portal-ui/src/screens/Console/Tenants/TenantDetails/pods/PodDetails.tsx
```diff
@@ -117,15 +117,15 @@ const PodDetails = ({ classes, match }: IPodDetailsProps) => {
             tenant={tenantName}
             namespace={tenantNamespace}
             podName={podName}
-            propLoading={loading}
+            propLoading={true}
```
bexsoft, line 120: This removes the ability of the component to load the PorEvents information on demand as you are forcing to re-render on every state change. Please review what these props do inside PodEvents & PodLogs.

bexsoft (changes requested):

PR #919, opened 2021-08-02
Pod events disappearing bug fix

adfost:
Pod events will not disappear anymore if you return to the pod events tab. Fixing If there are events for a pod, when switching tabs from Events to Logs in PodDetails, the events are lost #205.


bexsoft (approved):
LGTM

dvaldivia (approved):

PR #920, opened 2021-08-02
Fix Hop File Upload and Query Params

dvaldivia:
Fixes a bug preventing file uploads under hop and also passing query params which was making it so object details were not showing versions


harshavardhana:
Does this change allow for running console under a separate prefix? @dvaldivia

bexsoft (changes requested):
Issue during file upload under port :9090 & :5000, with. big files an error in blue is thrown. With small files it says that file was uploaded successfully, but file doesn't appear in the list. No errors are seen in server or MinIO

Screen Shot 2021-08-02 at 19 02 24
Screen Shot 2021-08-02 at 19 01 44

bexsoft (dismissed):
LGTM

harshavardhana:
> Does this change allow for running console under a separate prefix? @dvaldivia

@dvaldivia ^^?

dvaldivia:
Yes @harshavardhana

harshavardhana:
> Yes @harshavardhana

Can you document this behavior and also fix the title of this PR - we can't tell if that is what you are supporting here.

It will help @ravindk89 document this properly for k8s environments basically now we can have 

MINIO_SERVER_URL=https://ingress.example.com
MINIO_BROWSER_REDIRECT_URL=https://ingress.example.com/console

dvaldivia:
@harshavardhana it's fixing a problem where the hop screen from operator UI was not passing query parameters, and also the file upload was not working in this mode

ravindk89:
For my understanding:

Starting in the next Console release, you can do:

```
MINIO_SERVER_URL=https://ingress.example.com
MINIO_BROWSER_REDIRECT_URL=https://ingress.example.com/console
```

For prior Console releases, what is the behavior? Is the expectation there for `MINIO_BROWSER_REDIRECT_URL` to match `MINIO_SERVER_URL` ? It would be ideal for us to track both, as there will be users who cautiously upgrade and will need to know what the correct behavior is for their specific version.

dvaldivia:
@ravindk89 not starting on the next release but soon, we still need to test console running on a subdirectory

ravindk89:
OK - so what is the current expected behavior with these envvars? cc @harshavardhana  - trying to keep this straight, its a bit confusing.

dvaldivia:
@ravindk89 so this PR is only about fixing a couple of bugs regarding the hop screen, on a separate PR we can complete the setup of running it on a subfolder, so let's not announce it yet

Alevsk (dismissed):

Alevsk (approved):

bexsoft (approved):

PR #928, opened 2021-08-05
Select policies while creating new user

dvaldivia:
Also removed some duplicate structures in `swagger-operator.yaml`

Screen Shot 2021-08-05 at 12 37 52 PM


harshavardhana (approved):

adfost (dismissed):

bexsoft (changes requested):
There is an issue where if I set groups to a user, the policy list is returned as empty in `api/v1/user?name=<user_name>` request. Also during the create user query the policies are not returned in the response

dvaldivia:
@bexsoft fixed

bexsoft (approved):
LGTM, Works fine

harshavardhana (approved):

PR #956, opened 2021-08-17
Added pre validation of prometheus URL

### LINKED ISSUE #947: Console issues when Prometheus endpoint is configured but Prometheus cannot be reached
minio --version
minio version RELEASE.2021-08-05T22-01-19Z

Steps to repro:
1. Have prometheus endpoint configured for MinIO console (e.g. `MINIO_PROMETHEUS_URL=http://localhost:9090`)
2. Start minio when prometheus is not available (not started, network disconnect etc)
3. Log in to console
4. Dashboard shows all spinning icons (as expected)
5. Navigate to license page (or pages listed below)
  - Notice that the page never loads.  If you try to refresh the page, it hangs.  Pressing stop button in the browser causes full log out of console

If you start/reconnect to prometheus and refresh, things work as expected.

Observed it affecting the following pages:

Object browser
Account
Buckets
Users
Groups
IAM policies
License

bexsoft:
fixes #947 

## What does this do?

Adds a pre-validation to prometheus URL, so we can avoid long timeouts & unnecessary requests that will fail if URL is unreachable

## How does it look?
Screen Shot 2021-08-16 at 23 45 18

### DIFF
--- a/restapi/admin_info.go
+++ b/restapi/admin_info.go
@@ -19,7 +19,9 @@ package restapi
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"regexp"
 	"strings"
@@ -871,9 +873,37 @@ func unmarshalPrometheus(endpoint string, data interface{}) bool {
 	return false
 }
 
+func testPrometheusURL(url string) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url+"/-/healthy", nil)
+
+	if err != nil {
+		LogError("Error Building Request: (%v)", err)
+		return false
+	}
+
+	response, err := GetConsoleHTTPClient().Do(req)
+
+	if err != nil {
+		LogError("Non reachable Prometheus URL: %s (%v)", url, err)
+		return false
+
+	}
+
+	return response.StatusCode == http.StatusOK
+}
+
 func getAdminInfoWidgetResponse(params admin_api.DashboardWidgetDetailsParams) (*models.WidgetDetails, *models.Error) {
 	prometheusURL := getPrometheusURL()
 	prometheusJobID := getPrometheusJobID()
+
+	// We test if prometheus URL is reachable. this is meant to avoid unuseful calls and application hang.
+	if !testPrometheusURL(prometheusURL) {
+		error := errors.New("Prometheus URL is unreachable")
+		return nil, prepareError(error)
+	}
 
 	return getWidgetDetails(prometheusURL, prometheusJobID, params.WidgetID, params.Step, params.Start, params.End)
 }


restapi/admin_info.go
```diff
@@ -871,10 +873,39 @@ func unmarshalPrometheus(endpoint string, data interface{}) bool {
...
+	_, err := net.DialTimeout("tcp", urlQuery, timeout)
+
+	if err != nil {
+		LogError("Non reachable Prometheus URL: %s (%v)", urlQuery, err)
+		return false
+	}
+
+	return true
+}
+
```
harshavardhana, line 898: Do not attempt a non-TLS dial @bexsoft we should do a full HTTP call such as make a call to 

```
ctx, cancel := context.WithTimeout(context.Background(), 2 * time.Second)
defer cancel()
...
req, err := http.NewRequestWithTimeout(ctx, http.MethodGet, URL+"/-/healthy", nil)
...

then use 
GetConsoleHTTPClient().Do(req)
```

dvaldivia (approved):

restapi/admin_info.go
bexsoft: Sounds good, changes applied

dvaldivia (approved):

restapi/admin_info.go
```diff
@@ -871,10 +873,42 @@ func unmarshalPrometheus(endpoint string, data interface{}) bool {
...
+		LogError("Non reachable Prometheus URL: %s (%v)", url, err)
+		return false
+
+	}
+
+	if response.StatusCode == http.StatusOK {
+		return true
+	}
+
+	return false
```
harshavardhana, line 899:
```suggestion
        return response.StatusCode == http.StatusOK
```

restapi/admin_info.go
```diff
@@ -871,10 +873,42 @@ func unmarshalPrometheus(endpoint string, data interface{}) bool {
...
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url+"/-/healthy", nil)
+
+	if err != nil {
+		LogError("Error Building Request: (%v)", err)
+		return false
+	}
+
+	response, err2 := GetConsoleHTTPClient().Do(req)
```
harshavardhana, line 887:
```suggestion
	response, err := GetConsoleHTTPClient().Do(req)
```
Simplify

restapi/admin_info.go
```diff
@@ -871,10 +873,42 @@ func unmarshalPrometheus(endpoint string, data interface{}) bool {
...
+
+	if err2 != nil {
```
harshavardhana, line 889:
```suggestion
	if err != nil {
```

harshavardhana (changes requested):

restapi/admin_info.go
bexsoft: done

restapi/admin_info.go
bexsoft: done

restapi/admin_info.go
bexsoft: done

dvaldivia (approved):

harshavardhana (approved):

wasilak:
Hey guys, new validation broke integration with [VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics), which is Prometheus compatible database, but unfortunately doesn't have `/-/healthy` endpoint.

Issue #952, opened 2021-08-16
Subpath deployment not working


TomHellier:
I am trying to deploy minio on a subpath with the latest version 0.9.0

The browser blocks the loading of the javascript and stylesheet resources as the MIME type provided is text/html, instead of text/javascript or text/css

```
GET https://example.com/minio/ [HTTP/1.1 200 OK 213ms]
GET https://fonts.googleapis.com/css2?family=Lato:wght@400;500;700;900&display=swap [HTTP/3 200 OK 223ms]
GET https://example.com/minio/styles/root-styles.css [HTTP/1.1 200 OK 30ms]
GET https://example.com/minio/static/css/2.c018cba7.chunk.css [HTTP/1.1 200 OK 42ms]
GET https://example.com/minio/static/css/main.8cfac526.chunk.css [HTTP/1.1 200 OK 104ms] 
The resource from “https://example.com/minio/styles/root-styles.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
GET https://example.com/minio/static/js/2.da9d8f74.chunk.js [HTTP/1.1 200 OK 46ms]
GET https://example.com/minio/static/js/main.8f566b91.chunk.js [HTTP/1.1 200 OK 112ms]
The resource from “https://example.com/minio/static/css/2.c018cba7.chunk.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
The resource from “https://example.com/minio/static/css/main.8cfac526.chunk.css” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
The resource from “https://example.com/minio/static/js/2.da9d8f74.chunk.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
The resource from “https://example.com/minio/static/js/main.8f566b91.chunk.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
GET https://example.com/minio/static/js/2.da9d8f74.chunk.js [HTTP/1.1 200 OK 31ms]
The resource from “https://example.com/minio/static/js/2.da9d8f74.chunk.js” was blocked due to MIME type (“text/html”) mismatch (X-Content-Type-Options: nosniff).
Loading failed for the <script> with source “https://example.com/minio/static/js/2.da9d8f74.chunk.js”. minio:1:1
[...]

Issue #853, opened 2021-07-08
support  relative URL


perfectstorm88:
why minio console not use  relative URL?So that minio console website will not be bound to the base URL of where it is currently deployed.

such as: ONLY `http://127.0.0.1:9000/` worked,but when I use nginx to proxy, `http://127.0.0.1:8080/myminio/`  NOT WORKED!

the `index.html` use  absolute URL, why  not use  relative URL?
```html
<html>
    <head>
        <meta charset="UTF-8">
        <meta name="viewport" content="width=device-width, initial-scale=1">
        <title>MinIO Browser</title>

        <link rel="icon" type="image/png" sizes="32x32" href="/minio/favicon-32x32.png">
        <link rel="icon" type="image/png" sizes="96x96" href="/minio/favicon-96x96.png">
        <link rel="icon" type="image/png" sizes="16x16" href="/minio/favicon-16x16.png">

        <link rel="stylesheet" href="/minio/loader.css" type="text/css">
    </head>

```

harshavardhana:
There are no plans to support relative URLs @perfectstorm88 sorry.

perfectstorm88:
This is a small change, but it's easy to use。(Only change the href pah from  absolute URL  to relative URL)

It's strange why we don't support it.

PeterFalken:
Does this also apply to built in console on the minio server?
I'm experiencing a similar issue - when running the minio server behind a proxy (Traefik)

Using sample command from MinIO download/quickstart page:
`MINIO_ROOT_USER=admin MINIO_ROOT_PASSWORD=password ./minio server /mnt/data --console-address ":9001"`

- Traefik is configured to forward request sent to domainxxx.com/minio/ to localhost:9001
- Initial page is blank, WebConsole shows errors caused by resources using absolute paths under '/static' & '/styles'
- Previous console worked fine behind reverse proxies (both nginx & traefik)
- If relative URLs are out of the question, can there be a parameter to set the base URL?

TomHellier:
https://github.com/minio/console/issues/396 seems to say that this issue is fixed ?

TomHellier:
@dvaldivia, happy to try it out, testing with v0.8.1 still shows the same problem though, and it looks like master still has the slash at the start of the urls. https://github.com/minio/console/blob/master/portal-ui/public/index.html . Is this on a docker container I could test with?

dvaldivia:
[...]

Issue #396, opened 2020-11-16
Console UI behind nginx subpath


harshitmahapatra:
Hi,
Is there a way to set up the console UI at a subpath behind a reverse proxy like Nginx?
It would be nice if we could provide an environment variable like SERVER_BASEPATH to start the server at a subpath.

kevinlul:
Can't you just rewrite the request path in nginx?

dvaldivia:
We can look into this as soon as we have some buffer time for development, we also welcome PRs.

We'll need react to be ready to work with the sub folder

mqu:
hello, same behavior wanted here. It would be nice to have at same URL minio and minio-console in subpath. I have tried to put console to https://minio.example.org/admin, but it does not work.

thanks and best regards, Marc.

hypervtechnics:
We also need this

dvaldivia:
This is supported with the latest on master, will be there on the next release

TomHellier:
@dvaldivia what is the config option needed? or a link to the PR which adds this functionality? thanks :)

thesuperzapper:
@dvaldivia as of `v0.10.4` it is still not possible to use Minio Console under an HTTP prefix, see: https://github.com/minio/console/issues/853#issuecomment-947179997.

berendiwema:
@TomHellier were you able to find the config option?

TomHellier:
@berendiwema I've fixed this in #952 #964 I've been running the docker version created from that PR in ops for several months now without issue.

berendiwema:
@TomHellier I see. Does that fix work without nginx as well? #964 looks like it's depending on nginx rewrites right?

TomHellier:
dunno, I only tested it using nginx-ingress for kubernetes, I don't know enough about your use case to help more. Maybe just give it a go?

alternatively @reddec wrote something here:
https://github.com/reddec/miniconsole

PR #964, opened 2021-08-19
fixes 952, allows minio console to be served from a subpath

TomHellier:
continuing on from the issues #396 , #853 and #952 

I've managed to get this serving on a /minio subpath using nginx ingress on kubernetes. I'd appreciate some initial feedback @dvaldivia if possible?

Deploying this in kubernetes with the following ingress seems to work:
```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/rewrite-target: /$2
  name: minio-console
  namespace: monitoring
spec:
  rules:
    - host: example.com
      http:
        paths:
          - backend:
              serviceName: minio-console
              servicePort: 9090
            path: /minio(/|$)(.*)
```
It works with and without a subpath, but I don't know whether any regressions have been introduced as I don't really know the features of minio/console


portal-ui/src/common/api/index.ts
```diff
@@ -27,8 +27,8 @@ export class API {
       .send(data)
       .then((res) => res.body)
       .catch((err) => {
-        // if we get unauthorized, kick out the user
-        if (err.status === 401) {
+        // if we get unauthorized, kick out the user, unless they are attempting to log in
+        if (err.status === 401 && url !== "/api/v1/login") {
```
dvaldivia, line 31: I think we can get a 401 if the STS in the session is expired as it usually only last 30 minutes, so we need this in case the user walked away from the session and then came back (after 30 minutes), any action should log him out I think, cc @Alevsk

TomHellier: Hey @dvaldivia , thanks for looking. For more context, I made this change to allow the error information to propagate back up to the login page, as it looked like that was the only place that didnt use the api.ts file. There is no point redirecting to the login page if the login failed because the error would be silent to the user.

TomHellier:
@dvaldivia is there anything I can do to help get this merged?

dvaldivia:
Fixed by https://github.com/minio/console/pull/1247 use `CONSOLE_SUBPATH`

Issue #965, opened 2021-08-20
upload api  default content-type, need other content-type.


overstarry:
Files upload through console always get the content type: application/octet-stream
However, files upload through minio client (mc) or copy direct to data folder will get correct content type.
console upload: 
![image](https://user-images.githubusercontent.com/23260162/130221018-4308583b-dee4-4d7a-b6b0-0d1f95f0f6c3.png)
mc: 
![image](https://user-images.githubusercontent.com/23260162/130221089-118778f9-e796-44fe-b779-6f8f3b69eb2f.png)

example: upload png image file,I want the content-type of image/png, but the system gives me application/octet-stream.

https://github.com/minio/console/blob/2332a1347771ee62de58421723db07c75970dc39/restapi/user_objects.go#L510

dvaldivia:
Thanks for the report, we are aware of this and will fix as soon as we have some bandwidth, we also welcome PRs! I think we just need to detect the content type from the browser

harshavardhana:
Fixed in #968

overstarry:
thanks

PR #995, opened 2021-08-31
Added support to object browser for file preview

### LINKED ISSUE #902: Missing preview feature on new release
## New release
image

## Old release
image

image

bexsoft:
fixes #902 

## What does this do?

Adds file preview support to object browser list. Support inside file details will be added in an upcoming PR

## How does it look?

Screen Shot 2021-08-31 at 15 29 05
Screen Shot 2021-08-31 at 15 29 56
Screen Shot 2021-08-31 at 15 29 42
Screen Shot 2021-08-31 at 15 29 15


dvaldivia (dismissed):
Tested. LGTM.

harshavardhana (approved):

restapi/user_objects.go
```diff
@@ -80,16 +82,39 @@ func registerObjectsHandlers(api *operations.ConsoleAPI) {
...
+			if isPreview {
+
+				// This is done to avoid an issue with content-type detected as "application/octet-stream".
+				buf := new(bytes.Buffer)
+				correctTypeRead, err := getDownloadObjectResponse(session, params)
+				if err != nil {
+					user_api.NewDownloadObjectDefault(int(err.Code)).WithPayload(err)
+				}
+				buf.ReadFrom(correctTypeRead)
+				detectedMIME := http.DetectContentType(buf.Bytes())
```
harshavardhana, line 94: You are downloading the object twice here? and the entire content is read into memory? how will this work?

harshavardhana (changes requested):

restapi/user_objects.go
harshavardhana: You should never have to do this, you should expect that users have set the correct content-type when they are uploading the content. If they have no there is no reason to add preview's for them by adding this kind of code.

This can lead to memory crash if they click on a big video file.

harshavardhana: You should blindly preview without doing this kind of fallback logic, it's not worth the burden for us. It is up to the user to set the correct content-type during upload - which we already do.

harshavardhana (changes requested):

restapi/user_objects.go
bexsoft: This is done to get the correct MIME type from buffer, there was an issue with some files uploaded in console with MIME as "application/octet-stream" sometime ago (https://github.com/minio/console/issues/965). If We try to preview these objects, a download starts instead of allowing us to show them inline.

The idea here is to create a new buffer that allow us to read the correct MIME and send them in the inline header. this piece of code is duplicated as the io reader from mc returns a `io.ReadClose`, so once we read the object it is freed from memory.

If we don't do this, once I try to get the bytes response an error is thrown as stream is already closed after this validation.

harshavardhana: No you cannot do that - you cannot buffer this kind of memory on console - you should simply expect users to have correct content-type. 

If they do not don't preview - it is easier and safer for them to re-upload with content-type instead of we fiddling with the stream.

harshavardhana: > The idea here is to create a new buffer that allow us to read the correct MIME and send them in the inline header. this piece of code is duplicated as the io reader from mc returns a `io.ReadClose`, so once we read the object it is freed from memory.

There is no "free" memory in Go, bytes.Buffer grows memory by 2x for N sized buffer - so if you are going to preview 1GiB video file it will buffer 2GiB of pre-allocation. 

This can easily kill the GC and console server - we cannot afford to have such kind of buffer usage in console.

bexsoft: Logic changed. need to review a couple of things more

restapi/user_objects.go
```diff
@@ -72,24 +73,54 @@ func registerObjectsHandlers(api *operations.ConsoleAPI) {
...
+
+						if len(extension) > 1 {
+							mimeType := mime.TypeByExtension(fmt.Sprintf(".%s", extension[len(extension)-1]))
+
+							if mimeType != "" && mimeType != "application/octet-stream" {
+								contentType = mimeType
+							}
+						}
+					}
+				}
```
harshavardhana, line 119: For this there is already something called as `filepath.Ext()` 

All you need is this

```
contentType := stat.ContentType
if contentType == "application/octet-stream" {
        contentType = mimedb.TypeByExtension(filepath.Ext(filename))
}
```

harshavardhana (changes requested):

restapi/user_objects.go
bexsoft: nice!, will implement it, thank you!

bexsoft: done

dvaldivia (approved):

harshavardhana (approved):

PR #1006, opened 2021-09-02
Disable create service account button if no policy

adfost:
Screen Shot 2021-09-02 at 12 49 06 PM


portal-ui/src/screens/Console/Users/UserServiceAccountsPanel.tsx
```diff
@@ -71,6 +73,17 @@ const UserServiceAccountsPanel = ({
 
   useEffect(() => {
     if (loading) {
+      api
```
dvaldivia, line 76: why are you making this call twice? here and on line 87? you can merge them

dvaldivia (changes requested):

portal-ui/src/screens/Console/Users/UserServiceAccountsPanel.tsx
adfost: Just fixed this

dvaldivia (approved):
Tested. LGTM.

bexsoft (approved):
LGTM, It would be a nice to have, if we show a message explaining why the button is disabled in service accounts page

PR #1009, opened 2021-09-05
Fix NPE on ServerInfoCard

dvaldivia:
This problem was only manifested when running MinIO on Gateway Mode since there's no network object in the server response


bexsoft (approved):
LGTM

Alevsk (approved):

PR #1014, opened 2021-09-07
Applied workaround for MIME issue on windows

bexsoft:
## What does this do?

Applied workaround for Windows MIME issue with console (https://github.com/golang/go/issues/32350)

Thank you @krishnasrinivas for this solution!


kannappanr (approved):
LGTM

PR #1026, opened 2021-09-10
fix: leaking fd's large object uploads

harshavardhana:
Fixes multiple things

- Closes the local open files properly staged
  during multipart form upload.

- Removes these files upon upload properly, otherwise
  we run out of disk space on /tmp/ due to staging
  of these objects.

### DIFF
--- a/restapi/user_objects.go
+++ b/restapi/user_objects.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"mime/multipart"
 	"net/http"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


harshavardhana:
The next work is to make this API fully streaming instead of using staging, we are currently bottlenecked by using ParseMultipartForm() approach and implement a full streaming approach.

dvaldivia (approved):

dvaldivia:
@harshavardhana yeah, we should disallow multiple file uploads together, and do one at a time via UI, so we can stream the file into MinIO

harshavardhana:
> @harshavardhana yeah, we should disallow multiple file uploads together, and do one at a time via UI, so we can stream the file into MinIO

Even multiple files stream is possible just has to be implemented properly.

Alevsk (approved):

PR #1043, opened 2021-09-14
fix: TLS issues with console

harshavardhana:
This PR fixes two bugs one is

- incorrect termination of the HTTP connections
  when the resource URL path is `/`, since `/`
  doesn't exist we should never call h.ServeHTTP()
  instead should be directly served from public assets.

- add SSLHostFunc() such that if the Hostname is empty
  redirection is not empty and this value is handled
  properly when redirecting from 9090 to 9443.


dvaldivia (dismissed):
Tested. LGTM.

dvaldivia (approved):

Alevsk (approved):

Alevsk:
Screen Shot 2021-09-13 at 20 36 14

i think make assets was missed

Issue #1044, opened 2021-09-14
can't access user with '+' in the name


soudhaf:
### Description
when you try to access a user with '+' in the name, you will get redirected to /login page

### How to reproduce
example create user with name test+test try to access it from console

dvaldivia:
Will be fixed by https://github.com/minio/console/pull/1045

dvaldivia:
thanks for the report @soudhaf !

PR #1045, opened 2021-09-14
Fix Users with + in the name

dvaldivia:
Fixes https://github.com/minio/console/issues/1044

### DIFF
--- a/restapi/admin_users.go
+++ b/restapi/admin_users.go
@@ -22,6 +22,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/go-openapi/swag"
+
 	"github.com/go-openapi/errors"
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/minio/console/models"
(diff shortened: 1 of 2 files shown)


restapi/admin_users.go
```diff
@@ -278,6 +280,12 @@ func getUserInfoResponse(session *models.Principal, params admin_api.GetUserInfo
 
 	user, err := getUserInfo(ctx, adminClient, params.Name)
 	if err != nil {
+		// User doesn't exist, return 404
+		if madmin.ToErrorResponse(err).Code == "XMinioAdminNoSuchUser" {
+			var errorCode int32 = 404
+			errorMessage := "User doesn't exist"
+			return nil, &models.Error{Code: errorCode, Message: swag.String(errorMessage), DetailedMessage: swag.String(err.Error())}
+		}
```
Alevsk, line 288: this can be moved to the `errors.go` file in where we handle all the errors

Alevsk (changes requested):

restapi/admin_users.go
dvaldivia: @Alevsk we already handle that error there, and we return a `401` which logsout the user, I'm explicitely handling it here, to avoid a `401` and return a `404`

Alevsk (approved):

bexsoft (approved):

restapi/admin_users.go
soudhaf: @dvaldivia  401 doesn't actually logout the user, at least for this example, you might want to check it
it just redirects you to /login, I can still access the console if I type the address
you get the logout illusion because /login doesn't work as intended check #1050

dvaldivia: the API code on the client side, redirects you to login if it gets a `401` from the server, thus I'm preventing it here for `not found` users

soudhaf: I was saying that 401 doesn't technically losgout users (session is still valid) just a redirect, since this is by design it's all good

PR #1055, opened 2021-09-16
Fix oauth flow and remove go-oidc library

rikatz:
Based on some previous art from Minio Repo (https://github.com/minio/minio/commit/3137dc2eb36b98d793a88caffb06f54371f94899#diff-d155da4b36e796d1a3cdcff2074106787277070e62cc85ead4183b549e45ec5a) we can import the Oauth discovery flow to the code, instead of relying in go-oidc library.

So far, Minio Console relies on go-oidc to establish the `AuthURL` and `TokenURL` which can be gathered directly from `.well-known/openid-configuration` file.

The current implementation of oidc (NewProvider) uses the the "issuer" to build the well-known file, which may not be right. Usually this file can be inside some other path (although this is not common). See https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest -> "Using path components enables supporting multiple issuers per host. This is required in some multi-tenant hosting configurations. This use of .well-known is for supporting multiple issuers per host; unlike its use in RFC 5785 [RFC5785], it does not provide general information about the host"

Also it uses the "host URI" part of the well-known discovery file to assume that the returned "issuer" must be the same, which may not be the case. See: https://github.com/coreos/go-oidc/blob/15b94d97d90c9455981dbdd90a021d34254cfdaf/oidc/oidc.go#L145

Following the same above of "multiple issuers per host", a returned issuer may be different from the host that contains the discovery document.

As for the OAuth only `AuthEndpoint` and `TokenEndpoint` are required, only those needs to be part of the struct (although this can be expanded later, or if required)


pkg/auth/idp/oauth2/provider.go
```diff
@@ -49,6 +49,11 @@ type Config struct {
 	xoauth2.Config
 }
 
+type DiscoveryDoc struct {
```
harshavardhana, line 52: Please add the entire data structure @rikatz for DiscoveryDoc

harshavardhana (changes requested):

pkg/auth/idp/oauth2/provider.go
harshavardhana:
```
// DiscoveryDoc - parses the output from openid-configuration
// for example https://accounts.google.com/.well-known/openid-configuration
type DiscoveryDoc struct {
        Issuer                           string   `json:"issuer,omitempty"`
        AuthEndpoint                     string   `json:"authorization_endpoint,omitempty"`
        TokenEndpoint                    string   `json:"token_endpoint,omitempty"`
        UserInfoEndpoint                 string   `json:"userinfo_endpoint,omitempty"`
        RevocationEndpoint               string   `json:"revocation_endpoint,omitempty"`
        JwksURI                          string   `json:"jwks_uri,omitempty"`
        ResponseTypesSupported           []string `json:"response_types_supported,omitempty"`
        SubjectTypesSupported            []string `json:"subject_types_supported,omitempty"`
        IDTokenSigningAlgValuesSupported []string `json:"id_token_signing_alg_values_supported,omitempty"`
        ScopesSupported                  []string `json:"scopes_supported,omitempty"`
        TokenEndpointAuthMethods         []string `json:"token_endpoint_auth_methods_supported,omitempty"`
        ClaimsSupported                  []string `json:"claims_supported,omitempty"`
        CodeChallengeMethodsSupported    []string `json:"code_challenge_methods_supported,omitempty"`
}
```

Please add a proper comment as well for data structure

harshavardhana (changes requested):

rikatz:
@harshavardhana fixed :D

pkg/auth/idp/oauth2/provider.go
```diff
@@ -241,6 +261,32 @@ func validateOauth2State(state string) error {
...
+// into a DiscoveryDoc struct that have the correct endpoints
+func parseDiscoveryDoc(ustr string) (DiscoveryDoc, error) {
+	d := DiscoveryDoc{}
+	req, err := http.NewRequest(http.MethodGet, ustr, nil)
+	if err != nil {
+		return d, err
+	}
+	clnt := http.Client{
+		Transport: http.DefaultTransport,
+	}
```
harshavardhana, line 274:
```suggestion
	clnt := http.Client{
		Transport: GetConsoleHTTPClient().Transport,
	}
```

harshavardhana (changes requested):

pkg/auth/idp/oauth2/provider.go
harshavardhana: Or you can use `provHttpClient`

rikatz:
hum I have accepted your change (easier) but if you want I can use the provHttpClient as well

harshavardhana (approved):

Alevsk (changes requested):
testing

Alevsk:
@rikatz I tried to send a pr to your branch to include some fixes and validation of the oidc URL but the code on your branch its too old, anyways I created a new PR here https://github.com/minio/console/pull/1063 please verify your code and commits are included here, I tested this code against auth0 and its working fine, thanks.

rikatz:
Closed as #1063 got merged :)

Thanks!

PR #1063, opened 2021-09-20
Fix oauth flow and remove go-oidc library

Alevsk:
- Fix oauth flow
- Validate provided oidc URL

Screen Shot 2021-09-20 at 15 23 42


pkg/auth/idp/oauth2/config.go
```diff
@@ -30,9 +32,20 @@ func GetSTSEndpoint() string {
...
+	openIDConfigPath := "/.well-known/openid-configuration"
+	idpURL := env.Get(ConsoleIDPURL, "")
+	u, err := url.Parse(idpURL)
+	if err != nil {
+		panic(err)
+	}
+	if u.Path != openIDConfigPath {
+		return fmt.Sprintf("%s://%s%s", u.Scheme, u.Hostname(), openIDConfigPath)
+	}
+	return u.String()
```
harshavardhana, line 44: We should just take full endpoint, there is no reason for us to auto construct  `.well-known/openid-configuration` this is not a standard and it can change for various implementations.

harshavardhana (changes requested):

pkg/auth/idp/oauth2/config.go
Alevsk: @harshavardhana I added this because right now there could be deployments using just hostname URL (previous behavior) and this will break their login flow (happened to me locally)

rikatz: We need to take care on this "hardcoding". There are some specific cases where the well-known is not under the root directory (my case :) ) but under another path.

What if you check if there is only one hostname (if u.Path == "/") and if so, you  keep the old behavior, otherwise you use the whole "ConsoleIDPURL"?

harshavardhana: That's why the previous approach was wrong, to begin with hardcoding the "OIDC" config endpoint. We can break and move on @Alevsk - there is no point backpedaling here.

Alevsk: fixed it

harshavardhana (approved):

dvaldivia (changes requested):
On a working OpenID setup, it's breaking with the following error 
```
I: 2021/09/20 23:03:27 Serving operator at http://[::[]:9090
E: 2021/09/20 23:03:33 Refreshing subnet license failed: Secret "subnet-license" is invalid: data: Forbidden: field is immutable when `immutable` is set
E: 2021/09/20 23:03:42 original error -> (operator_login.go:44: invalid character '<' looking for beginning of value)
E: 2021/09/20 23:03:45 original error -> (operator_login.go:44: invalid character '<' looking for beginning of value)
```

dvaldivia (approved):
Tested with updated configuration, but we should document this on Operator `upgrade.md`

PR #1081, opened 2021-09-23
Minor Typo Fix

ravindk89:
Note that there is a fragment `buckets-assignation` that I don't know where to change, but given that its a component, it might be fine.


harshavardhana (approved):

bexsoft (approved):
LGTM

PR #1136, opened 2021-10-21
Only display Manage Bucket button if user has admin policy

jinapurapu:
Manage Bucket button is only visible on List Buckets screen if logged in user has admin policy set.

Screenshot from 2021-10-21 13-52-36
Screenshot from 2021-10-21 13-52-11


dvaldivia:
closed in favor of https://github.com/minio/console/pull/1162

PR #1167, opened 2021-10-29
Align right radio groups

dvaldivia:
Screen Shot 2021-10-29 at 2 04 15 PM


bexsoft (changes requested):
Please review design with Oscar

bexsoft (approved):
Verified style change with Oscar, seems ok

Alevsk (approved):

PR #1171, opened 2021-11-01
[WIP] Groups page ux update

prakashsvmx:
UX Link:
https://xd.adobe.com/view/4c56f281-5936-4ad2-9019-b2e230dfc8b4-1905/screen/897d44cb-47ab-4ccf-bac7-d549bd12c59f/specs/

Refactor Groups page as per UX.
- Add/Edit Group
- Assign policies to Groups
- Delete Group

### DIFF
--- a/portal-ui/src/screens/Console/Groups/GroupHelpBox.tsx
+++ b/portal-ui/src/screens/Console/Groups/GroupHelpBox.tsx
@@ -1,0 +1,49 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2021 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import React, { Fragment } from "react";
+import { GroupsIcon } from "../../../icons";
+import HelpBox from "../../../common/HelpBox";
+
+const GroupHelpBox = () => {
+  return (
+    <HelpBox
+      title={"Groups"}
+      iconComponent={<GroupsIcon />}
+      help={
+        <Fragment>
+          A group can have one attached IAM policy, where all users
+          with membership in that group inherit that policy. Groups
+          support more simplified management of user permissions on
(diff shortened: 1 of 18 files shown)


dvaldivia (approved):

portal-ui/src/screens/Console/Groups/GroupHelpBox.tsx
```diff
@@ -0,0 +1,49 @@
+// This file is part of MinIO Console Server
```
bexsoft, line 1: This file is not necessary as HelpBox is only called once with this configuration

portal-ui/src/screens/Console/Groups/GroupPageHeader.tsx
```diff
@@ -0,0 +1,44 @@
+// This file is part of MinIO Console Server
```
bexsoft, line 1: This file is not necessary as PageHeader is only called once with this configuration

portal-ui/src/screens/Console/Groups/GroupPolicies.tsx
```diff
@@ -0,0 +1,184 @@
...
+    groupName: selectedGroup
+  } = useParams<{ groupName?: string }>();
+
+  const {
+    error,
+    result = {},
+    execute: refreshGroupInfo
+  } = useAsync<GroupInfo>(fetchGroupInfo, [selectedGroup]);
+
+  const dispatch = useDispatch();
```
bexsoft, line 64: Same case, please lets use HOC instead of calling useDispatch like this

portal-ui/src/screens/Console/Groups/ReturnToGroups.tsx
```diff
@@ -0,0 +1,71 @@
+// This file is part of MinIO Console Server
```
bexsoft, line 1: This component is not necessary as we already have a component called BackLink. Please use that one instead

portal-ui/src/screens/Console/Groups/Styles.tsx
```diff
@@ -0,0 +1,97 @@
+// This file is part of MinIO Console Server
```
bexsoft, line 1: These styles can be added into styleLibrary.ts

portal-ui/src/screens/Console/Groups/SearchInput.tsx
```diff
@@ -0,0 +1,85 @@
+// This file is part of MinIO Console Server
```
bexsoft, line 1: Can we make this component more generic, so we can send the entity name and the onChange function. This way we can implement it in other parts of the application

bexsoft (changes requested):

bexsoft (changes requested):
Please review groups saving, as we should get redirected to main groups page after saving

prakashsvmx:
@bexsoft Thank you for the review.
I will be re working on this to make it similar to Users Page.

prakashsvmx:
Closing this in favour of PR #1183. Since the changes here are no more required.

PR #1202, opened 2021-11-09
STS session token and console session cookie have same duration

Alevsk:

### DIFF
--- a/restapi/config.go
+++ b/restapi/config.go
@@ -23,7 +23,6 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
-	"time"
 
 	miniov2 "github.com/minio/operator/pkg/apis/minio.min.io/v2"
 
@@ -46,9 +45,6 @@ var (
 
 	// TLSRedirect console tls redirect rule
 	TLSRedirect = "on"
-
-	// SessionDuration cookie validity duration
-	SessionDuration = 45 * time.Minute
 )
 
 func getMinIOServer() string {
--- a/restapi/utils.go
+++ b/restapi/utils.go
@@ -23,6 +23,8 @@ import (
 	"os"
 	"strings"
 	"time"
+
+	xjwt "github.com/minio/console/pkg/auth/token"
 )
 
 // Do not use:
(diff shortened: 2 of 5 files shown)


harshavardhana (approved):

restapi/utils.go
```diff
@@ -106,12 +107,14 @@ func FileExists(filename string) bool {
 }
 
 func NewSessionCookieForConsole(token string) http.Cookie {
+	sessionDuration := GetConsoleCookieDurationInSeconds()
+	expiresAt := time.Duration(int64(time.Second) * int64(ranx.Int31n(int32(sessionDuration))))
```
dvaldivia, line 111: This random can cause the sessions to expire in 1 second (or a few seconds) up to session duration, do we really need this? we should have a floor

Alevsk: This is not random, it will give you always the same session duration time based on the env variable, ie: https://play.golang.org/p/KtJtS9oRQXq

harshavardhana, line 111: This looks wrong in-fact this should be quite simply already in duration when returning.  The value can change if the console is restarted or server is restarted.

restapi/config.go
```diff
@@ -46,11 +45,17 @@ var (
...
 
+// GetConsoleCookieDurationInSeconds returns the default session duration for the console cookie in seconds (default 3600 seconds)
+func GetConsoleCookieDurationInSeconds() int {
+	duration, err := strconv.Atoi(env.Get(ConsoleSTSDurationSeconds, "3600"))
+	if err != nil {
+		duration = 3600
+	}
+	return duration
+}
+
```
harshavardhana, line 58:
```suggestion
// GetConsoleCookieDuration returns the default session duration for the console cookie in seconds (defaults to 1h)
func GetConsoleCookieDuration() time.Duration {
	duration, err := time.ParseDuration(env.Get(ConsoleSTSDurationSeconds, "1h"))
	if err != nil {
		duration = 1 * time.hour
	}
	return duration
}
```

It is better done this way @Alevsk we can break away from the current implementation.

restapi/utils.go
```diff
@@ -106,12 +107,14 @@ func FileExists(filename string) bool {
...
 	return http.Cookie{
 		Path:     "/",
 		Name:     "token",
 		Value:    token,
-		MaxAge:   int(SessionDuration.Seconds()), // 45 minutes
-		Expires:  time.Now().Add(SessionDuration),
+		MaxAge:   sessionDuration, // default 1 hr
+		Expires:  time.Now().Add(expiresAt * time.Second),
```
harshavardhana, line 117:
```suggestion
		MaxAge: int(sessionDuration.Seconds()),
		Expires:  time.Now().Add(sessionDuration),
```

Also we can only set one we don't need to set both

```
   If a cookie has both the Max-Age and the Expires attribute, the Max-
   Age attribute has precedence and controls the expiration date of the
   cookie.  If a cookie has neither the Max-Age nor the Expires
   attribute, the user agent will retain the cookie until "the current
   session is over" (as defined by the user agent).
```

https://datatracker.ietf.org/doc/html/rfc6265

harshavardhana (changes requested):

restapi/utils.go
adfost: Is it best though to use a pseudorandom number generator here? It just doesn't seem like good practice

harshavardhana: No it is not @adfost

harshavardhana (approved):

dvaldivia (approved):

harshavardhana:
An integration test has failed after this PR was merged.

PR #1222, opened 2021-11-12
Changed react-scripts to be only dev dependency

bexsoft:
## What does this do?

Forces updates on react-scripts sub-dependencies as mentioned in https://github.com/facebook/create-react-app/issues/11174

Followed resolutions method: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

**Note:** There is one dependency that was not updated as no patch is available at this time.

Also moved react-scripts dependency to be dev only.

## How does it look?
 
Screen Shot 2021-11-11 at 22 01 56

### DIFF
--- a/portal-ui/package.json
+++ b/portal-ui/package.json
@@ -36,8 +36,6 @@
     "lodash": "^4.17.21",
     "moment": "^2.29.1",
     "react": "^17.0.2",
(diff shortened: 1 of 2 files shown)


prakashsvmx (approved):
Tested. Changes look good to me.

Alevsk (approved):

PR #1235, opened 2021-11-18
Adding Download Directory to Console

### LINKED ISSUE #481: Add support to download folders
perhaps we can stream files into a zip stream

adfost:
fixes #481 

Allows directories to be downloaded as zip files from console.


restapi/user_objects.go
```diff
@@ -305,26 +317,73 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
...
+		objects, err := listBucketObjects(ctx, minioClient, params.BucketName, prefix, true, false, false)
+		if err != nil {
+			return nil, prepareError(err)
+		}
+		var folder string
+		folders := strings.Split(prefix, "/")
+		if len(folders) > 1 {
+			folder = folders[len(folders)-2]
+		}
+		fmt.Println(folders)
```
Alevsk: this can be removed

restapi/user_objects.go
```diff
@@ -305,26 +317,73 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
 func getDownloadObjectResponse(session *models.Principal, params user_api.DownloadObjectParams) (io.ReadCloser, *models.Error) {
 	ctx := context.Background()
 	var prefix string
-	if params.Prefix != "" {
```
Alevsk, line 308: Why are we not validating for an empty params.Prefix?

restapi/user_objects.go
```diff
@@ -305,26 +317,73 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
...
+		folders := strings.Split(prefix, "/")
+		if len(folders) > 1 {
+			folder = folders[len(folders)-2]
+		}
+		fmt.Println(folders)
+		w := new(bytes.Buffer)
+		zipw := zip.NewWriter(w)
+		for i := 0; i < len(objects); i++ {
+			name := folder + objects[i].Name[len(prefix)-1:]
+			s3Client, err := newS3BucketClient(session, params.BucketName, objects[i].Name)
```
Alevsk, line 345: Do we need a new s3 client for every object we want to download? why if there are 100,000 objects? check if is possible to reuse 1 client

restapi/user_objects.go
```diff
@@ -305,26 +317,73 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
...
+			object, err := downloadObject(ctx, mcClient, params.VersionID)
+			if err != nil {
+				return nil, prepareError(err)
+			}
+			f, err := zipw.Create(name)
+			if err == nil {
+				buf := new(bytes.Buffer)
+				buf.ReadFrom(object)
+				f.Write(buf.Bytes())
+			}
```
Alevsk, line 359: Validate the other way around, currently your code is swallowing errors

```suggestion
			if err != nil {
                             return nil, prepareError(err)
			}
			buf := new(bytes.Buffer)
			buf.ReadFrom(object)
			f.Write(buf.Bytes())
```

restapi/user_objects.go
```diff
@@ -305,26 +317,73 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
...
+			mcClient := mcClient{client: s3Client}
+			object, err := downloadObject(ctx, mcClient, params.VersionID)
+			if err != nil {
+				return nil, prepareError(err)
+			}
+			f, err := zipw.Create(name)
+			if err == nil {
+				buf := new(bytes.Buffer)
+				buf.ReadFrom(object)
+				f.Write(buf.Bytes())
```
Alevsk, line 358:
```suggestion
				objectBytes, err := ioutil.ReadAll(object)
				if err != nil {
                                     return nil, prepareError(err)
			        }
				f.Write(objectBytes)
```

restapi/user_objects.go
```diff
@@ -305,26 +317,73 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
...
+			return nil, prepareError(err)
+		}
+		// create a mc S3Client interface implementation
+		// defining the client to be used
+		mcClient := mcClient{client: s3Client}
+		object, err := downloadObject(ctx, mcClient, params.VersionID)
+		if err != nil {
+			return nil, prepareError(err)
+		}
+		return object, nil
```
Alevsk: There's no need on duplicating code here, I think your logic can be refactored and just add the special `IsFolder` case inside the existing `if params.Prefix != "" {` code

Alevsk (changes requested):

restapi/user_objects.go
```diff
@@ -305,6 +317,51 @@ func listBucketObjects(ctx context.Context, client MinioClient, bucketName strin
 func getDownloadObjectResponse(session *models.Principal, params user_api.DownloadObjectParams) (io.ReadCloser, *models.Error) {
 	ctx := context.Background()
 	var prefix string
+	if *params.IsFolder {
+		encodedPrefix := SanitizeEncodedPrefix(params.Prefix)
+		decodedPrefix, err := base64.StdEncoding.DecodeString(encodedPrefix)
```
Alevsk, line 322: You need to check here `if params.Prefix != "" {` as well, prefix can be empty and in this case the line above

```
name := folder + objects[i].Name[len(prefix)-1:]
```

May cause `NPE`, I think you can refactor your code and just add special logic to either produce a zip file or return a single object, both of them are type `io.ReadCloser` so there should not be a problem

Alevsk (dismissed):

restapi/user_objects.go
adfost: causes an error, reverting this

restapi/user_objects.go
adfost: Not sure, couldn't find a function to do this?

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
```diff
@@ -699,14 +699,28 @@ const ListObjects = ({
...
+      }
 
-    download(bucketName, encodeFileName(object.name), object.version_id);
+      download(
+        bucketName,
+        encodeFileName(object.name),
+        object.version_id,
+        false
+      );
+    }
```
bexsoft, line 723: This can be rewritten as follows:

```suggestion
    if (object.size > 104857600) {
        // If file is bigger than 100MB we show a notification
        setSnackBarMessage(
          "Download process started, it may take a few moments to complete"
        );
      }

      download(
        bucketName,
        encodeFileName(object.name),
        object.version_id,
        object.name.endsWith("/")
      );
```

bexsoft (changes requested):

swagger-console.yml
```diff
@@ -418,6 +418,11 @@ paths:
           required: false
           type: boolean
           default: false
+        - name: isFolder
```
dvaldivia, line 421: this can be discovered by the server, no need for the client to indicate it, else we have an undefined behavior if someone downloads an object passing `isFolder=true`

dvaldivia (changes requested):

portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
adfost: fixed

restapi/user_objects.go
```diff
@@ -313,6 +325,46 @@ func getDownloadObjectResponse(session *models.Principal, params user_api.Downlo
...
+		w := new(bytes.Buffer)
+		zipw := zip.NewWriter(w)
+		for i := 0; i < len(objects); i++ {
+			name := folder + objects[i].Name[len(prefix)-1:]
+			s3Client, err := newS3BucketClient(session, params.BucketName, objects[i].Name)
+			if err != nil {
+				return nil, prepareError(err)
+			}
+			mcClient := mcClient{client: s3Client}
+			object, err := downloadObject(ctx, mcClient, params.VersionID)
```
dvaldivia, line 352: Not all objects have the same `versionID`

dvaldivia (changes requested):

restapi/user_objects.go
adfost: outdated

swagger-console.yml
adfost: done

restapi/user_objects.go
adfost: Fixed

bexsoft:
Waiting for #1247 to get merged so we can test on 9090

bexsoft (changes requested):
This works fine with folders, but now when I try to download a single file, I get an empty file 
Screen Shot 2021-11-23 at 13 07 38

adfost:
> This works fine with folders, but now when I try to download a single file, I get an empty file Screen Shot 2021-11-23 at 13 07 38

Thought this was working and tested this but I'll check again.

bexsoft (approved):
LGTM

prakashsvmx:
When i click on download at a prefix or any file, it redirects to dashboard: ( Do i need any special setting/env variable to make it work ?) Ubuntu/Chrome


image


E.g: this is the url when i click on download on a prefix: http://localhost:5005/api/v1/buckets/testbucket/objects/download?prefix=bXktcHJlZml4Lw==

dvaldivia:
@prakashsvmx make assets and test on port 9090

prakashsvmx (approved):
Tested 
Single File download
Folder (Prefix) Download

Changes look good to me.

dvaldivia (approved):

ppizzo:
Are there any plans for multiple file download inside the same directory? My team is widely using this feature with the old version of the console and the lack of this feature in the new one is preventing us from upgrading.

Many thanks for your support.

PR #1247, opened 2021-11-22
Define base for assets and support for sub path

dvaldivia:
Defines a strict `<base href="/" />` tag to load all assets in relation to the root
Introduces support for the parameter `cpb` when requesting the index from the target console server, so the hop can replace this in the `<base />` tag
Introduces `CONSOLE_SUBPATH` to instruct the console server to set a base to that value

### To test this

You need the assets, so `make assets` before testing this PR.

### DIFF
--- a/operatorapi/proxy.go
+++ b/operatorapi/proxy.go
@@ -185,7 +185,8 @@ func serveProxy(responseWriter http.ResponseWriter, req *http.Request) {
 		responseWriter.WriteHeader(500)
 		return
 	}
-	targetURL.Path = strings.Replace(req.URL.Path, fmt.Sprintf("/api/proxy/%s/%s", tenant.Namespace, tenant.Name), "", -1)
+	tenantBase := fmt.Sprintf("/api/proxy/%s/%s", tenant.Namespace, tenant.Name)
+	targetURL.Path = strings.Replace(req.URL.Path, tenantBase, "", -1)
 
 	proxiedCookie := &http.Cookie{
 		Name:     "token",
@@ -207,8 +208,17 @@ func serveProxy(responseWriter http.ResponseWriter, req *http.Request) {
 		CheckRedirect: func(req *http.Request, via []*http.Request) error {
 			return http.ErrUseLastResponse
 		}}
+
+	// are we proxying something with cp=y? (console proxy) then add cpb (console proxy base) so the console
+	// on the other side updates the <base href="" /> to this value overriding sub path or root
+	if v := req.URL.Query().Get("cp"); v == "y" {
+		q := req.URL.Query()
+		q.Add("cpb", tenantBase)
+		req.URL.RawQuery = q.Encode()
+	}
 	// copy query params
 	targetURL.RawQuery = req.URL.Query().Encode()
+
 	proxRequest, err := http.NewRequest(req.Method, targetURL.String(), req.Body)
 	if err != nil {
 		log.Println(err)
--- a/portal-ui/package.json
+++ b/portal-ui/package.json
@@ -1,7 +1,7 @@
 {
   "name": "portal-ui",
   "version": "0.1.0",
-  "homepage": "/",
+  "homepage": ".",
   "private": true,
   "dependencies": {
     "@date-io/moment": "1.x",
--- a/portal-ui/package.json
+++ b/portal-ui/package.json
@@ -1,7 +1,7 @@
 {
   "name": "portal-ui",
   "version": "0.1.0",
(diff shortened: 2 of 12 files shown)


harshavardhana (approved):

bexsoft (changes requested):
There is an issue with HOP, Seems like server is returning some sort of empty page when trying to use this functionality. The rest of modules are working fine
Screen Shot 2021-11-22 at 12 55 45
Screen Shot 2021-11-22 at 12 55 15
Screen Shot 2021-11-22 at 12 55 08
Screen Shot 2021-11-22 at 12 54 55
Screen Shot 2021-11-22 at 12 58 51

bexsoft (approved):
Tested locally, LGTM, For HOP functionality had a demo with @dvaldivia

prakashsvmx:
Gets into this state: (2 Headers)

Steps:
Create a tenant -> View Details - Back -> Manage Tenant 

<summary> Manage Tenant <summary>

<details>

weired_state


</details>

dvaldivia:
@prakashsvmx it's hard to test this on the yarn server at `5005`, try using the go server on `9090`, did you `make assets` before running the go server?

harshavardhana (approved):

prakashsvmx (approved):
It works fine in  9090.

RafalSkolasinski:
Is it known when will this make it to the release? I tried with latest image (`RELEASE.2021-11-24T23-19-33Z`) from https://quay.io/repository/minio/minio and it does not seem to work

harshavardhana:
> Is it known when will this make it to the release? I tried with latest image (`RELEASE.2021-11-24T23-19-33Z`) from https://quay.io/repository/minio/minio?tag=latest&tab=tags and it does not seem to work

@RafalSkolasinski we haven't added support for this in MinIO yet. This is not a high-priority item for us at the moment. 

We will get it to it when we can.

harlequin:
I have tried today with the subpath feature ... somehow it's not working

My goal is to have a url like

https://domain.de/minio <- For UI/console access

Can somebody provide an example how to use?

Dragos0000:
@dvaldivia Where do we have to put this `CONSOLE_SUBPATH` and what should be the value?
I am trying to have nginx as a proxy and access minio console by having /minio instead of :9001 using docker for deployment.
It seems it doesn't load any of the static resources as is looking in the wrong place and it is not adding /minio/ in the requests.
docker compose file:
```version: '3'

services:
  minio-s3:
    image: minio/minio:latest
    ports:
      - ${MINIO_PORT}:9000
      - 9001:9001
    environment:
      - MINIO_ACCESS_KEY=${MINIO_ACCESS_KEY}
      - MINIO_SECRET_KEY=${MINIO_SECRET_KEY}
      - CONSOLE_SUBPATH=/minio/
    command:
      server --console-address :9001 /data
    volumes:
      - s3-disk:/data 
```
nginx config

```
events {
}

http {
  error_log /etc/nginx/error_log.log warn;
  client_max_body_size 20m;
  
  proxy_cache_path /etc/nginx/cache keys_zone=one:500m max_size=1000m;

  server {
    server_name localhost;

    location /minio {
       rewrite ^/minio(.*)$ /$1 break;

       proxy_set_header X-Real-IP $remote_addr;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header X-Forwarded-Proto $scheme;
       proxy_set_header Host $http_host;

       proxy_connect_timeout 300;
       # Default is HTTP/1, keepalive is only enabled in HTTP/1.1
       proxy_http_version 1.1;
       proxy_set_header Connection "";
       chunked_transfer_encoding off;

       proxy_pass http://minio-s3:9001; # If you are using docker-compose this would be the hostname i.e. minio
       # Health Check endpoint might go here. See https://www.nginx.com/resources/wiki/modules/healthcheck/
       # /minio/health/live;
    }

     listen 80;
     # To allow special characters in headers
     ignore_invalid_headers off;
     # Allow any size file to be uploaded.
     # Set to a value such as 1000m; to restrict file size to a specific value
     client_max_body_size 1000m;
     # To disable buffering
     proxy_buffering off;

  }  
}
```

PedroHRDiniz:
Hi @dvaldivia. I'm in the same situation as @Dragos0000. Could you please give us an example of _CONSOLE_SUBPATH_ configuration?

dvaldivia:
@PedroHRDiniz @Dragos0000 should be fixed by https://github.com/minio/console/pull/1384

PR #1250, opened 2021-11-23
Lazy load components

prakashsvmx:
Lazy load nested components and Modals using code splitting with reusable React Suspense HOC

Impacted Screens

Users
Groups
Service Accounts
Policies
Bucket Details

Tenants
Storage
Notification End points


bexsoft (approved):
LGTM

dvaldivia (approved):

PR #1276, opened 2021-12-01
Fix after change in policy info response

donatello:


donatello:
Closing this, as there is a compatibility issue where console is deployed separately from minio. We can move to the new API after releasing the new info policy API in minio, so that lets users update minio without breaking console.

PR #1294, opened 2021-12-06
Rewind mode list directory bug fix

adfost:
Prevents slashes from being deleted in prefix, so folders are not mistaken for objects.

### DIFF
--- a/restapi/client.go
+++ b/restapi/client.go
@@ -392,6 +392,9 @@ func newS3BucketClient(claims *models.Principal, bucketName string, prefix strin
 	}
 	if strings.TrimSpace(prefix) != "" {
 		u.Path = path.Join(u.Path, prefix)
+		if prefix[len(prefix)-1] == '/' {
+			u.Path += "/"
+		}
 	}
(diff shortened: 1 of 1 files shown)


restapi/client.go
```diff
@@ -392,6 +392,9 @@ func newS3BucketClient(claims *models.Principal, bucketName string, prefix strin
 	}
 	if strings.TrimSpace(prefix) != "" {
 		u.Path = path.Join(u.Path, prefix)
+		if prefix[len(prefix)-1] == '/' {
+			u.Path += "/"
```
harshavardhana, line 396:
```suggestion
                if strings.HasSuffix(prefix, "/") {
			u.Path += "/"
```

harshavardhana (changes requested):
Write this in a function

```
const slashSeparator = "/"

// pathJoin - like path.Join() but retains trailing SlashSeparator of the last element
func pathJoin(elem ...string) string {
        trailingSlash := ""
        if len(elem) > 0 {
                if strings.HasSuffix(elem[len(elem)-1], slashSeparator) {
                        trailingSlash = slashSeparator
                }
        }
        return path.Join(elem...) + trailingSlash
}
```

PR #1297, opened 2021-12-06
Rewind mode list directory bug fix

adfost:
Prevents slashes from being deleted in prefix, so folders are not mistaken for objects.

### DIFF
--- a/restapi/client.go
+++ b/restapi/client.go
@@ -389,7 +389,7 @@ func computeObjectURLWithoutEncode(bucketName, prefix string) (string, error) {
 		objectURL = path.Join(objectURL, bucketName)
 	}
 	if strings.TrimSpace(prefix) != "" {
-		objectURL = path.Join(objectURL, prefix)
+		objectURL = pathJoinFinalSlash(objectURL, prefix)
 	}
 
 	objectURL = fmt.Sprintf("%s://%s", u.Scheme, objectURL)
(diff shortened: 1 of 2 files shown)


restapi/client.go
```diff
@@ -418,6 +418,16 @@ func newS3BucketClient(claims *models.Principal, bucketName string, prefix strin
 	return s3Client, nil
 }
 
+// pathJoinFinalSlash - like path.Join() but retains trailing slashSeparator of the last element
+func pathJoinFinalSlash(elem ...string) string {
+	if len(elem) > 0 {
+		if strings.HasSuffix(elem[len(elem)-1], "/") {
```
bexsoft, line 424: This is not necessary, HasSuffix returns true of false for the whole string, you don't need to get the lat character for this, https://go.dev/play/p/BlT26DEoXXN

 ```suggestion
		if strings.HasSuffix(elem, "/") {
```

**Review by bexsoft — CHANGES_REQUESTED** [review]
Please review comments

File: restapi/client.go (hunk 1 above)
Comment (harshavardhana) [reply restapi/client.go:424]: It is an optimization @bexsoft to avoid through the entire string - look for only what you are interested

File: restapi/client.go [hunk 2]
```diff
@@ -418,6 +418,16 @@ func newS3BucketClient(claims *models.Principal, bucketName string, prefix strin
...
+			return path.Join(elem...) + "/"
```
harshavardhana, line 425:
```suggestion
			return path.Join(elem...) + slashSeparator
```

Prefer more constants as I mentioned before

harshavardhana (changes requested):

restapi/client.go
```diff
@@ -418,6 +418,16 @@ func newS3BucketClient(claims *models.Principal, bucketName string, prefix strin
 	return s3Client, nil
 }
 
+// pathJoinFinalSlash - like path.Join() but retains trailing slashSeparator of the last element
+func pathJoinFinalSlash(elem ...string) string {
+	if len(elem) > 0 {
+		if strings.HasSuffix(elem[len(elem)-1], SlashSeparator) {
```
dvaldivia, line 424: this is incorrect, it's a very complciatd `"/" == "/"` or last character, it should be `strings.HasSuffic(elem,SlashSeparator)`

dvaldivia (changes requested):

restapi/client.go
adfost: elem is an array of strings right?

dvaldivia: oh that's right

dvaldivia (approved):

bexsoft (approved):
LGTM

PR #1314, opened 2021-12-10
separate operator binary code to reduce binary size

harshavardhana:
remove logsearchapi dependency, no definitive struct
marshalling is needed since UI is schemaless for
search response.


bexsoft:
Is there any special configuration to run console server & operator server? If I try to run from source code I get `exit status 1` message, this happens also running from build executable (no error is seen, but application closes)

harshavardhana:
> Is there any special configuration to run console server & operator server? If I try to run from source code I get `exit status 1` message, this happens also running from build executable (no error is seen, but application closes)

Operator code won't work if not running in k8s cluster @bexsoft you can't test that locally. This build change is to allow other dependent projects to selectively ignore `operator` code. 

So if you do `make install` it always builds with operator to keep backward compatibility.

dvaldivia:
can you rebase @harshavardhana ?

harshavardhana:
> can you rebase @harshavardhana ?

done @dvaldivia

dvaldivia:
I guess we need an additional `make` target for the operator binary, include both binaries in the container and update the kubernetes resources to use the `./operator operator` binary instead of `./console operator` @harshavardhana , wanna do that in this PR?

harshavardhana:
> I guess we need an additional `make` target for the operator binary, include both binaries in the container and update the kubernetes resources to use the `./operator operator` binary instead of `./console operator` @harshavardhana , wanna do that in this PR?

It's not needed that is not the intention here. It is mainly meant to pave way for builds to be generated without operator code, i.e MinIO.

Binary is going be still the same for this repo, nothing changes.

dvaldivia:
one more rebase please @harshavardhana 🙏

dvaldivia (approved):

harshavardhana:
@bexsoft @Alevsk PTAL

cniackz (approved):

PR #1327, opened 2021-12-17
Chinese file name garbled

mulzhang:
Solve the problem of garbled file names when downloading Chinese files


harshavardhana (changes requested):
do not build assets - submit a proper PR @mulzhang

mulzhang:
I will resubmit PR

PR #1340, opened 2021-12-23
Autofocus support in InputBoxWrapper

dvaldivia:


prakashsvmx (approved):
LGTM 👍

bexsoft (approved):
For boolean true props, it is not necessary to set `prop={true}`, you can just add the prop without assigning any value. We can leave it like this right now. LGTM

PR #1342, opened 2021-12-25
Use default MinIO image when no prefered image is given

vadmeste:
Do not select the latest MinIO image when deploying a new image, but use
operator's hardcoded instead.

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -21,10 +21,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
-	"net/http"
 	"regexp"
 	"strings"
-	"time"
 
 	"github.com/minio/pkg/env"
 )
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

dvaldivia (changes requested):
I think it's better to control this via environment variable since we definitively want to encourage people to use the latest minio version when they create a new tenant, also if there's no fixed minio image we expose them to unwanted upgrades (upon upgrade of the operator) or unwanted downgrades

vadmeste:
> I think it's better to control this via environment variable since we definitively want to encourage people to use the latest minio version when they create a new tenant, 

That's not always good. Users will see issues when they create a new tenant and latest MinIO has a breaking change.

The thing is that, what we want is a stable environment whatever happens

> also if there's no fixed minio image we expose them to unwanted upgrades (upon upgrade of the operator) or unwanted downgrades

This looks like Operator issue and not Console.

dvaldivia:
closed in favor of `CONSOLE_MINIO_IMAGE`

PR #1350, opened 2021-12-30
Enable versioning if locking is enabled on bucket creation.

adfost:
Versioning is enabled automatically if locking is enabled when creating buckets.


prakashsvmx (approved):
Tested. Changes look good to me.  :+1: 

We need to make the versioning Enabled label to be read only in bucket summary page if object locking is enabled.  as of now it is enabled and when  updated , the error message detail indicates that locking is enabled. so it is just a ux update and can be done in a different PR.

dvaldivia (approved):

Issue #1148, opened 2021-10-25
Getting 413 when uploading an object doesn't show an error


dvaldivia:
This headers indicate the `payload too large` but the UI was not showing an error
```
:status: 413
Content-Type: text/html
Content-Length: 176
Date: Mon, 25 Oct 2021 22:24:29 GMT
Strict-Transport-Security: max-age=15724800; includeSubDomains
```
body
```
<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx</center>
</body>
</html>

```

PR #1379, opened 2022-01-11
Added informative message for 413 error file too large

jinapurapu:
Displays informative error message in browser when upload fails due to overlarge size. Closes Issue #1148
Screenshot from 2022-01-11 12-24-06
.

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
@@ -761,6 +761,9 @@ const ListObjects = ({
                 ) {
                   setSnackBarMessage(errorMessage);
                 }
+                if (xhr.status === 413) {
+                  setSnackBarMessage("Error - File size too large");
+                }
                 if (xhr.status === 200) {
                   completeObject(identity);
                   if (files.length === 0) {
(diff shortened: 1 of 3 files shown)


dvaldivia (approved):
LGTM.

prakashsvmx (approved):
LGTM. :+1:

sang8052:
Hi,gays!Why not console support upload file with [upload_part](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html)? I use nginx proxy my console , so when i upload large file , it will give message "413 file too large!"

ramondeklein:
The problem is that NGINX buffers the file during upload, so they only way to prevent this is to move the uploading part to the client (Javascript) and send the data in blocks. This effectively requires the client to do the following steps:
1. Client requests a presigned URL that allows uploading to the specific bucket/key. You don't want the user's credentials to be available in the browser.
2. Perform the multi-part upload from Javascript.

Although possible, it's more complicated, because all the smart handling should be done client-side. The current solution just uses a form-based file posting and the MinIO Go SDK will take care of the multi-part upload in the backend. This keeps front-end code simple and is a design choice. The console is created to perform some administrative tasks and is not intended to be used as a default front-end for the object-storage.

You may want to try the following NGINX settings to disable request body buffering ([source](https://serverfault.com/a/818090/209344)):
```
client_max_body_size 0;
proxy_http_version 1.1;
proxy_request_buffering off;
```

PR #1380, opened 2022-01-11
Activate Lifecycle panel for all buckets

t0xicCode:
Enables the Lifecycle panel for all buckets, as MinIO supports lifecycle
operations on buckets regardless of the mode. This brings parity between
the console and the mc tool, as the mc tool already supports managing
lifecycle policies on all buckets.

~[!] Note: the stable ID for the assets changed, not sure if it's due to the version of yarn/node that I used or if it's something else. If you have insight, please let me know and I'll amend my commit.~


harshavardhana:
> [!] Note: the stable ID for the assets changed, not sure if it's due to the version of yarn/node that I used or if it's something else. If you have insight, please let me know and I'll amend my commit.

Do not generate assets, we do not take externally generate assets in PRs @t0xicCode please only submit the relevant code.

t0xicCode:
@harshavardhana updated the commit, only the typescript change is included

harshavardhana (approved):

dvaldivia:
Do we support this on non-distributed setups @harshavardhana ?

t0xicCode:
> Do we support this on non-distributed setups 

The mc client and minio itself do support the lifecycle options (if only the automatic deletion of files).

cniackz:
Even when I tried to test this locally, I see the potential of having this enabled for the non-distributed mode. To me this change looks good. In the meantime I will keep trying to test this locally to learn more on how to test this type of changes in my laptop during the review process...

cniackz (approved):
Looks good to me

bexsoft (changes requested):
Need to verify lifecycle as by initial development it was a requirement that it should only run in distributed mode.

cniackz:
Ok, sorry, I have been informed we are not supposed to support this in not distributed mode.
So I will stop with the review of this until further discussion from the team.

t0xicCode:
Based on the documentation at
https://docs.min.io/docs/minio-bucket-lifecycle-guide.html and minio
itself, basic lifecycle does not require a distributed setup. Only the
transition feature require Erasure/distributed mode.

Should this PR be updated to put the transition features behind the
distributed gate or is the end goal to remove lifecycling from minio in
non-distributed mode altogether?

prakashsvmx:
Closing this, as Life cycle would not be officially supported in standalone mode. 
thank you. @t0xicCode  :+1:

WolfspiritM:
I've just created https://github.com/minio/console/issues/1521 but then I found this.
@prakashsvmx Why is that not officially supported in standalone mode? Where is that documented? As @t0xicCode said the official documentation at https://docs.min.io/docs/minio-bucket-lifecycle-guide.html doesn't mention distributed mode at all and we're using lifecycle policies for testing environments for years. The only problem is that it's not yet visible in the console sadly.

harshavardhana:
> I've just created #1521 but then I found this. @prakashsvmx Why is that not officially supported in standalone mode? Where is that documented? As @t0xicCode said the official documentation at https://docs.min.io/docs/minio-bucket-lifecycle-guide.html doesn't mention distributed mode at all and we're using lifecycle policies for testing environments for years. The only problem is that it's not yet visible in the console sadly.

Its not going to be supported until few changes. Wait for server side changes.

PR #1387, opened 2022-01-13
Added Prometheus monitoring CPU and memory selector.

jinapurapu:
Added CPU and memory resource request fields for Prometheus monitoring

Screenshot from 2022-01-12 16-03-03
Screenshot from 2022-01-12 16-05-20


dvaldivia (changes requested):
memory needs a unit, if we request `2` the value should be `2Gi` in the CRD, also if the storage class is empty don't attach it

```
  prometheus:
    diskCapacityGB: 5
    resources:
      requests:
        cpu: "2"
        memory: "2"
    storageClassName: ""
```

dvaldivia (approved):
LGTM. Tested.

portal-ui/src/screens/Console/Tenants/TenantDetails/EditTenantMonitoringModal.tsx
```diff
@@ -129,6 +137,20 @@ const EditTenantMonitoringModal = ({
       pattern: /^[0-9]?(10)?$/,
       customPatternMessage: "Must be an integer between 0 and 10",
     });
+    tenantMonitoringValidation.push({
+      fieldKey: `newCPURequest`,
+      required: false,
+      value: newCPURequest as any as string,
+      pattern: /^[0-9]?(10)?$/,
```
cniackz, line 144: Why the limit is 10?, where do I can see this limation coming from?

cniackz (approved):
Looks good to me, just have a question, why the limit of requested CPUs is 10? Can it be 11?

portal-ui/src/screens/Console/Tenants/TenantDetails/EditTenantMonitoringModal.tsx
jinapurapu: This limit was from a validation I reused, no actual reason. Please let me know if I should change it.

dvaldivia:
@cniackz that is a nice catch, why are we limiting to upper 10 @jinapurapu ? it's not hat critical, but it's good to know

PR #1388, opened 2022-01-13
to fail test when yarn does not build

cniackz:
The idea is to fail the test when `yarn build` does not work
Currently, test is passing because we are not providing an exit when yarn is failing to build.
Fixes miniohq/engineering#362


dvaldivia:
nice!

bexsoft (approved):
LGTM

dvaldivia:
sadly we can't merge until someone fixes the warnings in `master` hahaha

bexsoft (approved):

dvaldivia (approved):

PR #1391, opened 2022-01-13
update console to releave v7.0.21

harshavardhana:
fixes an issue when STS error response has
S3 error response structure, this can happen
when server returns an S3 Error instead of
STS Error for STS calls.


dvaldivia (approved):

prakashsvmx (approved):
LGTM. :+1:

dvaldivia (approved):

PR #1401, opened 2022-01-14
Subnet cluster registration

dvaldivia:
- Removed old registration flow
- Add support for new online and offline cluster registration flow
- Support login accounts with mfa enabled


prakashsvmx:
I have registered an offline/offline cluster  in SUBNET for the test account `health/1/257` the plan column in UI still shows upgrade / activate (Refreshed as well)
@Alevsk 


Suggestion: 
can we have a way to pass `--dev` as in mc command to test it locally if possible?

PR #1410, opened 2022-01-18
WIP URL and Mountpath fields added to TenantCredentials display

jinapurapu:
Adds Tenant URL and Tenant Mountpath fields to the TenantCredentials display shown at new tenant creation. 

Screenshot from 2022-01-18 14-01-08


swagger-operator.yml
```diff
@@ -1825,20 +1825,24 @@ definitions:
...
+      mountpoint:
+        type: string  
       externalIDP:
         type: boolean
       console:
         type: array
         items:
           $ref: "#/definitions/tenantResponseItem"
-
+
```
cniackz, line 1838: Maybe we should take care of those trailing whitespaces on the PR...

PR #1424, opened 2022-01-20
Make bucket Integration Test

cniackz:
To add a new test where we create or add a bucket testing with `locking` as `true` since that hasn't been tested yet. We need to make sure we can create the bucket even with this field as `true` 

Fixes miniohq/engineering#374

### DIFF
--- a/integration/buckets_test.go
+++ b/integration/buckets_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


bexsoft (dismissed):
LGTM

integration/buckets_test.go
```diff
@@ -208,6 +209,108 @@ func TestAddBucket(t *testing.T) {
...
+		"name":       "test1",
+		"versioning": false,
+		"locking":    true,
+	}
+
+	requestDataJSON, _ := json.Marshal(requestDataAdd)
+
+	requestDataBody := bytes.NewReader(requestDataJSON)
+
+	// get list of buckets
```
prakashsvmx, line 239: this code block creates a bucket so the comment can be updated accordingly.

```suggestion
	// create a object locking bucket without versioning flag
```

integration/buckets_test.go
```diff
@@ -208,6 +209,108 @@ func TestAddBucket(t *testing.T) {
...
+		"name":       "thujan",
+		"versioning": true,
+		"locking":    true,
+	}
+
+	requestDataJSON, _ = json.Marshal(requestDataAdd)
+
+	requestDataBody = bytes.NewReader(requestDataJSON)
+
+	// get list of buckets
```
prakashsvmx, line 277: Please update the comment here

integration/buckets_test.go
```diff
@@ -208,6 +209,108 @@ func TestAddBucket(t *testing.T) {
...
+	request.Header.Add("Content-Type", "application/json")
+
+	response, err = client.Do(request)
+	assert.Nil(err)
+	if err != nil {
+		log.Println(err)
+		return
+	}
+
+	// Verification part, bucket name is valid, we expect 201 when created.
```
prakashsvmx, line 295:
```suggestion
	// Verification part, bucket should be created with versioning enabled and locking enabled, we expect 201 when created.
```

prakashsvmx (approved):
minor code comment adjustments
otherwise
Changes look good to me :+1:

integration/buckets_test.go
cniackz: The change has been performed.
Thank you for the suggestion @prakashsvmx

integration/buckets_test.go
cniackz: Also this comment was updated accordingly

integration/buckets_test.go
cniackz: That is right, this is not about the name but the combination of versioning and locking when creating the bucket, thanks!

prakashsvmx (approved):

dvaldivia (approved):

PR #1430, opened 2022-01-21
Improving bucket tests in Console API

cniackz:
There is no task yet for this activity.
But I think we need to re-use the code in buckets_test.go since the same code is being called over and over in different spots. This is my first attempt to make this better. Let's see if the tests pass first.

Notice `AddBucket` is an atomic test that will only add a bucket and can be reused in many spots with the same result.


Alevsk (approved):

bexsoft (approved):
LGTM

adfost (approved):

PR #1492, opened 2022-01-29
remove quota FIFO support

harshavardhana:


dvaldivia:
@harshavardhana is this also gone from the server?

harshavardhana:
> @harshavardhana is this also gone from the server?

Yes it will be gone but need this PR first.

prakashsvmx (approved):
Tested:

- On Update of Existing Bucket with  FIFO Quota 
- On new Bucket creation
- Update of Bucket Quota post creation

Changes look good to me. :+1: 

Few UX Suggestions which can be addressed later:
- `Quota_type in body should be one of [hard]` error message can be updated  When editing an existing FIFO Quota and saving without selecting the Type. ( or we can default to hard always )
- `Quota Type` selection can be removed as it would be the default , and Quota label can be " Quota Size"

bexsoft (approved):
LGTM

PR #1500, opened 2022-02-01
(Un)Select multiple objects with single tap

tekig:
Adds the ability to (un)select multiple objects with single tap while holding the `shift` key.


tekig:
Test passed without errors
[https://github.com/tekig/console/pull/1](https://github.com/tekig/console/pull/1)

prakashsvmx:
@tekig  Thank you for the PR.

User Experience feedback:
--------------------------------------
When using `Shift` key along with `Ctrl` Key for selection/de-selection

it results in undesired selection of rows.

in case of  more rows  ( range) to be skipped, it is more difficult to get the desired selection/deselection.

- Holding and Clicking on the same Row does not un select
- `Shift` + Selection requires  at least 1 row selection without holding `Shift`
- Mix and match `Ctrl` + `Selection`  and `Shift` + Selection  is in-consistent. ( some times the selection does not respond)

prakashsvmx:
I have tested this further.
There are inconsistencies with Select/Unselect with range select / skip select. 
Please have a look.

prakashsvmx:
@tekig 
As you can observe, 
( just ignore the cursor position in the video)

some times the selection does not select or results in un expected selection:


https://user-images.githubusercontent.com/23444145/153409589-66c12aa9-93b6-4d1e-b4a0-7cf111ac5b25.mp4

tekig:
`Shift` makes the last action repeat from the previous element to the current one.
If the action was selected, it will select all elemen in rangets.
if the previous action was unselected, it will unselect all elements in range.

If that doesn't make sense, i can force Shift to only select.

prakashsvmx:
It is critical when we select and **delete** with scrolling in the list.  So it would be great to have a standard selection patterns. 
@tekig  . Thank you for the improvements.

tekig:
I don't know how to do differently. My version works like an analogue of explorer in Windows with `ctrl + shift` pressed.

I don't understand why it can be difficult to press `shift` when you need to duplicate an action on the following objects?

prakashsvmx:
@tekig Sorry for going back and forth and thank you for the patience and improvements

In Linux Files ( Explorer)  for example:

( For example, if we have files from 1-10)

Hold `Ctrl` 

Select 1, 2

Select 9, 10

Now Leave `Ctrl`   hold only `Shift`
Select 5

Observe that 5 -10 only gets selected.  
***1-2 get un-selected.***

 - this could not be achieved  in the object list

tekig:
Thank you, as there will be free time, I will try to add the ctrl key for interaction.

My idea was that selection already works like ctrl so it doesn't need to be pressed.

tekig:
I have an idea in the form of improving the selection:

- Remove checkboxes
- Change background color of selected objects
- By clicking on an object, it becomes selected

It seems to me that this is a global change, so I want to consult with you first.

prakashsvmx (changes requested):
Some times holding `ctrl`  while un-selecting requires double click which requires a fix. 
otherwise changes look good to me. :+1:  

@tekig thank you for the improvements.

tekig:
@prakashsvmx I can't reproduce this, can you write the steps to reproduce it please?

prakashsvmx:
@tekig Please resolve the conflicts , i shall upload the video to demonstrate the same.

prakashsvmx:
I see one more issue:

On initial load, hold `Shift` -> click on 2nd Row of the object List 
Actual:  2 rows get selected. 
Expected: Only one row should be selected.


https://user-images.githubusercontent.com/23444145/158326247-d1900a37-10d7-4fe5-9234-45a3e2cac473.mp4

tekig:
> I see one more issue:
> 
> On initial load, hold `Shift` -> click on 2nd Row of the object List Actual: 2 rows get selected. Expected: Only one row should be selected.
> 
>  Shift_Click_2nd_row_selects-1st_row_as_well.mp4

I specifically did that by default I chose from the beginning of the list. I will fix it soon)

prakashsvmx (approved):
@tekig 
Tested. Changes look good to me  :+1:  :
- Chrome
- Firefox
On Ubuntu

dvaldivia (changes requested):
I feel it's confusing to be using the select box UI element to indicate you can select multiple, but by default it's not selecting multiple, even if you try, I think this will confuse a lot of people. I did tried while pressing `shift` and it let me, wasn't the original goal to give an un-select capability?

dvaldivia:
I think due to the confusion of the implementation of the feature we are not going to take this PR for the time being, if you re-work it in a way that makes sense with the current multi-checkbox implementation please re-open the PR

PR #1506, opened 2022-02-02
Convert to table driven test

cniackz:
In this PR, I would like to start converting our current integration tests to be table driven tests because it will allow us to test positive case and negative case in the same function without the need of having duplicated functions that does the same with minor changes. Based on the same approach we are using for the unittest. Hope you like it and find it useful!

### DIFF
--- a/integration/buckets_test.go
+++ b/integration/buckets_test.go
@@ -545,16 +545,45 @@ func ListObjects(bucketName string, prefix string) (*http.Response, error) {
 
 func TestAddBucket(t *testing.T) {
 	assert := assert.New(t)
-
-	response, err := AddBucket("test1", false, false, nil, nil)
-	assert.Nil(err)
-	if err != nil {
-		log.Println(err)
-		return
+	type args struct {
+		bucketName string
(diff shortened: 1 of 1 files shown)


integration/buckets_test.go
```diff
@@ -545,16 +545,45 @@ func ListObjects(bucketName string, prefix string) (*http.Response, error) {
...
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{
+			name: "Add Bucket with valid name",
+			args: args{
+				bucketName:     "test1",
+				expectedStatus: 201,
```
dvaldivia, line 560: perhaps move `expectedStatus` as the same level as the test `name` since the `args` should be only input

dvaldivia (dismissed):

integration/buckets_test.go
cniackz: Ok, will do that, thanks @dvaldivia

dvaldivia (approved):

bexsoft (approved):

PR #1519, opened 2022-02-04
Service account policy UI

adfost:
UI for getting service account policies
Screen Shot 2022-02-04 at 11 04 33 AM
.


portal-ui/src/screens/Console/Account/ServiceAccountPolicy.tsx
```diff
@@ -0,0 +1,131 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2021 MinIO, Inc.
```
dvaldivia, line 2:
```suggestion
// Copyright (c) 2022 MinIO, Inc.
```

dvaldivia (changes requested):

harshavardhana:
Do we validate if its an invalid JSON in the UI?

harshavardhana:
Like do we react to what they pasted? - the way we are highlighting it has JSON, are we able to validate/parse it as well?

prakashsvmx:
@adfost Please resolve the conflicts.

prakashsvmx (approved):
Changes look good to me. :+1: 

Few minor changes can be part of #1525

Alevsk:
> Like do we react to what they pasted? - the way we are highlighting it has JSON, are we able to validate/parse it as well?

I think that highlight happens automatically by the `CodeMirror` library, currently theres no validation in place to check if whatever the user input its a valid json, however if the json is wrong minio will return an error and we will display that on the UI

Alevsk (dismissed):

dvaldivia (changes requested):
When I create a SA without policy, it shows me a weird policy, if the SA has no policy, we should show empty
Screen Shot 2022-02-08 at 10 37 18 AM

dvaldivia (approved):
Tested. LGTM

prakashsvmx (approved):
LGTM.

PR #1540, opened 2022-02-10
Inspect API

prakashsvmx:
Inspect API  ( `mc admin inspect [--encrypt] <alias> <path>` )

Steps to Test:
- With Encryption  flag
   - `.enc` file will be generated along with a decryption key 
   - use `inspect` tool and enter the decryption key to get the `.zip` file
   - use `xl-meta` tool  to verify the contents of `.zip` file
- Without Encryption flag
   - use `xl-meta` tool  to verify the contents of `.zip` file.
  

API URL : `/api/v1/admin/inspect?file=test.txt%2Fxl.meta&volume=test-bucket&encrypt=true` 

Fixes https://github.com/miniohq/engineering/issues/479


restapi/admin_inspect.go
```diff
@@ -0,0 +1,117 @@
...
+			w.Header().Set("Content-Type", "application/octet-stream")
+			w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", fileName))
+
+			_, err := io.Copy(w, r)
+
+			if err != nil {
+				LogError("Unable to write all the data: %v", err)
+			}
+		})
+	})
```
bexsoft, line 83: Is better to set this code as an independent method, Please follow the pattern from other endpoints Eg.

```
api.UserAPIUpdateBucketLifecycleHandler = user_api.UpdateBucketLifecycleHandlerFunc(func(params user_api.UpdateBucketLifecycleParams, session *models.Principal) middleware.Responder {
		err := getEditBucketLifecycleRule(session, params)
		if err != nil {
			user_api.NewUpdateBucketLifecycleDefault(int(err.Code)).WithPayload(err)
		}

		return user_api.NewUpdateBucketLifecycleOK()
	})
```

bexsoft (changes requested):

dvaldivia:
When I reach an invalid path (invalid bucket or objects) , I get status response `200` instead of `404`

harshavardhana:
> When I reach an invalid path (invalid bucket or objects) , I get status response `200` instead of `404`

The API is not designed to get 404 it's expected to see failures since it's a "debugging" API - Do not think of it like S3 API as it is allowed to download any path any file in its RAW form. 

Like you can download

- alias/.minio.sys/format.json
- alias/.minio.sys/** (recursively download all .minio.sys data from all nodes all drives)
- alias/bucket/path/** (recursively download all xl.meta, part.1 data from all nodes all drives)
- alias/bucket/path/object/* (just download files at the path - non-recursive but all files at this path)
- alias/bucket/path/object/<uuid>/part.1 (download just part.1 for a specific object data-dir at uuid)

The response is a Zip stream, so failures are allowed and normal.

harshavardhana:
The UI should be designed to support these requirements

dvaldivia (approved):

restapi/admin_inspect.go
prakashsvmx: @bexsoft  thank you for the suggestion. I will update it.

prakashsvmx: I have updated and tested . PTAL

bexsoft (approved):
LGTM

dvaldivia (approved):

PR #1547, opened 2022-02-10
Added delete bucket lifecycle rule capability

bexsoft:
## What does this do?

Adds bucket lifecycle delete rule capability

## How does it look?

Screen Shot 2022-02-10 at 15 28 58
Screen Shot 2022-02-10 at 15 29 01
Screen Shot 2022-02-10 at 15 29 05


restapi/user_buckets_lifecycle.go
```diff
@@ -356,3 +365,54 @@ func getEditBucketLifecycleRule(session *models.Principal, params user_api.Updat
...
+	}
+
+	if len(lfcCfg.Rules) == 0 {
+		return errors.New("no rules available to delete")
+	}
+
+	n := 0
+	for _, rule := range lfcCfg.Rules {
+		if rule.ID != params.LifecycleID {
+			lfcCfg.Rules[n] = rule
```
Alevsk, line 388: could `lfcCfg.Rules[n]` be undefined?

Alevsk (changes requested):

restapi/user_buckets_lifecycle.go
bexsoft: In this case no, as we are iterating in the range of Rules list. So if the list is empty, that loop won't be executed

jinapurapu (approved):

Alevsk (dismissed):

jinapurapu:
I get these errors after deleting the last rule, is that correct
Screenshot from 2022-02-10 15-57-04
Screenshot from 2022-02-10 15-56-50
?

bexsoft:
> I get these errors after deleting the last rule, is that correct Screenshot from 2022-02-10 15-57-04 Screenshot from 2022-02-10 15-56-50 ?

Yes, this error is expected as we are returning 404 when no rules are fetched. you can test this by opening this page in a clean bucket

dvaldivia (approved):

PR #1548, opened 2022-02-10
Fix NPE on permissions

dvaldivia:


Alevsk (approved):

harshavardhana (approved):

dvaldivia:
@harshavardhana we need to add tests for operator ui, which we are missing right now @harshavardhana , we have them in the queue but will prioritize them

harshavardhana:
> @harshavardhana we need to add tests for operator ui, which we are missing right now @harshavardhana , we have them in the queue but will prioritize them

👍🏽

PR #1568, opened 2022-02-14
[WIP] Add Notify Postgres End Point as a test in order to subscribe events …

cniackz:
Fixes https://github.com/miniohq/engineering/issues/433

This is still work in progress... But the idea is to be able to add events to buckets but first we need to add some notification; I decided to use postgresql notification. One container will be added and configured to share the network with minio container so we can test this api end point. For now I have only configured my docker instances locally, but I am creating this PR to see how to configure docker over GitHub Actions....

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,8 @@ test-integration:
 	@(docker network create --subnet=173.18.0.0/29 mynet123)
 	@echo "docker run with MinIO Version below:"
 	@echo $(MINIO_VERSION)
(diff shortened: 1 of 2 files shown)


cniackz:
I am currently fixing the issue I found in this PR over: https://github.com/minio/console/pull/1572

cniackz:
Will follow up on this task in 
https://github.com/minio/console/pull/1578

PR #1573, opened 2022-02-15
Add test for Put Objects legalhold status end point in Console API

cniackz:
Fixes https://github.com/miniohq/engineering/issues/385

Add test for `Put Object's legalhold status` end point in Console API


cniackz:
will handle this change on
https://github.com/minio/console/pull/1577

PR #1572, opened 2022-02-15
Adding 4 volumes to run MinIO Distributed on the tests

cniackz:

PR #1574, opened 2022-02-15
Add test for Delete Object retention from an object end point

cniackz:
Fixes https://github.com/miniohq/engineering/issues/387

Add test for `Delete Object retention from an object` end point in Console API


cniackz:
Integration test will until we get the fix from
https://github.com/minio/console/pull/1572

dvaldivia (approved):

bexsoft (approved):
LGTM

PR #1578, opened 2022-02-15
Add PostgreSQL Notification Test

cniackz:
Fixes https://github.com/miniohq/engineering/issues/520

The idea is to be able to add events to buckets but first we need to add some notification; I decided to use postgresql notification. One container will be added and configured to share the network with minio container so we can test this api end point.


prakashsvmx (dismissed):
LGTM :+1:

dvaldivia (approved):

PR #1579, opened 2022-02-15
Added delete all replication rules capability

bexsoft:
## What does this do?
Added delete all replication rules capability

## How does it look?


harshavardhana:
We should perhaps ask for more confirmation @bexsoft - asking them to type "yes" or "delete"

bexsoft:
> We should perhaps ask for more confirmation @bexsoft - asking them to type "yes" or "delete"

Fixed @harshavardhana 
Screen Shot 2022-02-15 at 14 14 56
Screen Shot 2022-02-15 at 14 14 48

swagger-console.yml
```diff
@@ -934,6 +934,12 @@ paths:
           in: path
           required: true
           type: string
+        - name: delete_all
```
dvaldivia, line 937: I think it's better to have a dedicated `delete_all` call

dvaldivia (changes requested):

swagger-console.yml
bexsoft: Resolved

dvaldivia (approved):

adfost (approved):

PR #1582, opened 2022-02-15
Giving more stability to our testcafe

cniackz:
Adding a bigger timeout selector to be able to wait enough time for each selector without the need of keep adding extra waits in the code:
https://testcafe.io/documentation/402829/guides/basic-guides/select-page-elements#selector-timeout
I just want to make our tests more stable, since right now they can fail or pass in an intermittent way.
This change may increase a bit the execution time, but it will make `testcafe` much more reliable than now.

Also remember that our GitHub Actions running on Ubuntu VM does not run as fast as our Laptops and this may explain why we get so much issues lately. Hopefully this PR will help address all these issues below:

```sh
✖ Upload File button is disable and Upload Folder button is enabled on
✖ Download button is clickable after Diagnostic is completed
✖ Upload button exists
✖ Object list table is enabled
```

Among others observed in different pull requests.


kaankabalak:
Would we need to remove the wait function invocations on `portal-ui/tests/permissions/diagnostics.ts`?

cniackz:
Yes, maybe we can remove now the `wait` function invocations on `portal-ui/tests/permissions/diagnostics.ts`; let me add that change here and see how it goes.

kaankabalak (approved):

dvaldivia (approved):

bexsoft (approved):

PR #1599, opened 2022-02-17
Adding IDs for Operator tests

cniackz:
Fixes https://github.com/miniohq/engineering/issues/534

I am preparing all CSS selectors for Operator tests.
It is easier to click or text over an element when is easy to find.
In most cases, CSS selector can be created with current attributes like:
```js
document.querySelector("a[href=\"/tenants\"]").click()
Tenant name: #tenant-name
Namespace: #namespace
Storage class: input[name="storage_class"]
Number of servers number of nodes; #nodes
#drivesps <--- Drives per Server
Total Size*: #volume_size / input[name="size_factor"]
input[name="ec_parity"] <-- Erasure Code Parity
#resourcesCPURequest
#resourcesMemoryRequest
input[name="resourcesSpecifyLimit"]
```
But in other cases, this is just not possible, and I was having my mind in Python/Selenium where Xpath Selectors were used in my other projects, I read a bit on XPATH in TestCafe and found some repos... Nevertheless, since our React architecture allow id additions, then I found this approach easier for now. Later in the future we can re-evaluate and see if we should rather use xpath selectors or keep improving our DOM Attributes.

### DIFF
--- a/portal-ui/src/screens/Console/Common/GenericWizard/WizardPage.tsx
+++ b/portal-ui/src/screens/Console/Common/GenericWizard/WizardPage.tsx
@@ -105,6 +105,7 @@ const WizardPage = ({
           {page.buttons.map((btn) => {
             return (
               <Button
+                id={"wizard-button-" + btn.label}
                 variant="contained"
                 color="primary"
                 size="small"
--- a/portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
+++ b/portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
@@ -174,6 +174,7 @@ const CredentialsPrompt = ({
         </Grid>
         <Grid item xs={12} className={classes.buttonContainer}>
           <Button
+            id={"done-button"}
             variant="outlined"
             className={classes.buttonSpacer}
             onClick={() => {
@@ -186,6 +187,7 @@ const CredentialsPrompt = ({
 
           {!idp && (
             <Button
+              id={"download-button"}
               onClick={() => {
                 let consoleExtras = {};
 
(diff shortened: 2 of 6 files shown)


dvaldivia (approved):

kaankabalak (approved):

portal-ui/src/icons/DocumentationIcon.tsx
```diff
@@ -21,6 +21,7 @@ const DocumentationIcon = (props: SVGProps<SVGSVGElement>) => (
   <svg
     xmlns="http://www.w3.org/2000/svg"
     className={`min-icon`}
+    id={"documentation-icon"}
```
bexsoft, line 24: Not sure if we should ass ids directly to the icons as you can have more than one icon of the same type on a single page. (Eg. buckets). I recommend to add ids only to buttons and clickable elements

bexsoft (changes requested):

portal-ui/src/icons/DocumentationIcon.tsx
dvaldivia: that's true, perhaps we should add them at button level? @cniackz

cniackz: Yeah, make sense, we can do that and look for other `CSS` selector when needed.
Actually, I just found `xpath` is also supported, so at the end we can become creative in terms of selectors.
Let me update this PR to do so.
Thank you guys for the review!

portal-ui/src/screens/Console/Common/GenericWizard/WizardPage.tsx
```diff
@@ -105,6 +105,7 @@ const WizardPage = ({
           {page.buttons.map((btn) => {
             return (
               <Button
+                id={"wizard-button"+btn.label}
```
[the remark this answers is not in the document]
cniackz: That's a fair point, I did not noticed that I did that, it is good enough with this line below:
`id={"wizard-button"+btn.label}`

kaankabalak (approved):

adfost (approved):

jinapurapu (approved):

bexsoft (approved):
LGTM

PR #1613, opened 2022-02-22
Get Operator files for testing

cniackz:
Fixes https://github.com/miniohq/engineering/issues/536

The idea is to get `YAML` files from `Operator` repository in order to test Operator UI without having duplicated files in both repositories. I will be using GitHub Actions to checkout Operator repo during Console tests.

Found a nested option in: https://github.com/actions/checkout
https://github.com/actions/checkout#checkout-multiple-repos-nested
```sh
- name: Checkout
  uses: actions/checkout@v2

- name: Checkout tools repo
  uses: actions/checkout@v2
  with:
    repository: my-org/my-tools
    path: my-tools
```

Which is pretty useful to get `operator` files nested in our console repo.


cniackz:
I think @adfost was right on previous PRs. We had many files and the way to reduce them is to get them from operator repository I think... Let's wait for the review process and see.

bexsoft (approved):

PR #1618, opened 2022-02-23
[WIP] - Execute parallel tests only if basics has passed and use cache to save execution time

cniackz:
Fixes https://github.com/miniohq/engineering/issues/539

We need to adopt common stages **to speed up our tests**.


cniackz:
I have a new idea in mind, since steps can't be parallel, at least we can improve this a bit by running all the tests only if basic ones has passed
Screen Shot 2022-02-25 at 7 27 00 AM

cniackz:
Screen Shot 2022-02-25 at 1 25 55 PM

Great `operator` test is now re-using the dependencies saved on cache from `yarn` this is why we can see the skip step over there, saving some time to get the assets.

cniackz:
Screen Shot 2022-02-25 at 1 28 01 PM

This is how it looks the diagram, `Operator Tests` is now re-using the files cached by previous Job. This is amazing Daniel, a bit time consuming to test and see the result, but it is totally worth it

cniackz:
will handle in https://github.com/minio/console/pull/1630

PR #1620, opened 2022-02-24
UX error alert banner

prakashsvmx:
UX error alert banner to be consistent with SUBNET

Impacted Screens:
- Login Page
- All screens with Error banner
- Modal with server side error

Closes https://github.com/miniohq/engineering/issues/519
Closes https://github.com/miniohq/engineering/issues/548

Hoe Does it look:
#### Login
Loginpage
#### Mobile
Mobile

Error

#### Common
Default

#### Modal
Modal


oscarocastellanos (dismissed):
LGTM

portal-ui/src/icons/AlertCloseIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    className={`min-icon`}
+    fill={"currentcolor"}
+    width="11"
+    height="11"
+    {...props}
+  >
+    <g
+      stroke="#FFF"
```
dvaldivia, line 30: we should remove the hardcoded colors so this icon can be styled from external CSS. Also add it to `IconsScreen.tsx` and visit to `/icons` to see if it can be styled like the rest of the icons

dvaldivia: Can we also get the icon with a viewport="0 0  256 256" @oscarocastellanos ? I dont' mind if we update this icon on a separate PR

dvaldivia (approved):

portal-ui/src/icons/AlertCloseIcon.tsx
prakashsvmx: I used the icon from SUBNET. I will update it. Thank you for the suggestion.

bexsoft (changes requested):
In some cases we may have two errors coming from backend (A generic error response and an advanced error). I think it would be better to add some logic to show only one of another, because I think error may get confusing. What do you think @oscarocastellanos ?

Screen Shot 2022-02-24 at 11 41 44

portal-ui/src/screens/LoginPage/LoginPage.tsx
```diff
@@ -552,8 +553,9 @@ const Login = ({
 
   return (
     <React.Fragment>
+      <CssBaseline />
```
bexsoft, line 556: Are we using this component?

bexsoft (changes requested):

portal-ui/src/icons/AlertCloseIcon.tsx
oscarocastellanos: > In some cases we may have two errors coming from backend (A generic error response and an advanced error). I think it would be better to add some logic to show only one of another, because I think error may get confusing. What do you think @oscarocastellanos ?
> 
> Screen Shot 2022-02-24 at 11 41 44



> In some cases we may have two errors coming from backend (A generic error response and an advanced error). I think it would be better to add some logic to show only one of another, because I think error may get confusing. What do you think @oscarocastellanos ?
> 
> Screen Shot 2022-02-24 at 11 41 44

I do agree @bexsoft. I would suggest that the more advanced error overrides the generic error.

oscarocastellanos: > Can we also get the icon with a viewport="0 0 256 256" @oscarocastellanos ? I dont' mind if we update this icon on a separate PR

PR is opened here: https://github.com/miniohq/engineering/issues/548

portal-ui/src/screens/LoginPage/LoginPage.tsx
prakashsvmx: Yes this is required for the css reset of the entire page . post login it is used in Console.tsx as well.

cniackz:
Two things:
1. Is error below related to these changes?
` ✖ Start New Diagnostic button exists after Diagnostic is completed` <-- Will re-run the workflow to see if it can be reproduced again.
2. Let's resolve the conflict to proceed.

bexsoft (approved):
LGTM

dvaldivia (approved):

PR #1621, opened 2022-02-24
Fix Hop hiding menu

dvaldivia:
Introduces logic to split the Hop screen and the embedded console screens like metrics, this is achieved by adding a prefix to the cookie being used to store the sub-session, also a unique identifier per operator session was added to avoid stale sessions (which is a known bug)


cniackz (approved):
LGTM

bexsoft (approved):
LGTM, Tested along with Daniel

PR #1638, opened 2022-03-01
Fix Kbar in operator mode

dvaldivia:
Screen Shot 2022-02-28 at 10 01 47 PM


prakashsvmx:
Observation:
do we apply permissions while listing the actions for users with policy?
e.g : User with policy which denies admin actions. (users, groups etc... are not expected as they would redirect to bucket page when clicked because of policy)

image

dvaldivia:
@prakashsvmx aaah, we need to deal with this problem, right now we load the kbar option once and don't reload after the session is loaded (when we get the right access policy) so we would need to introduce something to re-evaluate the kbar menu after we get the access policy to avoid listing these

prakashsvmx (approved):
Tested. Changes look good to me. We can improve progressively

Alevsk (approved):

bexsoft (approved):
LGTM

PR #1646, opened 2022-03-01
Fixed delete object behavior

bexsoft:
## What does this do?
Fixed delete object behavior, also fixed an issue with multi-select & panel sizes


dvaldivia (approved):

adfost:
> Should the "Delete all versions" switch be disabled if versioning is not enabled on the bucket?

It should, that's how I set it up.

Alevsk (approved):

PR #1647, opened 2022-03-01
Improving coverage limit check

cniackz:
Fixes https://github.com/miniohq/engineering/issues/559 

To improve `coverage limit check` by re-using the structure already in place in `integration` test, so that `limit check` is more light and re-use the cache to get the result that can be obtained already in previous step.

Also added the coverage from `pkg` since `go test` is being used as well in that spot.

# Result
```
Run echo "change directory to gocovmerge"
change directory to gocovmerge
download golang x tools
go mod tidy compat mode
go build gocoverage.go
put together the outs for final coverage resolution
grep to obtain the result
result:
47.9
greater than threshold, passed!
```

### DIFF
--- a/.github/workflows/jobs.yaml
+++ b/.github/workflows/jobs.yaml
@@ -113,7 +113,6 @@ jobs:
     needs:
       - lint-job
       - no-warnings-and-make-assets
-      - coverage
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -141,7 +140,6 @@ jobs:
     needs:
       - lint-job
       - no-warnings-and-make-assets
-      - coverage
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


cniackz:
Discussed with @dvaldivia on this PR, and it can be further improved, there are other jobs that can help us improve the coverage indicator and those will be included, so for now work in progress here...

dvaldivia:
I think this is great, if you add `make test-pkg` in the `test-pkg-on-go` job to the coverage we may break 47.2% coverage

dvaldivia (approved):

bexsoft (approved):
LGTM

cniackz:
Hey @dvaldivia actually I did, `test-pkg-on-go` was indeed added and we save `coverage/coverage-pkg.out` file in:
```
      - uses: actions/cache@v2
        id: coverage-cache-pkg
        name: Coverage Cache Pkg
        with:
          path: |
            ./pkg/coverage/
          key: ${{ runner.os }}-coverage-pkg-2-${{ github.run_id }}
```
So current coverage value includes that job.

PR #1651, opened 2022-03-02
Whitelist for preview files from the backend

Alevsk:
This PR adds a whitelist of safe files to download with
`Content-Disposition: inline;` from the backend, all other files will be
force download via `Content-Disposition: attachment;` existing svg files
will still be rendered in a secure way via the html `image` tag.

reference: https://digi.ninja/blog/svg_xss.php


dvaldivia (approved):

bexsoft (approved):
LGTM

PR #1655, opened 2022-03-03
Update all direct deps for console

harshavardhana:


dvaldivia:
it only failed to compile for `darwin/amd64`

cniackz:
I am reviewing this PR as requested by @harshavardhana 
To be quite honest, I am not sure how to solve this builing issue; but I've been reading on `CGO` and it seems we have `C` dependencies over `go-ieproxy` project, which I am still not convinced this is the problem. But I have seen that for this particular cross compilation, we can build if the CGO flag is enabled, so I added a new commit on top of this branch to test that possibility then I will consult with Harsha or Daniel as they might know more than me on this topic.

cniackz:
Sadly this option added will only work in my laptop
```go

cniackz:
Ok for now I have created an issue in:
https://github.com/mattn/go-ieproxy/issues/32

My next step would be to either keep the old version or skip the cross compilation until this is solved.
I will decide on that in a bit.

harshavardhana:
We shouldn't do these things @cniackz we want to compile without CGO - we have zero interest in using CGO.

cniackz (dismissed):
If we can skip cross compilation for `darwin/amd64` as it is not currently working; then I approve this change and in the meantime, I will be waiting on the response from:
mattn/go-ieproxy#32

cniackz:
Interesting, trying

harshavardhana:
I moved the dependency to v0.0.2 @cniackz

cniackz:
Yes I see that but we still get the same
Screen Shot 2022-03-08 at 1 28 25 PM
 problem:

harshavardhana:
> If we can skip cross compilation for `darwin/amd64` as it is not currently working; then I approve this change and in the meantime, I will be waiting on the response from: [mattn/go-ieproxy#32](https://github.com/mattn/go-ieproxy/issues/32)

- No that's fixed already - it is important that we can compile otherwise we won't be able to release darwin binaries. 
- Reverted the ie-proxy for v0.0.1

cniackz:
Ok, staying with `github.com/mattn/go-ieproxy v0.0.1 // indirect` then.
Thanks! I think that will work

cniackz (approved):
`github.com/mattn/go-ieproxy v0.0.1 // indirect` <---- Approved, this will work!

Thanks Harsha :)

Alevsk (approved):

PR #1664, opened 2022-03-04
[WIP] Call home API

bexsoft:
## What does this do?

Implements Call Home API in console


restapi/user_support.go
```diff
@@ -0,0 +1,206 @@
...
+
+		return user_api.NewGetCallHomeSetOptionOK().WithPayload(response)
+	})
+}
+
+// getCallHomeOption returns the selected option value
+func getCallHomeOption(session *models.Principal, params user_api.GetCallHomeOptionValueParams) (*models.CallHomeGetResponse, *models.Error) {
+	mAdmin, err := NewMinioAdminClient(session)
+	if err != nil {
+		return nil, prepareError(err)
```
Alevsk, line 63: Please use ErrorWithContext() function and pass the context from params.HTTP.Context() so we log all the error when audit log is enabled

Alevsk (changes requested):

restapi/user_support.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	// 20 seconds timeout
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
```
Alevsk, line 67: Please use params.HTTP.Context()

Alevsk (changes requested):

PR #1669, opened 2022-03-04
Add Set Bucket Versioning Test

cniackz:


dvaldivia (approved):

integration/user_api_bucket_test.go
```diff
@@ -2975,3 +2975,86 @@ func GetBucketVersioning(bucketName string) (*http.Response, error) {
...
+	if response != nil {
+		assert.Equal(201, response.StatusCode, inspectHTTPResponse(response))
+	}
+
+	// 2. Set versioning as False
+	response, err = SetBucketVersioning(bucket, false)
+	assert.Nil(err)
+	if err != nil {
+		log.Println(err)
+		assert.Fail("Error creating the bucket")
```
kaankabalak, line 3034: Hi @cniackz, should we change the error message here as we get this when we're setting the bucket versioning instead of creating the bucket?

cniackz: Yes, this has to be corrected. Thank you @kaankabalak

jinapurapu (approved):

adfost (approved):

PR #1683, opened 2022-03-08
Updated tenant creation credentials download to allow mc alias import

jinapurapu:
**What does this do:**

Changes tenant creation credentials to remove underscores, now in the format

    "accessKey": "15JVL8xGfQa8AZqh",
      "secretKey": "Flk1QUrqZItYRRlDQWE8B4TOoprjnSDo",


portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
```diff
@@ -197,8 +197,8 @@ const CredentialsPrompt = ({
                       console: [
                         {
                           url: consoleCreds.url,
-                          access_key: consoleCreds.accessKey,
-                          secret_key: consoleCreds.secretKey,
+                          accesskey: consoleCreds.accessKey,
+                          secretkey: consoleCreds.secretKey,
                           api: "s3v4",
                           path: "auto",
                         },
```
harshavardhana, line 204: this doesn't work @jinapurapu - this returns the values in this format

```
{
    console: [
    ]
}
```

this is not parsable by `mc`

harshavardhana (dismissed):

portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
```diff
@@ -194,30 +194,23 @@ const CredentialsPrompt = ({
                 if (consoleCreds) {
                   if (!Array.isArray(consoleCreds)) {
                     consoleExtras = {
-                      console: [
-                        {
                           url: consoleCreds.url,
-                          access_key: consoleCreds.accessKey,
-                          secret_key: consoleCreds.secretKey,
+                          accesskey: consoleCreds.accessKey,
+                          secretkey: consoleCreds.secretKey,
```
bexsoft, line 199: I think this should be in camelCase

```suggestion
                          accessKey: consoleCreds.accessKey,
                          secretKey: consoleCreds.secretKey,
```

portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
```diff
@@ -194,30 +194,23 @@ const CredentialsPrompt = ({
...
-                      ],
                     };
                   } else {
                     const cCreds = consoleCreds.map((itemMap) => {
                       return {
                         url: itemMap.url,
-                        access_key: itemMap.accessKey,
-                        secret_key: itemMap.secretKey,
+                        accesskey: itemMap.accessKey,
+                        secretkey: itemMap.secretKey,
```
bexsoft, line 208: Same Case:

```suggestion
                        accessKey: itemMap.accessKey,
                        secretKey: itemMap.secretKey,
```

bexsoft (changes requested):

prakashsvmx:
Observation:
When we cutomize the accesskey and secretKey and download,
It could not be imported. as the url is not present in the json  @jinapurapu 

- Create  New Service Account -> Customize Credentials ->   `test-svc-acct` and `test-svc-acct`  -> Create -> Download.

`mc alias import sa-creds ./sa_credentials.json`
```
mc: <ERROR> Invalid URL. URL `undefined` for MinIO Client should be of the form scheme://host[:port]/ without resource component.
```


`cat sa_credentials.json `

```json
{"url":"undefined","accessKey":"test-svc-acct","secretKey":"test-svc-acct","api":"s3v4","path":"auto"}
```

-  Similarly Customize + Restrict with Policy and download and observe that the url is undefined.

- [x] the default Credentials are successfully imported.

dvaldivia (approved):

portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
```diff
@@ -194,30 +194,23 @@ const CredentialsPrompt = ({
...
+                        secretKey: itemMap.secretKey,
                         api: "s3v4",
                         path: "auto",
                       };
                     });
-
-                    consoleExtras = {
-                      console: [...cCreds],
-                    };
+                    consoleExtras = cCreds[0];
```
bexsoft, line 213: Map can return more than one element in some cases (Eg. tenant creation with multiple users)

bexsoft (changes requested):

portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
harshavardhana: Generating multiple credentials files is actually useless @bexsoft it's not easily readable and also copy-paste friendly.

Anyone making an S3 API call only ever uses a single credential we should provide an option to download the first credential - as well as have a mechanism to download individual credentials separately. 

They should not be piled together in a single JSON file since those are quite meaningless. 

This JSON file should be treated like a `gcs-credentials.json` file that can be directly used with gcloud. In our case its `mc`. 

So currently without changing much of the UI - potentially choosing to show that you are downloading a single credential is enough since that itself is broken today. 

However potential for a UI change is also now visible because we do create multiple credentials during tenant provisioning but have no way to download them individually.

prakashsvmx:
@jinapurapu 
is the operator tenant creation credential download also updated to be `mc` import compatible ?   
otherwise changes look good to me :+1: 

I tried created a new tenant and downloaded the credentials. and it looks like the following

```
{"console":[{"url":"http://minio.new-tenant.svc.cluster.local:80","access_key":"t8yNwEmjW6iv7ZIS","secret_key":"VlyZRi2MIB1nwyP5bpJoprNmxoJLYWrf","api":"s3v4","path":"auto"}]}
```

prakashsvmx (approved):
Tested. changes look good to me. :+1: 

- Service Accounts,
- User Service accounts
- Operator Tenant Credentials.

portal-ui/src/screens/Console/Common/CredentialsPrompt/CredentialsPrompt.tsx
jinapurapu: I will work on a UI to download the entire credentials list in another PR

bexsoft: @harshavardhana I mention this because this component is used in multiple modules to handle different scenarios. 

@jinapurapu Ok, I think it would be better to make independent components to show credentials. This can be handled in another PR as you mention. Will approve this PR

bexsoft (approved):
A new component will be created to handle the multiple credentials scenario and replace it in create tenant screen.

PR #1689, opened 2022-03-09
Fix whitespace appearing on Login scroll after height decrease

### LINKED ISSUE #1688: White space is left when viewport height is decreased and user scrolls down on Login screen
As stated in the title of the issue, white space is left when the user decreases the viewport height and scrolls down on the Login screen. Please refer to the image below for further context:

Screen Shot 2022-03-09 at 12 43 27 PM

We can fix this by adding the CSS rule `overflow: auto` to the root class of the login field.

kaankabalak:
Fixes #1688


bexsoft (approved):
LGTM, Works fine

dvaldivia (approved):

PR #1690, opened 2022-03-10
Replaced CircularProgress component with new Loader

bexsoft:
## What does this do?

Replaced all CircularProgress implementations with the new Loader


portal-ui/src/screens/Console/HealthInfo/HealthInfo.tsx
```diff
@@ -266,7 +266,7 @@ const HealthInfo = ({
                   <div className={classes.localMessage}>{localMessage}</div>
                   {serverDiagnosticStatus === DiagStatInProgress ? (
                     <div className={classes.loading}>
-                      <CircularProgress size={25} />
+                      <Loader style={{ width: 25, height: 25 }} />
```
prakashsvmx, line 269: @bexsoft  any specific reason different screens have different height /width for the loader? 
we can have a size prop like size : small, medium , large if required so we an use it based the requirements.
@oscarocastellanos  any thoughts?

bexsoft: @prakashsvmx Yes, these sizes differ where the component is placed (In this case is for diagnostics page), some other sizes apply for tables & other non common components. 

@oscarocastellanos What do you think?

oscarocastellanos: @prakashsvmx I agree with @bexsoft, I think it was a good idea to adapt the loader for different screen sizes and uses. Mostly, as a per-component basis, I think works fine.

bexsoft:
> Will we have the same circular progress in index.html ?
> 
> otherwise changes look good to me. 👍

Yes, for that one will  implement the svg file, I will do it in a new PR

dvaldivia (approved):

PR #1704, opened 2022-03-11
Download all credentials

jinapurapu:
**REQUIRES PO 1683**

**What does it do:** If more than one set of accesskey/secretkey pairs is generated, provide a button to allow download of all accesskey/secretkey pairs. Adds tooltips on hover to explain difference between download formats. Also removes "Done" button to minimize accidental user closure of credential window.

**What does it look like:**
Screenshot from 2022-03-11 11-09-02
Screenshot from 2022-03-11 11-08-33


dvaldivia (approved):

bexsoft (approved):

dvaldivia (changes requested):
the UI tests on operator are failing due to the `Done` button being removed, perhaps the test should be updated to close the modal using the top right

dvaldivia (approved):

bexsoft (approved):
LGTM

PR #1710, opened 2022-03-14
Removing delete all versions option from non versioned bucket

adfost:
Removes option to delete all versions from non versioned bucket.


kaankabalak (approved):
LGTM

PR #1731, opened 2022-03-17
Enabled Delete selected version functionality

bexsoft:
## What does this do?
Enabled Delete selected version functionality

## How does it look?

Screen Shot 2022-03-16 at 19 13 13
Screen Shot 2022-03-16 at 19 13 01
Screen Shot 2022-03-16 at 19 12 53

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/DeleteObject.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/DeleteObject.tsx
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import React, { useState } from "react";
+import React, { useState, Fragment } from "react";
 import { connect } from "react-redux";
 import { DialogContentText } from "@mui/material";
 import { setErrorSnackMessage } from "../../../../../../actions";
@@ -32,6 +32,7 @@ interface IDeleteObjectProps {
   selectedBucket: string;
   setErrorSnackMessage: typeof setErrorSnackMessage;
   versioning: boolean;
+  selectedVersion?: string;
 }
 
 const DeleteObject = ({
@@ -41,6 +42,7 @@ const DeleteObject = ({
   selectedObject,
   setErrorSnackMessage,
   versioning,
+  selectedVersion = "",
 }: IDeleteObjectProps) => {
   const onDelSuccess = () => closeDeleteModalAndRefresh(true);
   const onDelError = (err: ErrorResponseHandler) => setErrorSnackMessage(err);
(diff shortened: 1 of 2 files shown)


Alevsk (dismissed):

prakashsvmx:
@oscarocastellanos 
The actions look ambiguous ( Preview, Download, Share ) when viewing object versions.
<summary> Delete marker Delete option  <summary>

<details> 


image

</details>

@bexsoft 
Delete marker version cannot be deleted from versions panel but can be deleted from main list using Delete all versions toggle in the modal

<summary> Delete marker Delete option <summary>

<details>

image

</details>

prakashsvmx (dismissed):
LGTM.

oscarocastellanos:
> @oscarocastellanos the actions are common for every item just ( enabled/disabled) so we can move the restore on every item to the right panel . and remove all the icons on each item. ?

You're right @prakashsvmx , they're redundant. Maybe we can rid of all those icons, but then we might start thinking on implementing the same sidebar to most of the screens. Thoughts @bexsoft?

bexsoft:
> @oscarocastellanos The actions look ambiguous ( Preview, Download, Share ) when viewing object versions.
> 
> Version actions and side panel actions
> @bexsoft Delete marker version cannot be deleted from versions panel but can be deleted from main list using Delete all versions toggle in the modal
> 
> Delete marker Delete option

Enabled button when selected version has delete marker. button still disabled when latest version is delete_marker and no version has been selected

Screen Shot 2022-03-17 at 13 39 02

oscarocastellanos:
> > > @oscarocastellanos the actions are common for every item just ( enabled/disabled) so we can move the restore on every item to the right panel . and remove all the icons on each item. ?
> > 
> > 
> > You're right @prakashsvmx , they're redundant. Maybe we can rid of all those icons, but then we might start thinking on implementing the same sidebar to most of the screens. Thoughts @bexsoft?
> 
> @oscarocastellanos Yes, I think those actions are redundant there after clicking one specific version, on the other side this helps to avoid clicking multiple times to get an action done when you know what you want to do
> 
> I think that if we remove them, we will have some empty space there that will look odd in the design.

@prakashsvmx @bexsoft Ahh, we have come to an interesting conundrum. Consider: We're also adding size and date as a column, so the empty space would not be a problem in the end

Captura de Pantalla 2022-03-17 a la(s) 13 45 57

bexsoft:
> > > > @oscarocastellanos the actions are common for every item just ( enabled/disabled) so we can move the restore on every item to the right panel . and remove all the icons on each item. ?
> > > 
> > > 
> > > You're right @prakashsvmx , they're redundant. Maybe we can rid of all those icons, but then we might start thinking on implementing the same sidebar to most of the screens. Thoughts @bexsoft?
> > 
> > 
> > @oscarocastellanos Yes, I think those actions are redundant there after clicking one specific version, on the other side this helps to avoid clicking multiple times to get an action done when you know what you want to do
> > I think that if we remove them, we will have some empty space there that will look odd in the design.
> 
> @prakashsvmx @bexsoft Ahh, we have come to an interesting conundrum. Consider: We're also adding size and date as a column, so the empty space would not be a problem in the end
> 
> Captura de Pantalla 2022-03-17 a la(s) 13 45 57

@oscarocastellanos Yeah, We can work on that change in a new task

prakashsvmx (approved):
LGTM :+1:

dvaldivia (approved):

PR #1738, opened 2022-03-19
Enabled Delete button for sub resources

bexsoft:
## What does this do?

Enabled Delete button for sub resources. This is a temporal bypass for policies that contains Read-only permissions for buckets but allow sub elements to have all the S3 Permissions.

 In case the user doesn't have a real permission to delete an object, an access denied error will be shown.


kannappanr (approved):
LGTM

dvaldivia (approved):

PR #1742, opened 2022-03-21
SSO Integration Test

cniackz:
Fixes https://github.com/miniohq/engineering/issues/647

The intention of this test is to make sure we can log in using SSO via `Keycloak`

# How it works:

1. We have a container for `keycloak`
2. We have a container for configuring `keycloak` which is `keycloak-config-cli`
3. We have `minio` server container running in distributed mode
4. We have `mc` running into another container to configure and restart `sso` on `mino`

So 4 containers in total, in order to post to the end point below:
```go
	request, _ := http.NewRequest(
		"POST",
		"http://localhost:9001/api/v1/login/oauth2/auth",
		requestDataBody,
	)
	request.Header.Add("Content-Type", "application/json")
```

Expected is `204` status and to get the `token` meaning we were able to login via Single Sign On method.


Makefile
```diff
@@ -80,6 +80,57 @@ test-integration:
...
+    -v /data1 -v /data2 -v /data3 -v /data4 \
+    --network my-net \
+    -d \
+    --name minio \
+    --rm \
+    -p 9000:9000 \
+    -p 9001:9001 \
+    -e MINIO_IDENTITY_OPENID_CLIENT_SECRET=0nfJuqIt0iPnRIUJkvetve5l38C6gi9W \
+    -e MINIO_ROOT_USER=minio \
+    -e MINIO_ROOT_PASSWORD=minio123 "quay.io/minio/minio:latest" server /data{1...4} --address :9000 --console-address :9001)
```
dvaldivia, line 120: the downside here is that this uses `minio:latest` instead of the code in the current PR, but we have a container image for the current PR so we could use that image instead, or the built minio binary

dvaldivia (changes requested):

dvaldivia (approved):

bexsoft (approved):

adfost (approved):

PR #1743, opened 2022-03-22
Adding link to mangage bucket from list objects.

adfost:
Adds a link to manage bucket from List Objects.
Screen Shot 2022-03-21 at 4 58 41 PM


bexsoft (changes requested):
This button is already present in the objects listing page. Are we planning to remove it?

Screen Shot 2022-03-21 at 18 04 42

PR #1767, opened 2022-03-25
Fix authorization code flow handling

donatello:
- Remove checking for unnecessary response types - this has been causing a
regression since MinIO release RELEASE.2022-03-03T21-21-16Z


harshavardhana:
This is needed check the original fix here @donatello - it would regress on the original fix.

donatello:
> This is needed check the original fix here @donatello - it would regress on the original fix.

Yes, this change is making the check looser than it is currently in the release. But I am not able to find why the last fix was needed. This change is breaking even a simple Dex IDP. From my reading of the spec, authorization code flow needs only the `code` response type as `code` response type is always exchanged for `id_token` and `access_token`. (https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)

donatello:
PTAL @dvaldivia @Alevsk

Alevsk (approved):
Tested against auth0, works fine

dvaldivia (approved):

PR #1770, opened 2022-03-28
Move Edit Bucket Replication to page

adfost:
Moves edit bucket replication to page from modal. Screen Shot 2022-03-28 at 10 07 01 AM


oscarocastellanos (changes requested):
- Let's keep consistent all the external labels. Destination label is in a different color than the rest, we should homogenize this.
- Remove "Back to replication panel". The back button should appear only on the top of the page ( "⬅️  Replication" )
- Action button should be filled in blue, not outlined.

PR #1777, opened 2022-03-29
Fix redundant Prefix listed as object

dvaldivia:


bexsoft (approved):
LGTM, Works fine

bexsoft (changes requested):
Found one issue with object details page:
Screen Shot 2022-03-29 at 17 03 40

PR #1807, opened 2022-04-07
UI site replication

prakashsvmx:
UI for site replication

How does it look:

S1
S2
image

S4
image

Site Replication status would be in a new PR.

### DIFF
--- a/portal-ui/src/common/SecureComponent/permissions.ts
+++ b/portal-ui/src/common/SecureComponent/permissions.ts
@@ -161,6 +161,7 @@ export const IAM_PAGES = {
   TIERS: "/settings/tiers",
   TIERS_ADD: "/settings/tiers/add",
   TIERS_ADD_SERVICE: "/settings/tiers/add/:service",
+  SITE_REPLICATION: "/settings/site-replication",
 
   /* Operator */
   TENANTS: "/tenants",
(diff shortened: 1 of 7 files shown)


oscarocastellanos (approved):
LGTM

dvaldivia:
@oscarocastellanos my only concern is that the `Delete` an `Edit` action buttons are different from everywhere else in the UI, should we adopt those squared icon buttons instead of the flat one or the circle ones? (which we should also make consistent btw hahaha)
Screen Shot 2022-04-07 at 10 19 37 AM

Screen Shot 2022-04-07 at 10 20 00 AM

oscarocastellanos:
> @oscarocastellanos my only concern is that the `Delete` an `Edit` action buttons are different from everywhere else in the UI, should we adopt those squared icon buttons instead of the flat one or the circle ones? (which we should also make consistent btw hahaha) Screen Shot 2022-04-07 at 10 19 37 AM
> 
> Screen Shot 2022-04-07 at 10 20 00 AM

You're right @dvaldivia, this slipped past me. Yeah, we should stick to the circle ones and try to keep them consistent everywhere.

oscarocastellanos (changes requested):
Please, let's use the action buttons for delete as circular style, as we use on other screens:
Captura de Pantalla 2022-04-07 a la(s) 12 26 39

prakashsvmx:
@dvaldivia  , @oscarocastellanos  - I will continue to update the pages for consistency

dvaldivia (approved):

oscarocastellanos (approved):
LGTM

Issue #1818, opened 2022-04-09
Upload large files gives timeout error sometimes


moracabanas:
I am running minio on Truenas, `mc` client works fine I uploaded 500Gb of photos at my full speed with no issues.

Now I am testing Console and getting upload errors. 

## Uploads starts normally

![image](https://user-images.githubusercontent.com/4382333/162569438-7c5c7bcf-e4bf-4c5e-b76f-3537d6f49e23.png)

## Some large uploads finished show red and pending.

![image](https://user-images.githubusercontent.com/4382333/162569540-ac41781c-9073-41d7-89f6-194b845c799f.png)

## Then waiting a couple of minutes some of them become green and OK. But others gives 500 error

![image](https://user-images.githubusercontent.com/4382333/162569605-c944cb2c-8b70-456a-98fe-be09269aaaff.png)

ERROR:

```
message: "Put \"https://s3.example.com/test/%5BOne%20Pace%5D%5B1-7%5D%20Romance%20Dawn%20%5B1080p%5D/%5BOne%20Pace%5D%5B1%5D%20Romance%20Dawn%2001%20%5B1080p%5D%5BFB72C13F%5D.mkv\": net/http: timeout awaiting response headers"

```

harshavardhana:
Console uploads as a single stream, there should be no timeouts here AFAIK

harshavardhana:
Which version of MinIO is this @moracabanas ?

moracabanas:
I am using RELEASE.2022-04-01T03-41-39Z on docker-compose and nginx as reverse proxy.

I also tested the next recommended nginx settings without luck

```nginx
ignore_invalid_headers off;
client_max_body_size 0;
proxy_buffering off;
```

When I check mc admin trace <minio alias> I can't see the PUT upload request, sometimes it shows later when upload gets OK on the Console, but in all cases the file is there when It reached 100% and I refresh the page.

harshavardhana:
> message: "Put \"https://s3.example.com/test/%5BOne%20Pace%5D%5B1-7%5D%20Romance%20Dawn%20%5B1080p%5D/%5BOne%20Pace%5D%5B1%5D%20Romance%20Dawn%2001%20%5B1080p%5D%5BFB72C13F%5D.mkv\": net/http: timeout awaiting response headers"

Okay, this is because a single PUT operation takes a longer time than the set ResponseHeaderTimeout since you have a slow network. We need to simply remove this value or keep it high enough to cater to slow networks.

[...]

PR #1827, opened 2022-04-11
Remove ResponseHeaderTimeout from default http transport

dvaldivia:
Fixes https://github.com/minio/console/issues/1818

### DIFF
--- a/restapi/tls.go
+++ b/restapi/tls.go
@@ -40,6 +40,7 @@ func prepareSTSClientTransport(insecure bool) *http.Transport {
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 10 * time.Second,
+		DisableCompression:    true,
 		TLSClientConfig: &tls.Config{
 			// Can't use SSLv3 because of POODLE and BEAST
 			// Can't use TLSv1.0 because of POODLE and BEAST using CBC cipher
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

harshavardhana (approved):

restapi/tls.go
```diff
@@ -32,16 +32,14 @@ func prepareSTSClientTransport(insecure bool) *http.Transport {
...
 			KeepAlive: 15 * time.Second,
 		}).DialContext,
 		MaxIdleConns:          1024,
 		MaxIdleConnsPerHost:   1024,
-		ResponseHeaderTimeout: 60 * time.Second,
-		IdleConnTimeout:       60 * time.Second,
+		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
-		ExpectContinueTimeout: 1 * time.Second,
-		DisableCompression:    true,
```
harshavardhana, line 44: you cannot remove this btw @dvaldivia

harshavardhana (dismissed):

restapi/tls.go
harshavardhana:
```
                                        // Set this value so that the underlying transport round-tripper
                                        // doesn't try to auto decode the body of objects with
                                        // content-encoding set to `gzip`.
                                        //
                                        // Refer:
                                        //    https://golang.org/src/net/http/transport.go?h=roundTrip#L1843
                                        DisableCompression: true,
```

It is explained in `mc` why

harshavardhana: a GetObject() would try to decompress automatically that has "content-encoding" header set to gzip.

dvaldivia: done

dvaldivia:
Only failing due to coverage

Alevsk (approved):

bexsoft (approved):

Issue #1790, opened 2022-04-01
${jwt:preferred_username} does not work


Pythonov:
## NOTE
When i use this IAM Policy everything works great, i see only 2 buckets and can get inside of them
<img width="333" alt="works2" src="https://user-images.githubusercontent.com/61191665/161285171-0613c10c-3ccf-42d2-b9f2-5ae35c70f79c.png">

But if i change name with ${jwt:preferred_username} i can see 2 buckets but can't get inside of them
 <img width="412" alt="dsntwrk2" src="https://user-images.githubusercontent.com/61191665/161285540-a68af4f7-7403-4e32-bedd-68a9345f3cda.png">
<img width="774" alt="dsntwork" src="https://user-images.githubusercontent.com/61191665/161285569-978da2a8-b19e-4537-b7a6-bb44f7575774.png">

## Expected Behavior
I expect the same behavior in both ways shown above.

## Current Behavior
Using ${jwt:preferred_username} leads to unable to get inside of the bucket

## Possible Solution
???

## Steps to Reproduce (for bugs)

1. Use KeyCloak for auth
2. Create user with name that equals the part of bucket's name
3. Test both IAM policies

## Your Environment

minio version RELEASE.2022-04-01T03-41-39Z
CentOs7

Issue #1597, opened 2022-02-17
Wildcard policy not allowing bucket browse


riazarbi:
Hello, 

# Background

We have a policy that we attach to all new users, which allows them full control of any buckets that have their username in the bucket name.

This policy works fine for minio, and things work as expected via the `mc` cli or python clients.

# Issue

However, things do not work as expected via the console. 

Observed behavior is that the buckets are listed under the buckets pane, but when you click browse you are redirected back to the bucket list. 

Browsing works fine in the console for other roles.

We are using the latest version of console (0.14.7) but this issue has been present for the entire time we have used console. It did not occur in the deprecated web UI.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::${aws:username}*"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::${aws:username},*"
            ]
        }
    ]
}
```

dvaldivia:
I'm not sure we substitute the variables on the client side, we will look into this, seems like an easy fix to me

harshavardhana:
> "arn:aws:s3:::${aws:username},*"

This policy is wrong with a `,`

riazarbi:
@harshavardhana I have changed it to the following to test if it resolves the issue, but it does not.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::${aws:username}*"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::${aws:username}*/*"
            ]
        }
    ]
}
```

The key thing to note is that I can browse my user prefixed buckets with `mc` or minio python, but not in the console.

riazarbi:
Issue persists in `0.15.2`

jaggerwang:
Same issue!

When using `aws:username` condition key in policy, user `team-10` cannot browse files in bucket `team-10` .
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetBucketLocation",
                "s3:ListBucket",
                "s3:ListBucketMultipartUploads"
            ],
[...]

Issue #1287, opened 2021-12-03
Console "Upload" button gone if PutObject Resource isn't full wildcard


enra64:
Hey, I'm trying to only allow my users to upload files to a subfolder corresponding to their name in the `config` bucket, using a policy like in the following example. However, that hides the upload button, which is necessary to keep this easy for us, even though errors might occur with missing permissions.

# Prefix Policy Example
### Policy
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListAllMyBuckets",
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::*"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:DeleteObject",
                "s3:GetObject",
                "s3:PutObject"
            ],
            "Resource": [
                "arn:aws:s3:::study-configs/${aws:username}/*"
            ]
        }
    ]
}
```

### Resulting UI

![prefix-policy-no-upload-button](https://user-images.githubusercontent.com/7622426/144575602-d999f460-76d4-4421-a318-1ea19f968a3b.png)

# Wildcard Policy Example

However, this hides the upload button that is available if using a wildcard resource like in the following policy:

### Policy
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListAllMyBuckets",
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::*"
            ]
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:DeleteObject",
                "s3:GetObject",
                "s3:PutObject"
            ],
            "Resource": [
                "arn:aws:s3:::study-configs/*"
            ]
        }
    ]
}
```

### Resulting UI

![image](https://user-images.githubusercontent.com/7622426/144575355-a4648f9b-78ec-4702-bbc9-54b295514a0a.png)

Is this intended behaviour, or is there maybe a workaround using the existing policies that I haven't found? Best, Arne

harshavardhana:
which version of console is this? @enra64

enra64:
So this is running on a standalone minio (for testing) that I downloaded yesterday, version should be `2021-11-24T23:19:33Z`:

![image](https://user-images.githubusercontent.com/7622426/144579361-ca5cdb24-7cb4-41e2-bc37-bc9eb0a6af11.png)

[...]

PR #1828, opened 2022-04-11
Replace `aws:username`, `jwt:` and `ldap:` policy variables in session policies

dvaldivia:
Fixes https://github.com/minio/console/issues/1803 
Fixes https://github.com/minio/console/issues/1287
Fixes https://github.com/minio/console/issues/1790
Fixes https://github.com/minio/console/issues/1597

Add support to replace `${aws:username}` for the current user username so the secure components can allow access to resoures with this variable in them

### DIFF
--- a/restapi/user_login.go
+++ b/restapi/user_login.go
@@ -17,13 +17,12 @@
 package restapi
 
 import (
-	"bytes"
 	"context"
 	"net/http"
 
-	"github.com/minio/minio-go/v7/pkg/credentials"
+	"github.com/minio/madmin-go"
 
-	iampolicy "github.com/minio/pkg/iam/policy"
+	"github.com/minio/minio-go/v7/pkg/credentials"
 
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/runtime/middleware"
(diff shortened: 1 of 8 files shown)


restapi/user_login.go
```diff
@@ -88,15 +88,19 @@ func login(credentials ConsoleCredentialsI, sessionFeatures *auth.SessionFeature
...
 	if err != nil {
 		return nil, err
 	}
-	return iampolicy.ParseConfig(bytes.NewReader(accountInfo.Policy))
+	return &accountInfo, nil
+}
+
+// replacePolicyVariables replaces known variables from policies with known values
+func replacePolicyVariables(accountInfo *madmin.AccountInfo) json.RawMessage {
+	rawPolicy := bytes.ReplaceAll(accountInfo.Policy, []byte("${aws:username}"), []byte(accountInfo.AccountName))
```
harshavardhana, line 102: this is the wrong way to do this, you should extract the SessionToken - claims have all the keys needed to replace the values here.

harshavardhana (changes requested):

restapi/user_login.go
dvaldivia: that will come next, right now I went after a single variable, all the JWT related variables do need the STS extracted

harshavardhana: it should come in this PR alone, its a generic fix @dvaldivia you can't just do `aws:username` as one off, there is also `aws:userid` 

You need support to replace

```
aws:
jwt:
ldap:
```

And this is a generic fix.

dvaldivia: We don't have a `userid` equivalent, should that be access key instead?

I added `jwt:` and `ldap:` to this PR @harshavardhana

Alevsk (approved):

bexsoft (approved):

PR #1845, opened 2022-04-14
Add site  in a new page for site replication

prakashsvmx:
Add Site in a New page.

How does it look:

image


oscarocastellanos (changes requested):
- Bullet points (blue squares) in help box doesn't match the alignment to their corresponding text. Please, make it match.
- When able to save the form, the save button should be blue background with white text.

dvaldivia:
@oscarocastellanos the save button behavior will work as expected

dvaldivia (approved):

oscarocastellanos (approved):
LGTM

PR #1848, opened 2022-04-14
Test for Get Object with Preview

dvaldivia:


integration/objects_test.go
```diff
@@ -0,0 +1,197 @@
...
+		"myobject.jpg", fileReader, int64(len(fakeFile)), minio.PutObjectOptions{ContentType: "application/octet-stream"})
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	assert := assert.New(t)
+	type args struct {
+		encodedPrefix string
+		versionId     string
```
kaankabalak, line 80: @dvaldivia, I think you might've already seen this, but there seems to be an error with the variable name here:

```suggestion
		versionID     string
```

integration/objects_test.go
```diff
@@ -0,0 +1,197 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			client := &http.Client{
+				Timeout: 3 * time.Second,
+			}
+			destination := fmt.Sprintf("/api/v1/buckets/%s/objects/download?preview=true&prefix=%s&version_id=%s", bucketName, tt.args.encodedPrefix, tt.args.versionId)
+			finalUrl := fmt.Sprintf("http://localhost:9090%s", destination)
```
kaankabalak, line 175: @dvaldivia, I think you might've already seen this too, but GitHub once again seems to be complaining with an issue with the variable name here:
```suggestion
			finalURL := fmt.Sprintf("http://localhost:9090%s", destination)
```

integration/objects_test.go
```diff
@@ -0,0 +1,197 @@
...
+			request, err := http.NewRequest("GET", finalUrl, nil)
```
kaankabalak, line 176: Similarly, if we are to replace `finalUrl` with `finalURL` as GitHub suggested, we would also need to do this:

```suggestion
			request, err := http.NewRequest("GET", finalURL, nil)
```

integration/objects_test.go
dvaldivia: that's right, I've updated the PR.

integration/objects_test.go
dvaldivia: that's right, I've updated the PR.

integration/objects_test.go
dvaldivia: that's right, I've updated the PR.

jinapurapu (dismissed):

cniackz (approved):
Looks good to me, added the increment:
```
-          threshold=35.60
+          threshold=36.5
```
Thank you Daniel for your help

cniackz:
And we are getting:
```
Error: buildx call failed with: error: failed to solve: unexpected status: 403 Forbidden
```
Reading: https://github.community/t/403-error-on-container-registry-push-from-github-action/173071/18

bexsoft (approved):

PR #1858, opened 2022-04-18
Disable Bucket replication tab when site replication is enabled in Bucket Summary page

prakashsvmx:
Disable Bucket replication tab when site replication is enabled in Bucket Summary page

How to Test:
When site replication is setup  , The Replication tab on the Bucket summary page is enabled.

Setup a Multi site replication `mc admin replicate add <site1_alias> <site2_alias>` (or use the Settings -> Site Replication UI)
Refresh, Verify that the Replication tab is disabled for a bucket

Remove site replication `mc admin replicate remove <site1_alias> <site2_alias>` 

Refresh , Verify that the Replication tab is enabled for a bucket

Closes https://github.com/miniohq/engineering/issues/713

### DIFF
--- a/portal-ui/src/reducer.ts
+++ b/portal-ui/src/reducer.ts
@@ -65,7 +65,6 @@ export function systemReducer(
   state = initialState,
   action: SystemActionTypes
 ): SystemState {
-  debugger;
   switch (action.type) {
     case USER_LOGGED:
       return {
--- a/portal-ui/src/reducer.ts
+++ b/portal-ui/src/reducer.ts
@@ -15,20 +15,21 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import {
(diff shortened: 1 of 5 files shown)


portal-ui/src/reducer.ts
```diff
@@ -63,6 +65,7 @@ export function systemReducer(
   state = initialState,
   action: SystemActionTypes
 ): SystemState {
+  debugger;
```
dvaldivia, line 68: I think this can be removed

dvaldivia (changes requested):

portal-ui/src/reducer.ts
prakashsvmx: Thank you :+1:  . I have updated.

dvaldivia (approved):

bexsoft (approved):

PR #1872, opened 2022-04-20
Fix the relative login redirects

dvaldivia:


prakashsvmx (approved):
LGTM.👍

portal-ui/src/screens/Console/Menu/Menu.tsx
```diff
@@ -106,8 +106,9 @@ const Menu = ({
       clearSession();
       userLoggedIn(false);
       localStorage.setItem("userLoggedIn", "");
+      localStorage.setItem("redirect-path", "");
```
harshavardhana, line 109: What is this used for ?

harshavardhana (changes requested):

portal-ui/src/screens/Console/Menu/Menu.tsx
dvaldivia: we use this localStorage item to redirect you back to a page if you need to login to go into an URL, but here I'm explicitly resetting it since this is the logout function

dvaldivia: @harshavardhana

PR #1879, opened 2022-04-20
More Policy Tests, Set, List, Delete

adfost:


integration/policy_test.go
```diff
@@ -33,10 +33,16 @@ import (
 func Test_PolicyAPI(t *testing.T) {
 	assert := assert.New(t)
 
+	AddUser("testuser1", "testtest", []string{}, []string{"readwrite"})
```
cniackz, line 36: Hello Adam,

These failures are expected. I helped with the first one: `service_account_test.go:74`
The reason is that in `integration/policy_test.go` a user is added `AddUser("testuser1", "testtest", []string{}, []string{"readwrite"})`
Hence by the time `TestAddServiceAccount` is executed, the user already exist so we are getting an expected 500:
```
{code: 500, detailedMessage: "Specified service account is used by another user",…}
code: 500
detailedMessage: "Specified service account is used by another user"
message: "an error occurred, please try again"
```
If the user is deleted then we can successfully post to `api/v1/service-account-credentials` to get other than 500.
Or you can use a different user to avoid the collision.
It is up to you how to handle this case, but make sure that any addition in any file is not breaking existing working scenarios :+1:

Best regards
-Cesar

cniackz, line 36: Hello Adam,

These failures are expected. I helped with the first one: `service_account_test.go:74`
The reason is that in `integration/policy_test.go` a user is added `AddUser("testuser1", "testtest", []string{}, []string{"readwrite"})`
Hence by the time `TestAddServiceAccount` is executed, the user already exist so we are getting an expected 500:
```
{code: 500, detailedMessage: "Specified service account is used by another user",…}
code: 500
detailedMessage: "Specified service account is used by another user"
message: "an error occurred, please try again"
```
If the user is deleted then we can successfully post to `api/v1/service-account-credentials` to get other than 500.
Or you can use a different user to avoid the collision.
It is up to you how to handle this case, but make sure that any addition in any file is not breaking existing working scenarios :+1:

Best regards
-Cesar

cniackz (changes requested):
Please avoid breaking existing scenarios. I helped to root cause the first failure observed in `service_account_test.go:74`.

Please start by fixing it and address the rest by debugging what is breaking the existing scenarios. Whatever is added in `policy_test.go` it may affect other tests as they are using the same running containers.

cniackz (dismissed):
LGTM

integration/policy_test.go
```diff
@@ -94,6 +102,84 @@ func Test_PolicyAPI(t *testing.T) {
 			expectedStatus: 500,
 			expectedError:  nil,
 		},
+		{
+			name: "Set Policy - Valid",
+			args: args{
+				api:        "/set-policy",
```
dvaldivia, line 108: since these test a different endpoint, please put them on their own test method

integration/policy_test.go
```diff
@@ -33,10 +33,16 @@ import (
...
+
 	type args struct {
-		api    string
-		name   string
-		policy *string
+		api        string
+		method     string
+		name       string
+		policy     *string
+		entityType string
```
dvaldivia, line 43: don't overload arguments, this test is meant to only test POST to `/policies`, use a new method for other endpoints that have their own handlers.

integration/policy_test.go
```diff
@@ -94,6 +102,84 @@ func Test_PolicyAPI(t *testing.T) {
...
+				api:    "/policy?name=test3",
+				method: "GET",
+			},
+			expectedStatus: 500,
+			expectedError:  nil,
+		},
+		{
+			name: "Get Policies - Valid",
+			args: args{
+				api:    "/policy?name=test",
```
dvaldivia, line 150: each api handler should end up with it's own test method

dvaldivia (changes requested):

cniackz (changes requested):
Please squash the commits to one only

cniackz (changes requested):
Squash commits

cniackz (dismissed):
LGTM

integration/policy_test.go
```diff
@@ -30,11 +31,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_PolicyAPI(t *testing.T) {
+func Test_AddPolicyAPI(t *testing.T) {
 	assert := assert.New(t)
 
 	type args struct {
 		api    string
+		method string
```
dvaldivia, line 39: method is redundant, all tests will be POST

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+
+func Test_SetPolicyAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	AddUser("policyuser1", "testtest", []string{}, []string{"readwrite"})
+	AddGroup("testgroup123", []string{})
+
+	type args struct {
+		api        string
+		method     string
```
dvaldivia, line 150: method is redundant, all tests will be PUT

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+func Test_SetPolicyMultipleAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	AddUser("policyuser2", "testtest", []string{}, []string{"readwrite"})
+	AddUser("policyuser3", "testtest", []string{}, []string{"readwrite"})
+	AddGroup("testgroup1234", []string{})
+
+	type args struct {
+		api    string
+		method string
```
dvaldivia, line 259: method is redundant, all tests will be PUT

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+		name           string
+		args           args
+		expectedStatus int
+		expectedError  error
+	}{
+		{
+			name: "List Policies",
+			args: args{
+				api:    "/policies",
+				method: "GET",
```
dvaldivia, line 358: method is redundant, all tests will be GET

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}
+
+}
+
+func Test_GetPolicyAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	type args struct {
+		api    string
+		method string
```
dvaldivia, line 404: method is redundant, all tests will be GET, other arguments such as the policy name should be arguments themselves, also this tests depends on the other previous tests setting up policies `test3` and `test` which means if we run both tests in parallel, one could fail due to the race condition, setup policies unique to this test before the test

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+			}
+			if response != nil {
+				assert.Equal(tt.expectedStatus, response.StatusCode, tt.name+" Failed")
+			}
+		})
+	}
+
+}
+
+func Test_ConfigAPI(t *testing.T) {
```
dvaldivia, line 466: This is a test for `config` on a file called `policy_test.go`, it should have it's own file `config_test.go`

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+	assert := assert.New(t)
+
+	type args struct {
+		api    string
+		method string
```
dvaldivia, line 471: method is redundant, all tests will be GET

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}
+
+}
+
+func Test_PolicyListUsersAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	type args struct {
+		api    string
+		method string
```
dvaldivia, line 529: method is redundant, all tests will be GET

integration/policy_test.go
```diff
@@ -106,15 +111,455 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			client := &http.Client{
+				Timeout: 3 * time.Second,
+			}
+
+			// Add policy
```
dvaldivia, line 555: This comments is incorrect

integration/policy_test.go
```diff
@@ -127,10 +572,209 @@ func Test_PolicyAPI(t *testing.T) {
...
 	}
 
 }
+
+func Test_PolicyListGroupsAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	type args struct {
+		api    string
+		method string
```
dvaldivia, line 590: method is redundant, all tests will be GET

integration/policy_test.go
```diff
@@ -127,10 +572,209 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}{
+		{
+			name: "List Users for Policy - Valid",
+			args: args{
+				api:    "/policies/test/groups",
+				method: "GET",
+			},
+			expectedStatus: 200,
+			expectedError:  nil,
+		},
```
dvaldivia, line 606: this tests doesn't cover invalid get

integration/policy_test.go
```diff
@@ -127,10 +572,209 @@ func Test_PolicyAPI(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		args           args
+		expectedStatus int
+		expectedError  error
+	}{
+		{
+			name: "List Users for Policy - Valid",
+			args: args{
+				api:    "/policies/test/groups",
```
dvaldivia, line 601: group name can be parametrized

integration/policy_test.go
```diff
@@ -127,10 +572,209 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			client := &http.Client{
+				Timeout: 3 * time.Second,
+			}
+
+			// Add policy
```
dvaldivia, line 616: This comments is inaccurate

integration/policy_test.go
```diff
@@ -127,10 +572,209 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}
+
+}
+
+func Test_DeletePolicyAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	type args struct {
+		api        string
+		method     string
```
dvaldivia, line 651: method is redundant, all tests will be DELETE

integration/policy_test.go
```diff
@@ -127,10 +572,209 @@ func Test_PolicyAPI(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			client := &http.Client{
+				Timeout: 3 * time.Second,
+			}
+
+			// Add policy
```
dvaldivia, line 749: This comments is inaccurate

dvaldivia (changes requested):

integration/policy_test.go
adfost: There needs to be some cleanup before delete.

integration/policy_test.go
```diff
@@ -127,10 +127,579 @@ func Test_PolicyAPI(t *testing.T) {
...
+			args: args{
+				api: "/policy?name=test3",
+			},
+			expectedStatus: 500,
+			expectedError:  nil,
+		},
+		{
+			name: "Get Policies - Valid",
+			args: args{
+				api: "/policy?name=test",
```
dvaldivia, line 400: this test relies on a policy created by another test which will cause a race condition, create a policy with a name unique for this test

dvaldivia (changes requested):

dvaldivia (approved):

integration/policy_test.go
```diff
@@ -30,7 +31,64 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_PolicyAPI(t *testing.T) {
+func AddPolicy(name string, definition string) (*http.Response, error) {
+	/*
+		This is an atomic function to add user and can be reused across
```
prakashsvmx, line 36: Comments could be updated to reflect "Policy Tests".

prakashsvmx (approved):
LGTM :+1:

bexsoft (approved):

PR #1880, opened 2022-04-21
Fixed breadcrumbs container overflow in object browser

bexsoft:
## What does this do?

Fixed breadcrumbs container overflow in object browser for screen resolutions greater than 1062px. 

An upcoming PR will add special handlers for mobile devices

## How does it look?

Screen Shot 2022-04-20 at 20 30 41

### DIFF
--- a/portal-ui/src/index.tsx
+++ b/portal-ui/src/index.tsx
@@ -102,7 +102,7 @@ const GlobalCss = withStyles({
     ul: {
       paddingLeft:20,
       listStyle: "none", /* Remove default bullets */
-      "& li::before": {
+      "& li::before:not(.Mui*)": {
         content: '"￭"',
         color: "#2781B0",
         fontSize: 20,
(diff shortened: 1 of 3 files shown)


prakashsvmx (approved):
Tested
- Firefox
- Chrome
-Changes look good to me :+1: 

Suggestion:
In BrowserBreadcrumbs.js

instead of 
```
 <span> / </span>
```
We could change it to :

```
 <Box component={"span"} sx={{ marginLeft: "2px", marginRight: "2px" }}>
          /
        </Box>
```

so that copying the path does not add extra spaces between prefixes. but can be done in a new PR.

dvaldivia (approved):

dvaldivia:
that is a good suggestion @prakashsvmx, should we also add a button to copy the path? cc @oscarocastellanos

oscarocastellanos:
Submited PR for this here: 

https://github.com/miniohq/engineering/issues/811

PR #1883, opened 2022-04-21
Fix test 2 by uploading file as stream from string

dvaldivia:


.github/workflows/jobs.yaml
```diff
@@ -493,7 +493,7 @@ jobs:
       - name: Run TestCafe Tests
         uses: DevExpress/testcafe-action@latest
         with:
-          args: '"chrome:headless" portal-ui/tests/permissions-2/ --skip-js-errors -c 3'
+          args: '"chrome:headless" portal-ui/tests/permissions-2/ --skip-js-errors -c 1'
```
kaankabalak, line 496: @dvaldivia, is there any specific reason why we changed this from 3 to 1? Also, if we would like to keep this as 1, can't we just get rid of the concurrency argument?

dvaldivia: just testing for now, will remove

cniackz (changes requested):
Hey Daniel, this was already fixed in https://github.com/minio/console/pull/1884
`uploadObjectToBucket` is working fine if we run from proper folder: `testcafe chrome  portal-ui/tests/permissions-2/ --skip-js-errors`, please close this PR

PR #1893, opened 2022-04-22
[WIP] - Add end point to get csr from tenant

cniackz:
Fixes: https://github.com/miniohq/engineering/issues/742


swagger-operator.yml
```diff
@@ -1550,6 +1575,28 @@ definitions:
       password:
         type: string
 
+  csrElement:
```
dvaldivia, line 1578: Can you add the status for object?

operatorapi/volumes.go
```diff
@@ -218,3 +228,36 @@ func getPVCEventsResponse(session *models.Principal, params operator_api.GetPVCE
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	clientset, err := cluster.K8sClient(session.STSSessionToken)
+	if err != nil {
+		return nil, errors.ErrorWithContext(ctx, err)
+	}
+	csrName := params.Tenant + "-" + params.Namespace + "-csr"
+	csrResult, csrError := clientset.CertificatesV1().CertificateSigningRequests().Get(ctx, csrName, metav1.GetOptions{})
+	if csrError != nil {
+		panic(csrError.Error())
```
dvaldivia, line 242: let's not use panics, just print the error and return the API with an error, better to return a model error here

dvaldivia (changes requested):

operatorapi/volumes.go
```diff
@@ -218,3 +228,36 @@ func getPVCEventsResponse(session *models.Principal, params operator_api.GetPVCE
 	})
 	return retval, nil
 }
+
+func getTenantCSResponse(session *models.Principal, params operator_api.ListTenantCertificateSigningRequestParams) (*models.CsrElement, *models.Error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
Alevsk, line 233:
```suggestion
	ctx, cancel := context.WithCancel(params.HTTPRequest.Context())
```

For now on, every time we need to use context on a Response function lets grab it from the HTTPRequest object in params, in contains important context information helpful for logging errors

operatorapi/volumes.go
Alevsk, line 242:
```suggestion
		return nil, errors.ErrorWithContext(ctx, err)
```

Alevsk (dismissed):

swagger-operator.yml
cniackz: Yes @dvaldivia I am working to get that done, will commit in a bit.

operatorapi/volumes.go
cniackz: Ok @dvaldivia make sense, thank you!

operatorapi/volumes.go
cniackz: Got it, didn't know @Alevsk thank you for the review 👍

operatorapi/volumes.go
cniackz: Thanks, same comment as Daniel, resolved already, thanks for the solution 👍

Alevsk (approved):

dvaldivia (approved):

harshavardhana (approved):

PR #1900, opened 2022-04-25
Fix Subpath parsing

dvaldivia:

### DIFF
--- a/restapi/configure_console.go
+++ b/restapi/configure_console.go
@@ -53,6 +53,10 @@ var additionalServerFlags = struct {
 	CertsDir string `long:"certs-dir" description:"path to certs directory" env:"CONSOLE_CERTS_DIR"`
 }{}
 
+const (
+	SubPath = "CONSOLE_SUBPATH"
+)
+
 var subPath = "/"
 var subPathOnce sync.Once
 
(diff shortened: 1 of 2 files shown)


restapi/configure_console.go
```diff
@@ -363,18 +367,29 @@ func configureServer(s *http.Server, _, _ string) {
...
+	// also add pro-actively at the end.
+	subPath = filepath.Clean(filepath.ToSlash(v))
+	if !strings.HasPrefix(subPath, SlashSeparator) {
+		subPath = SlashSeparator + subPath
+	}
+	if !strings.HasSuffix(subPath, SlashSeparator) {
+		subPath = subPath + SlashSeparator
+	}
+	return subPath
+}
```
harshavardhana, line 391: Do not use `filepath` since that is OS-specific always use `path.Clean` and also do not blindly convert `\` to `/` since `\` is different and cannot be assumed to be `/` in browsers.

harshavardhana (changes requested):

restapi/configure_console.go
dvaldivia: filepath was the original xD I just moved it hahaha, I'll change to `path.Clean`

harshavardhana (approved):

Alevsk (approved):

bexsoft (approved):

PR #1912, opened 2022-04-27
Fixing integration tests to use kubectl proxy

cniackz:
Our `operator-integration` test are failing for new added end points, this is because we are missing the `kubectl proxy` to communicate to our cluster's API.


dvaldivia (approved):

adfost (approved):
Thanks for this PR, should fix my PR.

PR #1941, opened 2022-05-03
Test remote buckets WIP

adfost:
Test for adding remote bucket

This will test:

`~/console/swagger-console.yml`

```sh
  /remote-buckets:
    get:
      summary: List Remote Buckets
      operationId: ListRemoteBuckets
      responses:
        200:
          schema:
            $ref: "#/definitions/listRemoteBucketsResponse"
        default:
          schema:
            $ref: "#/definitions/error"
      tags:
        - Bucket
    post:
      summary: Add Remote Bucket
      operationId: AddRemoteBucket
      parameters:
        - name: body
          in: body
          required: true
          schema:
            $ref: "#/definitions/createRemoteBucket"
      responses:
        201:
        default:
          schema:
            $ref: "#/definitions/error"
      tags:
        - Bucket
```

### DIFF
--- a/integration/remote_buckets_test.go
+++ b/integration/remote_buckets_test.go
@@ -1,0 +1,329 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2022 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package integration
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/minio/console/models"
+	"github.com/minio/console/restapi"
+	"github.com/minio/madmin-go"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func AddRemoteBucket(accessKey string, secretKey string, targetURL string, sourceBucket string, targetBucket string) (string, error) {
+	ctx := context.Background()
+	mAdminClient, err := restapi.NewMinioAdminClient(&models.Principal{
+		STSAccessKeyID:     "minioadmin",
+		STSSecretAccessKey: "minioadmin",
+	})
+	u, err := url.Parse(targetURL)
+
+	host := u.Host
+	if u.Port() == "" {
+		port := 80
+		host = host + ":" + strconv.Itoa(port)
+	}
+	creds := &madmin.Credentials{AccessKey: accessKey, SecretKey: secretKey}
+	remoteBucket := &madmin.BucketTarget{
+		TargetBucket:    targetBucket,
+		Secure:          false,
+		Credentials:     creds,
+		Endpoint:        host,
+		Path:            "",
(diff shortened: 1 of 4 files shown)


integration/remote_buckets_test.go
```diff
@@ -0,0 +1,117 @@
...
+	tests := []struct {
+		name           string
+		args           args
+		expectedStatus int
+		expectedError  error
+	}{
+		{
+			name: "Create Remote Bucket - Valid",
+			args: args{
+				api:          "/remote-buckets",
```
dvaldivia, line 54: all these tests are testing the same API, please remove this argument

dvaldivia (changes requested):

integration/remote_buckets_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_AddRemoteBucketAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	AddBucket("remotebucketsource", false, true, map[string]interface{}{}, map[string]interface{}{})
+	AddBucketWithPort("remotebucketdest", false, true, map[string]interface{}{}, map[string]interface{}{}, "9001")
```
dvaldivia, line 35: no need to duplicate the code, just add some options to the `AddBucket` helper method to pass the port

dvaldivia (changes requested):

integration/user_api_bucket_test.go
```diff
@@ -69,6 +69,38 @@ func AddBucket(name string, locking bool, versioning bool, quota map[string]inte
 	return response, err
 }
 
+func AddBucketWithPort(name string, locking bool, versioning bool, quota map[string]interface{}, retention map[string]interface{}, port string) (*http.Response, error) {
```
dvaldivia, line 72: this is duplicate code, just update the existing AddBucket to accept a optional options which can specify the port, this was also my previous comment

dvaldivia (changes requested):

cniackz (changes requested):
Get the token if you are going to use a different port other than `9090` otherwise `401` is observed while testing:

Screen Shot 2022-05-12 at 4 00 14 PM

cniackz:
Hello Adam, good morning. Sorry I did not answer yesterday. I was busy with IBM Cloud Catalog.
Yes, I can see both buckets are being created and I see same error when testing with mc:
$ mc admin bucket remote add myminio/source http://minio:minio123@localhost/target --service "replication" --insecure
mc: <ERROR> Unable to configure remote target. Remote service connection error - please check remote service credentials and target bucket (Remote service endpoint or target bucket not available: target 
	Access Denied.).
Couple of comments:
If you plan on having two servers, I recommend separating them for they to work in parallel, because both in same function only one will work:
	// start console server
	go func() {
		fmt.Println("start server 9090")
		srv, err := initConsoleServer("9090")
		if err != nil {
			log.Println(err)
			log.Println("init fail")
			return
		}
		srv.Serve()
	}()
	go func() {
		fmt.Println("start server 9091")
		srv2, err := initConsoleServer("9091")
		if err != nil {
			log.Println(err)
			log.Println("init fail")
			return
		}
		srv2.Serve()

	}()
2. Second comment is, we only have one container for these tests:
@(docker run -v /data1 -v /data2 -v /data3 -v /data4 --net=mynet123 -d --name minio --rm -p 9000:9000 -p 9001:9001 -e MINIO_KMS_SECRET_KEY=my-minio-key:OSMM+vkKUTCvQs9YL/CVMIMt43HFhkUpqJxTmGl6rYw= $(MINIO_VERSION) server /data{1...4} --console-address ':9001' && sleep 5)
So don't use 9001 as it is collision with the container, try different port, also, I don't think this can be consider as remote because we are using the same container, we should be using two containers for the remote test. Nevertheless, I think we can safely use https://play.min.io for that matter as shown in:
https://github.com/cniackz/minio/wiki/remote-bucket
That way you don't need to create more containers, etc, etc... All you need is two create two buckets with versioning in both sides, in our current container, you can even reuse one if created and in play side with versioning, a simple mc command prior running the test can do the trick for us. This could simplify our testing a lot. But it is up to you as long as it works :slightly_smiling_face: and don't break other tests.

PR #1947, opened 2022-05-03
Add User Service Account screen

jinapurapu:
Screenshot from 2022-05-03 14-09-00


portal-ui/src/common/SecureComponent/permissions.ts
```diff
@@ -126,6 +126,8 @@ export const IAM_PAGES = {
   GROUPS_VIEW: "/identity/groups/:groupName+",
   ACCOUNT: "/identity/account",
   ACCOUNT_ADD: "/identity/new-account",
+  USER_ACCOUNT: "/identity/new-user-sa",
+  USER_ACCOUNT_ADD: "/identity/new-user-sa/:userName+",
```
prakashsvmx, line 130: does it also require an entry in the `IAM_PAGES_PERMISSIONS` ? 
with all or one of the following permissions :
```
admin:CreateServiceAccount
admin:UpdateServiceAccount
admin:RemoveServiceAccount
admin:ListServiceAccounts
```

ravindk89 (dismissed):
Text-wise this looks good. Thanks.

bexsoft (approved):
LGTM, I think we should add the states of user selected tabs to a reducer so when you finish the creation of the service account it displays the service accounts tab.

That may be handled in another PR as it may require a refactor on user details page

dvaldivia (approved):

PR #1956, opened 2022-05-05
Add streaming zip downloads

klauspost:
Do not keep either objects nor the intermediate zip file in memory, and stream both the final zip and the objects as they are read.

Existing code can easily OOM the server.

* Use better compression library :)
* Add modtime to files

Due to error handling not being available early we continue to the next file if we encounter an error.

I haven't tested locally, so please test if possible.


harshavardhana (approved):

prakashsvmx (approved):
Tested. changes look good to me :+1:

dvaldivia (approved):

PR #1957, opened 2022-05-05
Fixed Watch start button to show stop button once watch has started

jinapurapu:
Screenshot from 2022-05-05 11-05-11

### DIFF
--- a/portal-ui/src/screens/Console/Watch/Watch.tsx
+++ b/portal-ui/src/screens/Console/Watch/Watch.tsx
@@ -245,17 +245,7 @@ const Watch = ({
               }}
               variant="standard"
             />
-            {!start && (
-              <Button
-                type="submit"
-                variant="contained"
-                color="primary"
-                onClick={() => setStart(true)}
-              >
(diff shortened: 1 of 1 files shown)


dvaldivia (dismissed):

kaankabalak:
@jinapurapu, if we want to save space, we can alternatively use the [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) as follows:

```
            {start ? (
              <Button
                type="submit"
                variant="contained"
                color="primary"
                onClick={() => setStart(false)}
              >
                Stop
              </Button>
            ) : (
              <Button
                type="submit"
                variant="contained"
                color="primary"
                onClick={() => setStart(true)}
              >
                Start
              </Button>
            )}
```

So, if the `start` condition is true, we'll render the Button that says Stop and if the start condition is false, we would render the Button that says Start

kaankabalak (approved):

dvaldivia (approved):

PR #1964, opened 2022-05-08
Removed error snack bar message for status calls

bexsoft:
## Wat does this do?

Removed error snack bar message & returned `console.error` instead for versioning, quota & object locking API calls in List Objects panel

## How does it look?
Screen Shot 2022-05-07 at 23 16 23


dvaldivia (approved):

kannappanr (approved):
LGTM

PR #1966, opened 2022-05-09
Integration Tests for Get config

adfost:
Test for configs/{name} endpoint


integration/config_test.go
```diff
@@ -56,13 +54,65 @@ func Test_ConfigAPI(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		args           args
+		expectedStatus int
+		expectedError  error
+	}{
+		{
+			name: "Get Config - Valid",
+			args: args{
+				api: "/configs/storage_class",
```
dvaldivia, line 92: All test use the same API, just pass the parameter

restapi/admin_config.go
```diff
@@ -43,7 +43,7 @@ func registerConfigHandlers(api *operations.ConsoleAPI) {
 	api.ConfigurationConfigInfoHandler = cfgApi.ConfigInfoHandlerFunc(func(params cfgApi.ConfigInfoParams, session *models.Principal) middleware.Responder {
 		config, err := getConfigResponse(session, params)
 		if err != nil {
-			return cfgApi.NewConfigInfoDefault(int(err.Code)).WithPayload(err)
+			return cfgApi.NewConfigInfoDefault(404).WithPayload(err)
```
dvaldivia, line 46: This is invalid, some errors might be 400, please process them according to potential error codes coming from MinIO

integration/config_test.go
```diff
@@ -56,13 +54,65 @@ func Test_ConfigAPI(t *testing.T) {
...
+			}
+			if response != nil {
+				assert.Equal(tt.expectedStatus, response.StatusCode, tt.name+" Failed")
+			}
+		})
+	}
+
+}
+
+func Test_GetConfigAPI(t *testing.T) {
```
dvaldivia, line 77: This is duplicate code from the above test

integration/config_test.go
```diff
@@ -56,13 +54,65 @@ func Test_ConfigAPI(t *testing.T) {
 			client := &http.Client{
```
dvaldivia, line 56: You are instantiating a new client on each loop, this is inefficient, reuse the same client

integration/config_test.go
```diff
@@ -56,13 +54,65 @@ func Test_ConfigAPI(t *testing.T) {
...
 
-			requestDataPolicy := map[string]interface{}{}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			client := &http.Client{
+				Timeout: 3 * time.Second,
+			}
 
-			requestDataJSON, _ := json.Marshal(requestDataPolicy)
```
dvaldivia, line 62: Why ignore the response?

dvaldivia (changes requested):

integration/config_test.go
adfost: No this isn't, the above test lists configs with /configs, this test is /configs/{name}

integration/config_test.go
adfost: The code you are highlighting is unnecessary code that I removed, this was from an endpoint that had a body, and these endpoints don't have a body.

integration/config_test.go
dvaldivia: ok

integration/config_test.go
```diff
@@ -31,6 +29,48 @@ import (
...
+		})
+	}
+
+}
+
+func Test_GetConfigAPI(t *testing.T) {
+	assert := assert.New(t)
+
 	type args struct {
 		api string
```
dvaldivia, line 75: this is no longer the `api` it's the `name` parameter, it should be called that and not include the slash

integration/config_test.go
dvaldivia: this endpoint does have a body, it returns a `configuration`
```yaml
  /configs/{name}:
    get:
      summary: Configuration info
      operationId: ConfigInfo
      parameters:
        - name: name
          in: path
          required: true
          type: string
      responses:
        200:
          schema:
            $ref: "#/definitions/configuration"
        default:
          schema:
            $ref: "#/definitions/error"
      tags:
        - Configuration
```

dvaldivia (changes requested):

integration/config_test.go
adfost: It doesn't have a body in the request, only in the response. This code that I removed was to set up a json struct for request body.

integration/config_test.go
```diff
@@ -41,13 +79,21 @@ func Test_ConfigAPI(t *testing.T) {
 		expectedError  error
 	}{
 		{
-			name: "Config - Valid",
+			name: "Get Config - Valid",
 			args: args{
-				api: "/configs",
+				name: "/storage_class",
```
dvaldivia, line 84: the `/` is not needed for the config name, that is part of the URL

dvaldivia (approved):

integration/config_test.go
```diff
@@ -56,12 +102,8 @@ func Test_ConfigAPI(t *testing.T) {
 				Timeout: 3 * time.Second,
 			}
 
-			requestDataPolicy := map[string]interface{}{}
-
-			requestDataJSON, _ := json.Marshal(requestDataPolicy)
-			requestDataBody := bytes.NewReader(requestDataJSON)
 			request, err := http.NewRequest(
-				"GET", fmt.Sprintf("http://localhost:9090/api/v1%s", tt.args.api), requestDataBody)
+				"GET", fmt.Sprintf("http://localhost:9090/api/v1/configs%s", tt.args.name), nil)
```
bexsoft, line 106: Please change this to be 
```suggestion
				"GET", fmt.Sprintf("http://localhost:9090/api/v1/configs/%s", tt.args.name), nil)
```
and remove the slashes from `args.name`

bexsoft (changes requested):

bexsoft (approved):

dvaldivia (approved):

PR #1967, opened 2022-05-10
Simplified Get Bucket details request

bexsoft:
## What does this do?

- Added versioning
- Added responses for quota information
- Added Object locking
- Updated summary panel
- Fixed tests

## How does it look?
Screen Shot 2022-05-09 at 20 14 11

### DIFF
--- a/restapi/user_buckets.go
+++ b/restapi/user_buckets.go
@@ -573,6 +573,7 @@ func getBucketSetPolicyResponse(session *models.Principal, params bucketApi.Buck
 	if err := setBucketAccessPolicy(ctx, minioClient, bucketName, *req.Access, req.Definition); err != nil {
 		return nil, ErrorWithContext(ctx, err)
 	}
+
 	// set bucket access policy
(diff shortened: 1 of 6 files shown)


restapi/user_buckets.go
```diff
@@ -675,6 +681,40 @@ func getBucketInfo(ctx context.Context, client MinioClient, adminClient MinioAdm
 		return nil, err
 	}
 
+	// Versioning Information
+	versioning, err := client.getBucketVersioning(ctx, bucketName)
```
dvaldivia, line 685: mmm, all these sequential calls will slowdown the get bucket info call, since we no longer have timeouts, this could be a really really long time, I'd much prefer to keep each call independent, or parallelize them, but then again we would need to add timeouts which we recently removed

dvaldivia (changes requested):

bexsoft:
Closing PR as this feature is not compatible with recent changes in Console.

PR #1969, opened 2022-05-10
Fix Icons Clipping

dvaldivia:
Fixes Clipping on multiple icons

### Before
Screen Shot 2022-05-10 at 9 55 15 AM

### After
Screen Shot 2022-05-10 at 9 55 25 AM


kaankabalak (approved):

oscarocastellanos (approved):
lgtm

prakashsvmx:
Yes. we need to be careful when converting the SVG icon to a component especially with ` clipPath="url(#clip-path)"`  if it has unique id, then we will not face this issue. else the first clip-path id is applied wherever it is referred.

prakashsvmx (approved):

bexsoft (approved):
LGTM, id & data-name can be removed too

PR #1973, opened 2022-05-11
Fixed disabled Change Password button and added clarifying text

jinapurapu:
Screenshot from 2022-05-10 18-52-55


dvaldivia (dismissed):

portal-ui/src/screens/Console/Account/Account.tsx
```diff
@@ -240,7 +240,7 @@ const Account = ({
                 icon={<PasswordKeyIcon />}
                 color={"primary"}
                 variant={"outlined"}
-                disabled={selectedSAs.length === 0}
+                disabled={false}
```
bexsoft, line 243: If disabled will be always false, then we can remove this prop.

Also, Please review this behavior as I think that when user is logged with IDP this button should disappear

bexsoft (changes requested):

portal-ui/src/screens/Console/Account/Account.tsx
```diff
@@ -97,10 +98,23 @@ const Account = ({
...
+    .invoke("GET", `/api/v1/session`)
+   .then((res: ISessionResponse) => {
+     setUserIDP(res.features.includes("external-idp"))
+   })
+   .catch((err: ErrorResponseHandler) => {
+     setErrorSnackMessage(err);
+   });
+}, []);
+
   useEffect(() => {
```
bexsoft, line 118: This information is stored in a reducer after session login. I think the key is `console.features`. You can omit this session request & check if the string `external-idp` is present in the reducer value

Screen Shot 2022-05-11 at 17 12 32
Screen Shot 2022-05-11 at 17 12 27

bexsoft (changes requested):

portal-ui/src/screens/Console/Account/ChangePasswordModal.tsx
```diff
@@ -120,6 +128,9 @@ const ChangePassword = ({
       }}
       titleIcon={<ChangePasswordIcon />}
     >
+      <div>
+        This will change your Console password. Please note your new password down, as it will be required to log into Console after this session.
```
prakashsvmx, line 132: We could update the text such that it is generic because , same credentials may be used in clients like  mc , or any application(SDK) .

prakashsvmx (approved):
:+1:  Changes look good to me.

bexsoft (approved):
LGTM

dvaldivia (approved):

PR #1978, opened 2022-05-11
Fix route navigation paths to match menus

prakashsvmx:
Fix route navigation paths  (route hierarchy) to match menus

Impacted menu/route paths:

- Buckets -> Add Bucket  ( Will highlight bucket sidebar menu)

- Users -> Create user (Will highlight Users sidebar menu)
  Users -> Click on a user -> Service Accounts ->  Create Service Account  (Will highlight Users sidebar menu)

- Service Accounts -> Create new Service account  ( will highlight the Service Accounts)

- Groups -> Create Group ( will highlight Groups)

### DIFF
--- a/portal-ui/src/common/SecureComponent/permissions.ts
+++ b/portal-ui/src/common/SecureComponent/permissions.ts
@@ -113,21 +113,21 @@ export const IAM_SCOPES = {
 export const IAM_PAGES = {
   /* Buckets */
   BUCKETS: "/buckets",
-  ADD_BUCKETS: "/add-bucket",
+  ADD_BUCKETS: "/buckets/add-bucket",
   BUCKETS_ADMIN_VIEW: "/buckets/:bucketName/admin*",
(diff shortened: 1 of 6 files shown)


dvaldivia (dismissed):

adfost (approved):

jinapurapu (approved):
If a user has no policies, the Create Service Account button on the User Details screen is disabled, is that correct behavior?

harshavardhana:
> If a user has no policies, the Create Service Account button on the User Details screen is disabled, is that correct behavior?

Those service accounts without policies for their parent user won't be useful. Perhaps it is a good idea to disable them.

dvaldivia:
@jinapurapu @harshavardhana yes, I thought we already did that, if not in a separate PR let's disable the create SA for users without policies which we can validate via the session redux state

jinapurapu:
I understood why it is the way it is after I typed that up and forgot to delete it before approving. I think it is behaving correctly.

PR #1988, opened 2022-05-12
Changed styles of Object versions for small screens

bexsoft:
## What does this do?

Changed styles of Object versions for small screens

## How does it look?
Screen Shot 2022-05-11 at 22 44 12


dvaldivia:
I think that for this resolution, we should decrease the padding of the page layout, but that can be done on a separate PR

prakashsvmx (approved):
:+1:

PR #1990, opened 2022-05-12
Added copy path button to breadcrumbs bar

bexsoft:
## What does this do?

Added copy path button to breadcrumbs bar

## How does it look?

Screen Shot 2022-05-12 at 14 14 10
Screen Shot 2022-05-12 at 14 14 06


jinapurapu:
If a path is created with a single special character, the path displayed in the breadcrumbs is incorrect - the special character name is to the left of the parent rather than to the right. The copy path does give the correct path, it is only the displayed breadcrumbs that is off. 
Screenshot from 2022-05-12 12-38-13
Screenshot from 2022-05-12 12-38-25

.

oscarocastellanos (approved):
LGTM

dvaldivia (approved):

PR #1991, opened 2022-05-12
Upgrade Redux to Redux Toolkit

dvaldivia:
Upgrading the Redux library to the new recommended Redux Toolkit, which is compatible with the current code, replacing the current dependency and updating the create store function is enough to do this.


bexsoft (changes requested):
Uploading a new file / folder is throwing a redux error
Screen Shot 2022-05-12 at 19 37 01

Create tenant screen hard crash trying to display initial information
Screen Shot 2022-05-12 at 19 39 06

Hard crash on pools creation & edit wizards
Screen Shot 2022-05-12 at 19 40 52

PR #1998, opened 2022-05-13
Get access rules test and test for adding access rule to non existent bucket

adfost:
Added the increment to this PR:

Screen Shot 2022-05-24 at 7 55 04 PM


integration/access_rules_test.go
```diff
@@ -94,3 +94,61 @@ func Test_AddAccessRuleAPI(t *testing.T) {
...
+			},
+			expectedStatus: 200,
+			expectedError:  nil,
+		},
+		{
+			name: "Get Access Rule - Invalid",
+			args: args{
+				bucket: "testaccessruleget2",
+			},
+			expectedStatus: 200,
```
dvaldivia, line 125: invalid should not return `200`, it should return `404`

dvaldivia (changes requested):

integration/access_rules_test.go
adfost: ok we just showed the code is broken. I can fix it.

cniackz (changes requested):
Also remember to squash the commits from many to one 👍

cniackz (approved):

cniackz (approved):
LGTM

cniackz (approved):
LGTM

dvaldivia (approved):

bexsoft (approved):

PR #2006, opened 2022-05-17
Store bucket path if it is redirected to restore on login

prakashsvmx:
Store bucket path if it is redirected to restore on login

Steps:
Share a Bucket/Prefix path 
Open the link in a new Browser. -> Complete the Login and the path should be restored.


bexsoft (approved):
LGTM

dvaldivia (approved):

bexsoft (approved):

PR #2017, opened 2022-05-19
Update Operator API Test

cniackz:
@dvaldivia did an excellent Job in finding the root cause of this problem.
It was matter of updating `k8s` and `kind` to reproduce the issue locally.
To fix this problem, we should get token in a new way:

```sh
kubectl apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: console-sa-secret
  namespace: minio-operator
  annotations:
    kubernetes.io/service-account.name: console-sa
type: kubernetes.io/service-account-token
EOF


SA_TOKEN=$(kubectl -n minio-operator  get secret console-sa-secret -o jsonpath="{.data.token}" | base64 --decode)
echo $SA_TOKEN
```

Then assess whether we obtained the token or not and based on that we either fail or proceed with the test.

### DIFF
--- a/operator-integration/tenant_test.go
+++ b/operator-integration/tenant_test.go
@@ -126,50 +126,40 @@ func TestMain(m *testing.M) {
 	go func() {
 		fmt.Println("start server")
 		srv, err := initConsoleServer()
+		fmt.Println("Server has been started at this point")
 		if err != nil {
+			fmt.Println("There is an error in console server: ", err)
 			log.Println(err)
 			log.Println("init fail")
 			return
 		}
+		fmt.Println("Start serving with Serve() function")
 		srv.Serve()
+		fmt.Println("After Serve() function")
 	}()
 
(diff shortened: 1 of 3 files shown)


dvaldivia (approved):

Alevsk (approved):

operator-integration/tenant_test.go
```diff
@@ -126,50 +126,40 @@ func TestMain(m *testing.M) {
 	go func() {
 		fmt.Println("start server")
 		srv, err := initConsoleServer()
+		fmt.Println("Server has been started at this point")
 		if err != nil {
+			fmt.Println("There is an error in console server: ", err)
 			log.Println(err)
 			log.Println("init fail")
 			return
```
reivaj05, line 134: I believe this "return" will continue passing the job as successful even though `initConsoleServer` returns an error

PR #2023, opened 2022-05-20
Increase coverage threshold

cniackz:
I remember we mentioned that we were at 50.00% but I am not sure if that is currently true anymore.
I will increase the code coverage threshold as much as possible with latest code to keep things updated:

Screen Shot 2022-05-20 at 5 17 51 PM


reivaj05 (approved):

bexsoft (approved):
LGTM

PR #2028, opened 2022-05-22
fix: capacity reported usage value

harshavardhana:
do not use unix-epoch to be displayed, instead
use the actual value at that epoch to be displayed.

### DIFF
--- a/portal-ui/src/common/utils.ts
+++ b/portal-ui/src/common/utils.ts
@@ -59,14 +59,10 @@ export const niceBytesInt = (n: number, showK8sUnits: boolean = false) => {
   while (n >= 1024 && ++l) {
     n = n / 1024;
   }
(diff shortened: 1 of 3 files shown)


harshavardhana:
To test this PR 

```
$ gh pr checkout 2028
$ make assets && make install
$ CONSOLE_PROMETHEUS_URL=http://play.min.io:9091 CONSOLE_MINIO_REGION=us-east-1 CONSOLE_MINIO_SERVER=https://play.min.io:9000 console srv
```

harshavardhana:
Also the display box is wrong, the value here is currently "free" space instead - its reporting it as `used space`

dvaldivia (approved):

prakashsvmx (approved):
LGTM :+1:

bexsoft (approved):
LGTM

PR #2029, opened 2022-05-23
fix: getTimeFromTimestamp() incorrect padding for days

harshavardhana:
use a simpler function to return localeString() instead.


Alevsk (approved):

prakashsvmx (approved):
:+1:

PR #2058, opened 2022-05-31
Disabled Assign Policy button if multiple groups selected

jinapurapu:
Jill:
```
If one selected multiple groups on the Groups screen, then clicked the Assign Policy button, 
only the first group selected is passed to the SetPolicy modal, which may be confusing to users. 
I will work on a functional set policy for multiple groups in another PR, but by only enabling the 
Assign Policy button when a single group is selected prevents confusion in the current UI
```

Daniel
```
in that case shouldn't we fix the checkedGroups ? to make sure it includes all the selected groups?
```

Jill:
```
yes, I will work on that, this is just a stopgap
```


cniackz (approved):
As a stopgap it looks good to me, later on there will be another PR for it.

bexsoft (approved):
LGTM. Let's improve logic on assign groups in an upcoming PR. Thank you!

PR #2082, opened 2022-06-03
Add bucket replication test

cniackz:
To test these endpoints using `https://play.min.io` as our remote bucket:

```yaml
  /remote-buckets:
    get:
      summary: List Remote Buckets
      operationId: ListRemoteBuckets
      responses:
        200:
          schema:
            $ref: "#/definitions/listRemoteBucketsResponse"
        default:
          schema:
            $ref: "#/definitions/error"
      tags:
        - Bucket
    post:
      summary: Add Remote Bucket
      operationId: AddRemoteBucket
      parameters:
        - name: body
          in: body
          required: true
          schema:
            $ref: "#/definitions/createRemoteBucket"
      responses:
        201:
        default:
          schema:
            $ref: "#/definitions/error"
      tags:
        - Bucket
```

I believe this `play` approach is better and easier than having to create two internal containers with two internal servers because `play` is already configured and easy to use, we require less lines of code to perform the same testing and it will actually test a real remote bucket connection rather than having two containers under the same or different network. Plus we are already using `play` for coverage and it has been proven useful.

### To test this in IntelliJ or GoLand

Make sure to set the environment variable as our `Makefile` does, to do so edit the configurations and set the environmental variables:

Screen Shot 2022-06-03 at 2 38 31 PM

Where you should get the bucket from play: `TARGET_BUCKET=d39b1f1167408636e0ae`

### Coverage:

```sh
-          threshold=51.00
+          threshold=51.40
```

### DIFF
--- a/.github/workflows/jobs.yaml
+++ b/.github/workflows/jobs.yaml
@@ -1368,7 +1368,22 @@ jobs:
           echo "We are going to use the built image on test-integration";
           VERSION="minio/minio:$VERSION";
           echo $VERSION;
-          make test-integration MINIO_VERSION=$VERSION;
+          
+          echo "Create bucket for replication with versioning"
+          echo "Download mc for Ubuntu"
+          wget -q https://dl.min.io/client/mc/release/linux-amd64/mc
+          echo "Change the permissions to execute mc command"
+          chmod +x mc
+          echo "Create the folder to put the all.out file"
+          TARGET_BUCKET=`echo $RANDOM | md5sum | head -c 20; echo;`
+          echo "TARGET_BUCKET: ${TARGET_BUCKET}"
+          ./mc mb --ignore-existing play/${TARGET_BUCKET}/
+          ./mc version enable play/${TARGET_BUCKET}
+          # Via API we are going to test:
+          # mc admin bucket remote add myminio/source https://minioadmin:minioadmin@play.min.io/target --service "replication"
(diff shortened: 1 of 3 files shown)


.github/workflows/jobs.yaml
```diff
@@ -1368,6 +1368,20 @@ jobs:
...
           echo $VERSION;
+          
+          echo "Create bucket for replication with versioning"
+          echo "Download mc for Ubuntu"
+          wget -q https://dl.min.io/client/mc/release/linux-amd64/mc
+          echo "Change the permissions to execute mc command"
+          chmod +x mc
+          echo "Create the folder to put the all.out file"
+          ./mc rb play/target
+          ./mc mb --ignore-existing play/target/
```
dvaldivia, line 1379: the name of the bucket should be randomized else, two PRs might hit a race condition where the bucket is not ready for use and fail the test

dvaldivia (changes requested):

.github/workflows/jobs.yaml
cniackz: Good catch Daniel, let me work on that as a race condition can definitely break this approach.

cniackz: Resolved, I am using a random bucket name

cniackz:
Currently WIP until all the tests pass

bexsoft (dismissed):

Alevsk (approved):

bexsoft (approved):

PR #1798, opened 2022-04-05
Operator Console improvements

bexsoft:
##What does this do?

- Added new design to Tenants page list
- Added Pool details initial page

## How does it look?

Screen Shot 2022-04-04 at 23 07 01
Screen Shot 2022-04-04 at 23 06 50
Screen Shot 2022-04-04 at 23 06 41

Signed-off-by: Benjamin Perez <benjamin@bexsoft.net>

PR #2083, opened 2022-06-03
Have TenantItem charts use capacity instead of raw capacity

dvaldivia:


bexsoft (approved):

cniackz:
Tested this change to understand how the pie color changes based on original PR:
https://github.com/minio/console/pull/1798

Screen Shot 2022-06-03 at 5 15 12 PM

Screen Shot 2022-06-03 at 5 15 03 PM

Screen Shot 2022-06-03 at 5 14 33 PM

cniackz:
I guess `tenant.capacity` will turn into different colors in a more accurate way I believe since it is more near to the used value that the raw capacity, interesting change!.

cniackz (approved):
LGTM, tested!

PR #2086, opened 2022-06-04
Assign policy for Multiple groups 

jinapurapu:
Screenshot from 2022-06-03 15-25-40

Screenshot from 2022-06-03 15-26-01
Screenshot from 2022-06-03 15-26-13


cniackz:
It looks good this change, but in a next PR we should assign the policy only once if already is assigned, there is no need to have same policy assigned again:

Screen Shot 2022-06-03 at 9 57 17 PM

cniackz (approved):
LGTM, next PR let's assign the policy only once.

bexsoft (approved):
LGTM

PR #2099, opened 2022-06-07
Added Initial Time support to Prometheus dashboard

bexsoft:
## What does this do?

Added Initial Time support to Prometheus dashboard metrics, also incremented initial time for data usage growth chart

## How does it look?
Screen Shot 2022-06-07 at 14 34 55


dvaldivia (dismissed):

cniackz:
I got prometheus working locally in port 9000 with these instructions https://docs.min.io/minio/baremetal/monitoring/metrics-alerts/collect-minio-metrics-using-prometheus.html?ref=con#minio-metrics-collect-using-prometheus
But how you test this on port 5005? 🤔

bexsoft:
> I got prometheus working locally in port 9000 with these instructions https://docs.min.io/minio/baremetal/monitoring/metrics-alerts/collect-minio-metrics-using-prometheus.html?ref=con#minio-metrics-collect-using-prometheus But how you test this on port 5005? 🤔

You can start console server using this flag: `CONSOLE_PROMETHEUS_URL=http://localhost:9091/` changing the port according your prometheus settings (just be aware prometheus is using the same ports as MinIO & Console)

cniackz (approved):
It looks good to me even when no data is ignored from the graph.

cniackz:
Actually this is much more accurate even when it looks ugly is more accurate:

Screen Shot 2022-06-07 at 9 48 46 PM

### Code change:

```go
$ git diff restapi/admin_info.go
diff --git a/restapi/admin_info.go b/restapi/admin_info.go
index 31434a07..31a49efa 100644
--- a/restapi/admin_info.go
+++ b/restapi/admin_info.go
@@ -296,7 +296,7 @@ var widgets = []Metric{
                                Expr:         `sum(minio_bucket_usage_total_bytes{$__query}) by (instance)`,
                                LegendFormat: "Used Capacity",
                                InitialTime:  -180,
-                               Step:         900,
+                               Step:         10,
                        },
                },
        },
```

restapi/admin_info.go
```diff
@@ -294,6 +295,8 @@ var widgets = []Metric{
 			{
 				Expr:         `sum(minio_bucket_usage_total_bytes{$__query}) by (instance)`,
 				LegendFormat: "Used Capacity",
+				InitialTime:  -180,
+				Step:         900,
```
cniackz, line 299: `900` will give a very small number of steps and graph will be confusing and not accurate enough:

image

for the same data with `Steps` equal `10` we can get a much better resolution of what I did during these hours:

image

cniackz (changes requested):
Can we discuss this tomorrow? For me to show you my scenario, then we can decide if we change the `Step` value.

cniackz:
My question is, do we want to show just an approximation of the usage or real data?...
If we just want an approximation then this change might be ok, but for some customers I bet this will be confusing as graph may not contain desired points in time.

restapi/admin_info.go
bexsoft: Agree on this, what happens if we set Step to 60? does it still show good information? 

The issue with setting to 10 is that the chart will slow down on old computers as we will be throwing too much information to plot each point in the chart this can lead to a hard hang in the application. I think a value between 60 and 300 could work, what do you think?

cniackz: Let me quickly test it, one sec

cniackz: image

In a quick run we are going to still have some innacurate behaviour with `300` but it will be much better than `900` I believe. I guess this is more like a aestetic change to perform. So let's try to put this in 300 and adjust according to the end user input.

bexsoft:
Setting Step to 10, will see how it behaves, in case something goes strange, we can increment this value

PR #2106, opened 2022-06-08
Reduce renders in User Name when adding a user

cniackz:
## What does this do?

Reduce renders in User Name when adding a user (Notice only that field is being rendered, not the entire screen):

reduce-renders

### DIFF
--- a/portal-ui/src/screens/Console/Users/AddUserScreen.tsx
+++ b/portal-ui/src/screens/Console/Users/AddUserScreen.tsx
@@ -14,13 +14,15 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import React, { Fragment, useState } from "react";
+import UserSelector from "./UserSelector";
+import React, { Fragment } from "react";
 import { Theme } from "@mui/material/styles";
 import createStyles from "@mui/styles/createStyles";
 import withStyles from "@mui/styles/withStyles";
+import {createUserAsync, resetFormAsync} from "./thunk/AddUsersThunk";
 import {
-  formFieldStyles,
-  modalStyleUtils,
+    formFieldStyles,
+    modalStyleUtils,
 } from "../Common/FormComponents/common/styleLibrary";
 import Grid from "@mui/material/Grid";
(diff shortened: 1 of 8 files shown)


portal-ui/src/screens/Console/Users/AddUserScreen.tsx
```diff
@@ -119,7 +126,7 @@ const AddUser = ({ classes }: IAddUserProps) => {
 
   const resetForm = () => {
     setSelectedGroups([]);
-    setAccessKey("");
+    setUserName("");
```
dvaldivia, line 129: you probably need to `dispatch()` this, and in that case you can start wariting a `resetForm()` action on the slice that resets the whole state to `initialState`

dvaldivia (dismissed):

cniackz:
Wait, this is breaking the `Clear` functionality and I still don't know why, will continue tomorrow, I am tired of trying with no success for many hours... Will ping Daniel tomorrow to get some advice on this. Mui Button onclick property.

cniackz:
Cannot clear: 

Screen Shot 2022-06-08 at 9 01 45 PM

cniackz:
By doing this activity and spending the whole day, I have learned more of Redux/React than any other activity I have done so far. Now it is clear to me how to states are managed and how things are organized. I am becoming better at this. By the way, it is not the same reading or taking small tuto than performing this type of activities, amazing! And glad to contribute with this project. 👍

cniackz:
By the way:

1. Creating a user -> PASS
2. Clearing the fields (resetForm) -> PASS
3. Selecting Policies -> PASS

I think this change isn't breaking current functionality.
Please double check!

cniackz:
Please don't review this for now, I am about to add more changed to complete this PR.

dvaldivia (approved):

cniackz:
I will resolve the conflict later today or tomorrow morning.

cniackz:
Please don't approve, there is a bug in this PR.
After adding the user, the page stay on the same place but it should move back to user.
Need to investigate why...

dvaldivia (approved):

Alevsk (approved):

PR #2126, opened 2022-06-15
Added decimal support to bucket quota selectors

bexsoft:
## What does this do?
- Added decimal support to bucket quota selectors
- Fixed an issue with calculateBytes function
- Fixed add bucket form validations

## How does it look?
Screen Shot 2022-06-14 at 23 07 20
Screen Shot 2022-06-14 at 23 07 10
Screen Shot 2022-06-14 at 23 07 04
Screen Shot 2022-06-15 at 14 55 46
Screen Shot 2022-06-15 at 14 55 37

### DIFF
--- a/portal-ui/src/common/utils.ts
+++ b/portal-ui/src/common/utils.ts
@@ -596,7 +596,7 @@ export const calculateBytes = (
   // Get unit for measure
   const i = Math.floor(Math.log(bytes) / Math.log(k));
 
-  const fractionDigits = showDecimals ? 0 : 1;
+  const fractionDigits = showDecimals ? 1 : 0;
 
   const bytesUnit = bytes / Math.pow(k, i);
 
--- a/portal-ui/src/history.ts
+++ b/portal-ui/src/history.ts
@@ -1,5 +1,3 @@
 // check if we are using base path, if not this always is `/`
 const baseLocation = new URL(document.baseURI);
 export const baseUrl = baseLocation.pathname;
-
-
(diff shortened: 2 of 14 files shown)


prakashsvmx (commented):
Observations:

With `mc`
```
mc admin bucket quota local/b-bucket --hard 1.7GB
Successfully set bucket quota of 1.6 GiB with hard type on `b-bucket`

mc admin bucket quota local/b-bucket 
Bucket `b-bucket` has hard quota of 1.6 GiB

```
so as per the above mc commands, `~0.1` is lesser capacity is set for quota 

but if we update from the UI, it sets the exact value. 
`e.g 2.6 GB gets updated as 2.6 GB`


Otherwise changes look good to me :+1:

bexsoft:
> Observations:
> 
> 1. We could apply same changes in Create Bucket screen as well.
> 2. With `mc`
> 
> ```
> mc admin bucket quota local/b-bucket --hard 1.7GB
> Successfully set bucket quota of 1.6 GiB with hard type on `b-bucket`
> 
> mc admin bucket quota local/b-bucket 
> Bucket `b-bucket` has hard quota of 1.6 GiB
> ```
> 
> so as per the above mc commands, `~0.1` is lesser capacity is set for quota
> 
> but if we update from the UI, it sets the exact value. `e.g 2.6 GB gets updated as 2.6 GB`
> 
> Otherwise changes look good to me 👍

I think this is related to GB vs GiB conversion, so this should be expected

cniackz (approved):
LGTM

dvaldivia (approved):

PR #2137, opened 2022-06-21
Added Rename modal for filenames longer than 200 characters in Windows

bexsoft:
## What does this do?

Added Rename modal for filenames longer than 200 characters in Windows

## How does it look?

Captura3
Captura
Captura2


cniackz:
Hello @bexsoft 

How do you test this change?. I put a file name of 201 characters, but I can download without any message on screen.
I am guessing this has to be tested on Windows only to see that modal, right?...

cniackz (approved):
LGTM

bexsoft:
> Hello @bexsoft
> 
> How do you test this change?. I put a file name of 201 characters, but I can download without any message on screen. I am guessing this has to be tested on Windows only to see that modal, right?...

That's correct. Modal should only be visible on windows.

prakashsvmx:
@bexsoft  a minor observation

I have used browserstack for testing

When  checkbox selection is used and clicking download does not prompt rename modal.
but when a row is clicked and download is pressed, then the rename modal is shown

https://user-images.githubusercontent.com/23444145/174941386-d1f43eec-9fc6-459f-95d0-b44281df3104.mp4


Also in the rename dialog, we could add more than 250 characters
image

bexsoft:
> @bexsoft a minor observation
> 
> I have used browserstack for testing
> 
> When checkbox selection is used and clicking download does not prompt rename modal. but when a row is clicked and download is pressed, then the rename modal is shown
> 
>  rename_modal.mp4 
> Also in the rename dialog, we could add more than 250 characters image

Yes @prakashsvmx, In this case we want to enforce the zip download when checkbox is used to start the download, Let me check the length of the input issue

reivaj05 (approved):

PR #2140, opened 2022-06-23
Added improvements to Rename modal in windows

bexsoft:
## What does this do?

- Disabled button when filename is less than 200 characters long
- Selector to override & accept long name
- Added this behavior when one file is checked in objects list

## How does it look?
Screen Shot 2022-06-22 at 20 14 16
Screen Shot 2022-06-22 at 20 14 06
Screen Shot 2022-06-22 at 20 14 02


prakashsvmx (approved):
Tested. Changes Look good to me :+1:

PR #2144, opened 2022-06-23
Security context component stack in column on resize

jinapurapu:
Screenshot from 2022-06-23 14-50-23

### DIFF
--- a/portal-ui/src/screens/Console/Tenants/securityContextSelector.tsx
+++ b/portal-ui/src/screens/Console/Tenants/securityContextSelector.tsx
@@ -1,0 +1,133 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2022 MinIO, Inc.
+//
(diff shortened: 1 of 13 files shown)


portal-ui/src/screens/Console/Tenants/securityContextSelector.tsx
```diff
@@ -0,0 +1,132 @@
...
+    runAsNonRoot: boolean;
+    setRunAsUser: any;
+    setRunAsGroup: any;
+    setFSGroup: any;
+    setRunAsNonRoot: any;
+}
+
+const styles = () =>
+  createStyles({
+    tabsContainer: {
```
prakashsvmx, line 39:
```suggestion
    tabsContainer: {
      display: "flex",
      height: "100%",
      width: "100%",
      gap: "8px"
    }
```

`gap` property could be added to give spacing horizontally and vertically in responsive mode.
e.g:
image

image

PR #2146, opened 2022-06-24
UX Top bar height

prakashsvmx:
Fixes #https://github.com/miniohq/engineering/issues/890

How does it look like:
image

image


cniackz (approved):
LGTM, no test is being broken by this change.

PR #2150, opened 2022-06-27
Add object-level error message display in Downloads/Uploads panel

### LINKED ISSUE #2111: Display detailed error message on upload failure
## Current behaviour

As of now, when an error occurs during file upload, a snackbar is displayed in the upper portion of the screen and the progress bar of the affected file turns red:

minio-upload-error

Neither the snackbar message (_Uploaded files X/Y_) nor the _Downloads/Uploads_ list give the user a clear indication of what actually went wrong. Consequently, in order to figure out the possible reason behind that error, the user has to open dev tools and inspect the HTTP response.

## Possible solution

It seems that detailed error messages are already extracted from the server response (but not stored in `objectBrowserSlice.objectManager`):

https://github.com/minio/console/blob/e3d96b5bb396200541f930f5fb5e7ae1e29dfac6/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx#L912-L923

The error message could be displayed in the _Downloads/Uploads_ list below the progress bar, like so (mock):

error-poc

Let me know your feedback on this issue. I can open a PR this weekend that will implement the solution visualized above.

kffl:
Resolves #2111

This PR adds support for displaying object-level error messages in the Downloads/Uploads panel. It works for both uploads and downloads (including downloads initiated via the object version list).

While working on this feature, I have noticed that the `/buckets/{bucket_name}/objects/download` endpoint produces empty body (content-type of which is declared as `application/octet-stream`) and a 403 code when the user isn't allowed to perform `GetObject` action instead of returning an object conforming to `#/definitions/error`. Is that behavior expected or is it a potential issue? For the purpose of this PR, I've worked around that by composing an error message that includes only the status code if the response body is not available.

<details>
  <summary>Screenshots</summary> 

Download error (missing GetObject permission):

download-error
Upload error (missing PutObject permission):

upload-error-message
</details>

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
@@ -783,8 +783,8 @@ const ListObjects = () => {
       () => {
         dispatch(completeObject(identityDownload));
       },
-      () => {
-        dispatch(failObject(identityDownload));
+      (msg: string) => {
+        dispatch(failObject({ instanceID: identityDownload, msg }));
       },
       () => {
         dispatch(cancelObjectInList(identityDownload));
@@ -804,6 +804,7 @@ const ListObjects = () => {
         waitingForFile: true,
         failed: false,
         cancelled: false,
+        errorMessage: "",
       })
     );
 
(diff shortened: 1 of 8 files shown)


bexsoft (approved):
Tested, LGTM. Thank you!
Screen Shot 2022-06-28 at 14 37 34

prakashsvmx (approved):
Tested. Changes look good to me. :+1: 

image

PR #2159, opened 2022-07-01
Added Prettier test to workflow

bexsoft:
## What does this do?

Added a test to check if files are prettified inside portal-ui

To prettify the files now onwards, we must run `make prettify` inside portal-ui folder. Also Prettier must be version `2.7.1`


Alevsk (approved):

prakashsvmx (approved):
Thank you. :+1:

PR #2167, opened 2022-07-11
Allow multiple IDPs config to be passed via struct

donatello:
- This removes support for env based IDP configuration for console

### DIFF
--- a/cmd/console/server.go
+++ b/cmd/console/server.go
@@ -98,7 +98,7 @@ func buildServer() (*restapi.Server, error) {
 		return nil, err
 	}
 
-	api := operations.NewConsoleAPI(swaggerSpec)
+	api := operations.NewConsoleAPI(swaggerSpec, nil)
 	api.Logger = restapi.LogInfo
 	server := restapi.NewServer(api)
 
--- a/integration/buckets_test.go
+++ b/integration/buckets_test.go
@@ -71,7 +71,7 @@ func initConsoleServer() (*restapi.Server, error) {
 	restapi.LogInfo = noLog
 	restapi.LogError = noLog
 
-	api := operations.NewConsoleAPI(swaggerSpec)
+	api := operations.NewConsoleAPI(swaggerSpec, nil)
 	api.Logger = noLog
 
 	server := restapi.NewServer(api)
--- a/pkg/auth/idp/oauth2/config.go
+++ b/pkg/auth/idp/oauth2/config.go
@@ -25,6 +25,21 @@ import (
 	"github.com/minio/pkg/env"
 )
 
+// ProviderConfig - OpenID IDP Configuration for console.
+type ProviderConfig struct {
+	URL                      string
+	ClientID, ClientSecret   string
+	HMACSalt, HMACPassphrase string
+	Scopes                   string
+	Userinfo                 bool
+	RedirectCallbackDynamic  bool
+	RedirectCallback         string
+}
+
+type OpenIDPCfg map[string]ProviderConfig
+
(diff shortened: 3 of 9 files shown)


cniackz:
- I got the steps from manual testing prior these changes
- If env vars are no longer supported then:

1. How can I test this manually? Before I was doing this:

```sh
export CONSOLE_IDP_URL=http://dex:5556/dex/.well-known/openid-configuration
export CONSOLE_IDP_CLIENT_ID=minio-client-app
export CONSOLE_IDP_SECRET=minio-client-app-secret
export CONSOLE_IDP_CALLBACK=http://127.0.0.1:9090/oauth_callback
cd ~/console                                                               
make install                                 
~/go/bin/console server  
```

But above is no longer working, so how can we test this change?...

2. I don't know why this is failing, I compare between working scenario and failing scenario but I don't know why we don't get the token with this changes yet...

3. How did we test this is working with console?... I want to learn that, the reason is to get the steps so that I can update the test if this is actually working...

Tomorrow I will synch with Donatello on this to learn the testing part since I haven't been able to figure this out yet...

Also if this change is still work in progress, not complete, and there is something missing from console point of view; it might be good idea to involve Daniel and Lenin on this as well, so that we can get this faster...

donatello:
@cniackz This way:

```
export CONSOLE_IDP_URL=http://dex:5556/dex/.well-known/openid-configuration
export CONSOLE_IDP_CLIENT_ID=minio-client-app
export CONSOLE_IDP_SECRET=minio-client-app-secret
export CONSOLE_IDP_CALLBACK=http://127.0.0.1:9090/oauth_callback
cd ~/console                                                               
make install                                 
~/go/bin/console server  
```

will not work, because this env support is what is being removed (when running as minio console - it should continue to work in operator mode).

In the test, in this change, I am already passing the IDP info via lines like:

image

in `sso_test.go`.

I locally tested this change, by building minio and having local dex service set up.

cniackz:
@donatello found the issue, it is the scopes.

Previously the scopes were like this:

scopes:

- openid
- profile
- email

Then we appended again openid and ended up with 4:

- openid
- profile
- email
- openid

And that was working fine before.
With these changes I see different pattern, none is obtained and we just append openid so we have:

- "" empty blank space
- openid

And this isn't working, we can't retrieve the credentials and we end up getting the 500 due to different scopes:

Screen Shot 2022-07-13 at 1 28 18 PM

But if we correct these line and replace with the old one, then tests will pass again and no issue will be observed while using SSO:

Screen Shot 2022-07-13 at 1 37 32 PM

If this is not intended; I suggest finding another way to get proper scopes to get this working. But I believe this is not a problem with the test but a problem in the code. Test is only trying to login as the user would do via Browser, so it is exposing real problem I think.

Let me correct the code after this and see if the test passes in Github as well...

cniackz:
This is what I will commit to test and prove my point:

Screen Shot 2022-07-13 at 1 43 30 PM

cniackz:
Yes, all tests are passing

donatello:
Thanks @cniackz for debugging, it's clear now that my change didn't have a default set of scopes to use when nothing was specified. This is fixed now, using the old default value for scopes. This should get all the tests passing.

donatello:
This is ready for review

cniackz (approved):
LGTM

harshavardhana (approved):
Based on this I think we can come up with `ConsoleContext` struct that encompasses all features that are re-parsed again by console that can be avoided.

Allowing for a more typed approach to the settings to be passed into Console.

cniackz (approved):
LGTM

PR #2178, opened 2022-07-19
Add support for adding LDAP admins based on user/group DNs

Alevsk:
This is useful for when users are locked out, or don't know the immediate DNs for users/groups of their tenant, when they try to login they will get an error that looks like this: 

```
Expecting a policy to be set for user `uid=dillon,ou=people,ou=swengg,dc=min,dc=io` or one of their groups: `cn=projecta,ou=groups,ou=swengg,dc=min,dc=io`,`cn=projectb,ou=groups,ou=swengg,dc=min,dc=io` - rejecting this request
```
Then they can go back to Operator-Console and quickly add that user/group under `identity-provider > LDAP > users`

Signed-off-by: Lenin Alevski <alevsk.8772@gmail.com

Screenshot from 2022-07-18 17-44-36
>


cniackz (approved):
LGTM

dvaldivia (approved):

PR #2196, opened 2022-07-24
fix:Incorrect template

jiuker:
The first input to this function is the template. If it is a normal string error it will show the wrong string.
like:
```
E: 2022/07/24 21:15:05 Get "": unsupported protocol scheme ""
%!(EXTRA *url.Error=Get "": unsupported protocol scheme "", *errors.errorString=unable to contact configured identity provider)
```


bexsoft (approved):
LGTM. @Alevsk Any thoughts?

Alevsk (approved):

PR #2199, opened 2022-07-26
Add button to download widget data as CSV file

jinapurapu:
Screenshot from 2022-07-25 21-16-33
Screenshot from 2022-07-25 21-16-40

Screenshot from 2022-07-25 21-16-25
closes https://github.com/miniohq/engineering/issues/661


prakashsvmx (commented):
I tried on a fresh cluster the download is not functioning due to errors
Also the layout seems to be little off for couple of widgets

<details>

image

</details>

portal-ui/src/screens/Console/Dashboard/Prometheus/Widgets/WidgetDownloadButton.tsx
```diff
@@ -0,0 +1,98 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2021 MinIO, Inc.
```
reivaj05, line 2: 2022

reivaj05 (dismissed):

dvaldivia:
@oscarocastellanos can you give us your thoughts on this button placement/design? We can improve on this on subsequent PRs

oscarocastellanos:
It's best to have the button to the right. We prefer to have the title first always and have it this way feels weird. Maybe we can get rid off the "Expand graph " text and just keep in there the icons.

dvaldivia:
@oscarocastellanos @jinapurapu what about a combo button which defaults to `expand graph` but upon clicking the arrow you can get the option to download CSV?

for example
Screen Shot 2022-08-02 at 10 46 32 AM

oscarocastellanos:
What I would be worried about is that those are two different type of actions. Expand is a visualization action while download is an action of a very different nature. I would be worried that the grouping of those two wouldn't seem natural @dvaldivia @jinapurapu

oscarocastellanos:
_unless_ we rename the expand action to something closer to the same nature as download @dvaldivia 🤔

bexsoft:
> _unless_ we rename the expand action to something closer to the same nature as download @dvaldivia 🤔

I made a couple of proposals for this, What do you think @oscarocastellanos @dvaldivia?

Basically, Why don't we remove the Expand Graph label (having an icon should be enough), and add the download icon (or something more representative) next to the expand one or inside a floating box that could appear on hover.

examples

oscarocastellanos:
I like this idea @bexsoft ! What about having the download icon always visible but the one that shows on hover is the expand icon? @dvaldivia

PR #2203, opened 2022-07-27
Added DirectPV mode to Operator console

bexsoft:
## What does this do?

Adds DirectPV mode support to operator console.

To start in DirectPV Mode, tags should be the same as operator, also `DIRECTPV_MODE=on` env variable is required.

**NOTE:** This PR only includes UI, APIs and their connections will be handled in a upcoming PRs

## How does it look?

Screen Shot 2022-07-27 at 18 40 21
Screen Shot 2022-07-27 at 18 40 12
Screen Shot 2022-07-27 at 18 40 02
Screen Shot 2022-07-27 at 18 39 55


Praveenrajmani (approved):
just a small question - shouldn't we list only the PVCs which has "directpv-min-io" storage-class set? instead on listing all the PVCs?

LGTM otherwise

jinapurapu:
In Firefox, after clicking Documentation tab, tabs no longer direct to new screens.

reivaj05 (approved):

PR #2217, opened 2022-08-01
Auto register API key once it is retrieved from subnet

reivaj05:
Auto register to SUBNET once API Key is retrieved.


portal-ui/src/screens/Console/Support/Register.tsx
```diff
@@ -55,17 +55,13 @@ import {
...
-
 import RegisterHelpBox from "./RegisterHelpBox";
-import RegistrationStatusBanner from "./RegistrationStatusBanner";
 import { selOpMode, setErrorSnackMessage } from "../../../systemSlice";
 import { useAppDispatch } from "../../../store";
 import Tabs from "@mui/material/Tabs";
 import Tab from "@mui/material/Tab";
 import { TabPanel } from "../../shared/tabs";
-import { ClusterRegistered, FormTitle } from "./utils";
+import { ClusterRegistered, FormTitle, ProxyConfiguration } from "./utils";
```
prakashsvmx, line 64: We could just show the `Registered` once the cluster is registered or hide the form fields.  because even after registering, the offline registration form is shown as in the below screenshot

<details>

image

</details>

prakashsvmx (dismissed):
1. The buttons in the modal can be consistent with other buttons

image


2. Also the offline registration (form) needs an upgrade,  where we just need to show the command  to generate a SUBNET registration URL: `mc license register <alias>  <api_key_from_subnet>  --airgap`  @dvaldivia

reivaj05:
> 1. The buttons in the modal can be consistent with other buttons
> 
> image
> 
> 2. Also the offline registration (form) needs an upgrade,  where we just need to show the command  to generate a SUBNET registration URL: `mc license register <alias>  <api_key_from_subnet>  --airgap`  @dvaldivia

I'll take care of point 1 in this PR and i'll create a new PR for the second one

portal-ui/src/screens/Console/Support/Register.tsx
prakashsvmx: 👍

dvaldivia:
the auto register works fine, but I do get an invalid license key error
Screen Shot 2022-08-02 at 10 49 46 AM

dvaldivia (approved):

reivaj05:
> the auto register works fine, but I do get an invalid license key error Screen Shot 2022-08-02 at 10 49 46 AM

For me too but seems like subnet gringotts license is no longer active. Getting this error in my account 
Screen Shot 2022-08-02 at 12 16 55

pjuarezd:
In the "Online Activation" form, once the register failed after I inserted my 2FA token, I think would be preferable to return  to the Original Email & password form.

Screen Shot 2022-08-02 at 8 31 32 PM

Currently it only stays in the "verification token" form with no option to return.

reivaj05:
> In the "Online Activation" form, once the register failed after I inserted my 2FA token, I think would be preferable to return to the Original Email & password form.
> 
> Screen Shot 2022-08-02 at 8 31 32 PM
> 
> Currently it only stays in the "verification token" form with no option to return.

@pjuarezd I'll add a cancel button that will return to the original form, i don't want to automatically return after error because users might have just mistyped the verification code and do all the process.

Since this change is outside of the scope of this PR, I'll do it in a different PR

reivaj05:
@pjuarezd @prakashsvmx Could you please review one more time?

pjuarezd (approved):

prakashsvmx (approved):

PR #2241, opened 2022-08-10
Created button to download Widget data as CSV or PNG file

jinapurapu:
Screenshot from 2022-08-12 09-09-56
Screenshot from 2022-08-12 09-10-06
Screenshot from 2022-08-12 09-10-16

closes https://github.com/miniohq/engineering/issues/661
https://github.com/miniohq/engineering/issues/660


oscarocastellanos (commented):
Can we improve the alignment of the button? I'm leaving here an example of the desired alignment. It should rest in the same horizontal line as the box title and should have a margin to the right the same as the margin to the right of the box title. 
Captura de Pantalla 2022-08-10 a la(s) 16 39 52

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "@reduxjs/toolkit": "^1.8.2",
+    "react-component-export-image": "^1.0.6",
+    "react-router-dom": "6",
+    "testcafe": "^1.19.0"
+  }
+}
```
kaankabalak, line 8: Shouldn't all this be inside the portal-ui/package.json file instead of a new file?

kaankabalak, line 8: Instead of creating a new package.json file in the console directory, I think the following would be sufficient:

```diff
diff --git a/portal-ui/package.json b/portal-ui/package.json
index 2a1d9001..3c33f191 100644
--- a/portal-ui/package.json
+++ b/portal-ui/package.json
@@ -33,6 +33,7 @@
     "moment": "^2.29.4",
     "react": "^18.1.0",
     "react-chartjs-2": "^2.9.0",
+    "react-component-export-image": "^1.0.6",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^18.1.0",
     "react-dropzone": "^11.4.2",
```

kaankabalak, line 8: @jinapurapu Looks like we already have all the packages listed here in portal-ui/package.json, I don't think we would need this file at all

kaankabalak:
When I try to download the PNG for the Object Size Distribution, the download doesn't start and I get the following error on the browser dev console:

Screen Shot 2022-08-16 at 4 17 02 PM

portal-ui/src/screens/Console/Dashboard/DownloadWidgetDataButton.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const onDownloadError = (err: ErrorResponseHandler) =>
+    dispatch(setErrorSnackMessage(err));
+
+  const convertToCSV = (objectToConvert: any) => {
+    const array = [Object.keys(objectToConvert[0])].concat(objectToConvert);
+    return array
+      .map((it) => {
+        return Object.values(it).toString();
+      })
+      .join(",");
```
kaankabalak, line 66: Each item in the CSV file downloaded needs to be in its own row, currently we have everything in a single row as below:

Screen Shot 2022-08-16 at 4 27 03 PM

The logic here works great apart from only this `join` statement. We can fix this as follows as each item needs to be in its own row:

```suggestion
      .join("\n");
```

After applying the suggestion, our CSV file would look like the following, where every item is in its own row:

Screen Shot 2022-08-16 at 4 29 55 PM

portal-ui/src/screens/Console/Dashboard/Prometheus/Widgets/BarChartWidget.tsx
```diff
@@ -170,6 +197,7 @@ const BarChartWidget = ({
       )}
       {!loading && (
         <div
+          ref={thisRef as React.RefObject<HTMLDivElement>}
```
kaankabalak, line 200: When I try to download the PNG for the Object Size Distribution, the download doesn't start and I get the following error on the browser dev console:

Screen Shot 2022-08-16 at 4 17 02 PM

Looks like we can remove `thisRef` here and use only `componentRef` to fix this. I think we can fix this by applying the following diff:

```diff
diff --git a/portal-ui/src/screens/Console/Dashboard/Prometheus/Widgets/BarChartWidget.tsx b/portal-ui/src/screens/Console/Dashboard/Prometheus/Widgets/BarChartWidget.tsx
index a6c9182c..8b454930 100644
--- a/portal-ui/src/screens/Console/Dashboard/Prometheus/Widgets/BarChartWidget.tsx
+++ b/portal-ui/src/screens/Console/Dashboard/Prometheus/Widgets/BarChartWidget.tsx
@@ -96,9 +96,7 @@ const BarChartWidget = ({
   const [data, setData] = useState<any>([]);
   const [result, setResult] = useState<IDashboardPanel | null>(null);
   const [hover, setHover] = useState<boolean>(false);
-  const componentRef = useRef();
-
-  const thisRef = React.createRef();
+  const componentRef = useRef<HTMLElement>();
 
   const onHover = () => {
     setHover(true);
@@ -197,7 +195,7 @@ const BarChartWidget = ({
       )}
       {!loading && (
         <div
-          ref={thisRef as React.RefObject<HTMLDivElement>}
+          ref={componentRef as React.RefObject<HTMLDivElement>}
           className={
             zoomActivated ? classes.zoomChartCont : classes.contentContainer
           }
```

reivaj05 (approved):

kaankabalak (approved):

oscarocastellanos (approved):
LGTM

bexsoft (approved):
LGTM, In the line charts CSV the first column can be changed to be a date. this can be done in another PR

PR #2246, opened 2022-08-13
Added Color customization to embedded object browser

bexsoft:
## What does this do?

Implements custom color support for embedded object browser.

To test:
- Make assets
- Start browser in embedded mode, include one more query param called `ov_st` which will contain a base64 encoded string of the following JSON:
```JSON
{
  "backgroundColor": "#2787c6",
  "fontColor": "#ff0",
  "buttonStyles": {
    "backgroundColor": "#edeaa8",
    "textColor": "#2b2a2a",
    "hoverColor": "#edeaa8",
    "hoverText": "#2b2a2a",
    "activeColor": "#edeaa8",
    "activeText": "#2b2a2a"
  }
}
```

**NOTES:** Button colors may not be consistent, this will be handled during the replacement of buttons with mds system. this same case applied for application titles.

## How does it look?

Screen Shot 2022-08-12 at 23 35 10

Screen Shot 2022-08-12 at 23 35 03

Screen Shot 2022-08-12 at 23 34 56

### DIFF
--- a/portal-ui/src/utils/stylesUtils.ts
+++ b/portal-ui/src/utils/stylesUtils.ts
@@ -1,0 +1,152 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2022 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import { IEmbeddedCustomStyles } from "../common/types";
+import { createTheme } from "@mui/material";
+
+export const getOverrideColorVariants: (
+  customStyles: string
+) => false | IEmbeddedCustomStyles = (customStyles) => {
+  try {
+    return JSON.parse(atob(customStyles)) as IEmbeddedCustomStyles;
+  } catch (e) {
+    console.error("Error processing override styles, skipping.", e);
+    return false;
+  }
+};
+
+export const generateOverrideTheme = (overrideVars: IEmbeddedCustomStyles) => {
+  const theme = createTheme({
+    palette: {
+      primary: {
+        light: overrideVars.buttonStyles.hoverColor || "#073052",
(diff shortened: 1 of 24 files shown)


portal-ui/src/utils/stylesUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+    return false;
+  }
+};
+
+export const generateOverrideTheme = (overrideVars: IEmbeddedCustomStyles) => {
+  const theme = createTheme({
+    palette: {
+      primary: {
+        light: overrideVars.buttonStyles.hoverColor,
+        main: overrideVars.buttonStyles.backgroundColor,
```
kaankabalak, line 36: Hi @bexsoft, I think this line is causing the Embedded Console to display a blank screen when we omit as follows the `backgroundStyles` key for the `buttonStyles` object for the query parameter that is being encoded:

```json
{
  "backgroundColor": "#2787c6",
  "fontColor": "#ff0",
  "buttonStyles": {
    "textColor": "#2b2a2a",
    "hoverColor": "#edeaa8",
    "hoverText": "#2b2a2a",
    "activeColor": "#edeaa8",
    "activeText": "#2b2a2a"
  }
}
```

We might either want to revert to a default value if a value isn't provided for `overrideVars.buttonStyles.backgroundColor`, or include this parameter to the list of parameters that give us an error message in restapi/utils.go:166.

kaankabalak: Seems like we get the following error displayed on the browser dev console too:
```
MUI: The color (primary) provided to augmentColor(color) is invalid. color.main should be a string, but undefined was provided instead. Did you intend to use one of the following approaches? import { green } from "@mui/material/colors"; const theme1 = createTheme({ palette: { primary: green, } }); const theme2 = createTheme({ palette: { primary: { main: green[500] }, } });
```

bexsoft: Will add fallbacks for this

prakashsvmx: a question/request, could we read the styles,from a file if specified in some environment variable ?

bexsoft: The original idea is to change only colors for this use case, but we can work on it in a second PR so user can have both customization options

bexsoft: Fixed this issue

kaankabalak (approved):

prakashsvmx (approved):
Tested. Changes look good to me.
:+1: 

Note: we need to instruct that html color names like `red` , `blue` etc...  are not supported as mui does not support it. using it would lead to crash and blank screen. @bexsoft 
`The following formats are supported: #nnn, #nnnnnn, rgb(), rgba(), hsl(), hsla(), color().`


<details>

image

</details>

bexsoft:
> 

Will work on this in the ENV styles color support PR

PR #2259, opened 2022-08-23
read license from envt variable

sinhaashish:
Earlier the license were read from config , and only admin had the privilige to read those the config , so any normal user with read write permission were seeing AGPL license even when they had the license .

With this PR the license is being read from envt variable which is set in MinIO at initialization.

### DIFF
--- a/restapi/admin_subnet.go
+++ b/restapi/admin_subnet.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 
 	xhttp "github.com/minio/console/pkg/http"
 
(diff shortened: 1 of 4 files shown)


bexsoft (changes requested):
Please run `make prettify` in `portal_ui` folder

kannappanr (dismissed):

restapi/admin_subnet.go
```diff
@@ -307,22 +308,11 @@ func GetSubnetRegisterResponse(session *models.Principal, params subnetApi.Subne
...
-	}
-	if subnetTokens.APIKey == "" {
-		return nil, ErrorWithContext(ctx, ErrLicenseNotFound)
-	}
 	client := &xhttp.Client{
 		Client: GetConsoleHTTPClient(),
 	}
-	licenseInfo, err := subnet.ParseLicense(client, subnetTokens.License)
+
+	licenseInfo, err := subnet.ParseLicense(client, os.Getenv("CONSOLE_SUBNET_LICENSE"))
```
bexsoft, line 315: We are now returning an error when license is not present. Not sure if we want to return this error as it will be throwing unwanted error messages in register pages:
Screen Shot 2022-08-24 at 19 29 31

bexsoft (changes requested):

restapi/admin_subnet.go
sinhaashish: Thanks @bexsoft  .

prakashsvmx (approved):
LGTM :+1:

bexsoft (approved):
LGTM

PR #2262, opened 2022-08-23
Interactive Bucket Naming Rules component

jinapurapu:
Screenshot from 2022-08-25 13-32-09
Screenshot from 2022-08-25 13-32-22
Screenshot from 2022-08-25 13-32-32
Screenshot from 2022-08-25 13-33-12

 closes https://github.com/miniohq/engineering/issues/961

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/AddBucket.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/AddBucket.tsx
@@ -213,7 +213,6 @@ const AddBucket = ({ classes }: IsetProps) => {
                   )}
                   <br />
                   <br />
-                  
                 </Fragment>
               }
             />
@@ -407,7 +406,6 @@ const AddBucket = ({ classes }: IsetProps) => {
                       }
                     />
                   </Grid>
-                  
                 </React.Fragment>
               )}
             </Grid>
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/AddBucket.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/AddBucket.tsx
@@ -118,7 +118,9 @@ const AddBucket = ({ classes }: IsetProps) => {
     "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(.|$)){4}$"
   );
   const bucketName = useSelector((state: AppState) => state.addBucket.name);
-  const validationResult = useSelector((state: AppState) => state.addBucket.bucketNameErrorList);
+  const validationResult = useSelector(
+    (state: AppState) => state.addBucket.bucketNameErrorList
+  );
   const errorList = validationResult.filter((v) => !v);
   const hasErrors = errorList.length > 0;
   const [records, setRecords] = useState<string[]>([]);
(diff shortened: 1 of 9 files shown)


portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/BucketNamingRules.tsx
```diff
@@ -0,0 +1,203 @@
...
+    };
+    fetchRecords();
+  }, [dispatch]);
+
+  useEffect(() => {
+    if (validBucketCharacters.test(bucketName)) {
+      setValidCharacters(true);
+    } else {
+      setValidCharacters(false);
+    }
```
reivaj05, line 87: maybe setValidCharacters(validBucketCharacters.test(bucketName));

portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/BucketNamingRules.tsx
```diff
@@ -0,0 +1,203 @@
...
+    if (ipAddressFormat.test(bucketName)) {
+      setNotIPFormat(false);
+    } else {
+      setNotIPFormat(true);
+    }
```
reivaj05, line 92: setNotIPFormat(!ipAddressFormat.test(bucketName));

reivaj05 (dismissed):

oscarocastellanos (commented):
Can we keep the colors the same as they're used on other components?
I think blue should be #0288D1 while red should be #C83B51.

Is it possible to have greyed out bullets and they show the checkmark just once that specific rule has passed?

Thanks!

prakashsvmx:
@jinapurapu 
Observations:
- The bucket list api is called twice on load
image
- Also if we could refactor and highlight the bucket name field when at least one rule is not met, it would be a great user experience.( or we could expand the rules by default ). for example if a rule fails, the user has to expand and see 
@oscarocastellanos  any thoughts?

portal-ui/src/screens/Console/Buckets/ListBuckets/AddBucket/AddBucket.tsx
```diff
@@ -108,6 +113,17 @@ const AddBucket = ({ classes }: IsetProps) => {
   const dispatch = useAppDispatch();
   const navigate = useNavigate();
 
+  const validBucketCharacters = new RegExp(`^[a-z0-9.-]*$`);
+  const ipAddressFormat = new RegExp(
+    "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(.|$)){4}$"
+  );
+  const bucketName = useSelector((state: AppState) => state.addBucket.name);
+  const validationResult = useSelector(
```
prakashsvmx, line 121: this can be handled with a local state. do we really need a redux state.? as it is not shared with any other screens?
for example:

[reafactor_patch.txt](https://github.com/minio/console/files/9421670/reafactor_patch.txt)

prakashsvmx (approved):
Tested. Looks good to me. :+1:

oscarocastellanos:
> @jinapurapu Observations:
> 
> * Also if we could refactor and highlight the bucket name field when at least one rule is not met, it would be a great user experience.( or we could expand the rules by default ). for example if a rule fails, the user has to expand and see
>   @oscarocastellanos  any thoughts?
> * All  of unmet rules need to be highlighted for errors. e.g  `A-..`

Yeah, this would be great to implement! Thanks for the suggestion @prakashsvmx

reivaj05 (approved):

bexsoft (approved):
LGTM, works ok

PR #2266, opened 2022-08-24
Catch error in gofmt

cniackz:
Our test is not catching error in `gofmt` in version of go 1.19, look -d is not exiting with status of 1 hence we just print the difference but we don't notice this is actually failing:

https://github.com/golang/go/issues/46289

What I am doing in this PR is to actually catch the issue and fail the test if gofmt is having any difference.

### Dependencies:

* https://github.com/minio/console/pull/2278

### Related: 

* https://github.com/minio/console/pull/2283

### DIFF
--- a/verify-gofmt.sh
+++ b/verify-gofmt.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# Expanding gofmt to cover more areas.
+# This will include auto generated files created by Swagger.
+# Catching the difference due to https://github.com/golang/go/issues/46289
+DIFF=$(GO111MODULE=on gofmt -d .)
+if [[ -n $DIFF ]];
+then
+	echo "$DIFF";
+	echo "please run gofmt";
+	exit 1;
+fi
--- a/Makefile
+++ b/Makefile
@@ -28,10 +28,7 @@ verifiers: getdeps fmt lint
 
 fmt:
 	@echo "Running $@ check"
-	@GO111MODULE=on gofmt -d restapi/
-	@GO111MODULE=on gofmt -d pkg/
-	@GO111MODULE=on gofmt -d cmd/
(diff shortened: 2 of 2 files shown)


verify-gofmt.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+for VARIABLE in pkg cmd cluster restapi
+do
+	DIFF=$(GO111MODULE=on gofmt -d $VARIABLE/)
```
reivaj05, line 5: Do we want to run this in auto generated files  (e.g. files created by Swagger)?

reivaj05 (approved):

kaankabalak (approved):

bexsoft (approved):

verify-gofmt.sh
cniackz: @reivaj05 I would say yes as this sounds as a good practice for us to follow `gofmt` in all possible places. Let me try to expand this to all possible folders and see how it goes. Thanks for the feedback!

cniackz: @reivaj05 expanded to all folder 👍

reivaj05 (approved):

Alevsk (approved):

bexsoft:
Seems like we will have issues with gofmt on 1.19 every time we regenerate swagger files.

Not sure if should add an option during file regeneration to handle this.

cniackz:
Sure @bexsoft that sounds like a good idea to me, I can add that format after generation of swagger files. Let me spend some time to add this on a separate PR.

cniackz:
Hey @bexsoft what do you think about it?: https://github.com/minio/console/pull/2283

bexsoft (approved):
LGTM

PR #2276, opened 2022-08-26
To migrate site replication test

cniackz:
### Objective:

To migrate site replication test to datacenter

### Executed in our Datacenter:

Screen Shot 2022-09-09 at 1 31 54 PM

* In Github Runner, it takes ~4 min to be executed.
* In our Datacenter, it takes ~1 min to be executed.
* Saved time is ~3 min per test.


cniackz:
@dvaldivia please let me know if we can add runners with image from https://github.com/miniohq/engineering-tools/pull/24 and once this is done, let's talk about a security concern raised by Lenin to understand better this risk.

bexsoft (approved):

kaankabalak (approved):

cniackz:
The reason for not merging this PR is because our Runners does not have `docker.sock` available and hence tests are failing this way:

```
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

bexsoft (approved):

kaankabalak (approved):

PR #2291, opened 2022-09-01
WIP - Show Tiers status

jinapurapu:
Screenshot from 2022-09-01 10-29-25

closes https://github.com/miniohq/engineering/issues/681


reivaj05 (approved):

portal-ui/src/screens/Console/Configurations/TiersConfiguration/ListTiersConfiguration.tsx
```diff
@@ -105,16 +110,55 @@ const ListTiersConfiguration = ({ classes }: IListTiersConfig) => {
...
+            endpoint = tier.azure?.endpoint || "";
+            break;
+          default:
+            endpoint = "";
+        }
+        const xhr = new XMLHttpRequest();
+        xhr.open("HEAD", endpoint);
+        xhr.send();
+        xhr.onreadystatechange = (res: any) => {
+          if (res) {
```
kaankabalak, line 148: If we want to check if the request is successful, we can use the [readyState property of XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState) to make sure that the request is finished and the response is ready, and we can use the [status property](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status) to make sure that 200 OK is returned.

Therefore, we can use the following conditional instead:
```
if (xhr.readyState == 4 && xhr.status == 200)
```

portal-ui/src/screens/Console/Configurations/TiersConfiguration/ListTiersConfiguration.tsx
```diff
@@ -105,16 +110,57 @@ const ListTiersConfiguration = ({ classes }: IListTiersConfig) => {
...
+        xhr.send();
+        xhr.onreadystatechange = () => {
+          if (xhr.readyState === 4 || xhr.readyState === 2) {
+            tier.status = true;
+          } else {
+            tier.status = false;
+          }
+          xhr.onerror = () => {
+            tier.status = false;
+          };
```
kaankabalak, line 155: I think we can take this out of the expression for `xhr.onreadystatechange`. So, we can have something like the following:

```js
        xhr.onreadystatechange = () => {
          if (xhr.readyState === 4 || xhr.readyState === 2) {
            tier.status = true;
          } else {
            tier.status = false;
          }
        };
        xhr.onerror = () => {
          tier.status = false;
        };
```

kaankabalak (approved):

reivaj05 (approved):

PR #2306, opened 2022-09-09
do not add preceding '/' for putRequests

### LINKED ISSUE #2192: double slash when put objects from console
Hi, 

I got some issues when uploading files via console.
MINIO version: RELEASE.2022-06-17T02-00-35Z, refer to console version [v0.18.1](https://github.com/minio/minio/blob/RELEASE.2022-06-17T02-00-35Z/go.sum#L614)

====behaviour====

- condition:

```
upload file name: 1.jpg
under the root path of the bucket (i.e. the prefix should be empty string) named mytestbucet
```

- what I expected: 

as the prefix is empty, it could be skipped during POST API call on the console
```
POST https://myminio.com/console/api/v1/buckets/mytestbucet/objects/upload \
--data-raw $'------WebKitFormBoundary xxxx...xxx--\r\n'
```

- what I got:

instead of getting an empty prefix, I got a prefix with a leading shash ("/"), this makes the signature verification fail.

```
POST https://myminio.com/console/api/v1/buckets/mytestbucet/objects/upload?prefix=Lw== \
--data-raw $'------WebKitFormBoundary xxxx...xxx--\r\n'

```
with the message: 

```
code: 500
detailedMessage: "The request signature we calculated does not match the signature you provided. Check your key and signing method."
message: "an error occurred, please try again"
```

Attach some logs form ingress

```
172.18.0.1 - - [22/Jul/2022:07:18:06 +0000] "PUT /mytestbucet//1.jpg HTTP/1.1" 403 437 "-" "MinIO (linux; amd64) minio-go/v7.0.28" 74055 0.016 [minio-service-9000] [] 10.244.0.67:9000 437 0.016 403 79e291a8a2e03c2315bae318abe9c92f
```

====Why this is out of expectation====

when uploading the file to the default path of a bucket, the prefix should be an empty string, not a leading slash shown here.
Having a leading slash could make no difference on the server side (in terms of delivering files) but I got an invalid signature message due to this. I believe the double slash makes the signature wrong.

### LINKED ISSUE #2305: Upload impossible in bucket root
Hi.

Since we upgraded our distributed minio production setup this summer, we can't upload files (`s3.PutObject`) in buckets root or newly created folders in bucket roots using the minio embedded web console.
It is working with Python SDK, GUI and CLI clients but with the console, we have upload errors. It was working fine before the update to v2022-08-22 and v2022-09-07.
It seems the console is trying to make a PUT request like `PUT /bucket-name//file.txt` instead of `PUT /bucket-name/file.txt` . As a result our reverse-proxy automatically fix it (as `PUT /bucket-name/file.txt`) before sending the request to the backends, even before the `RewriteRule ... [P]` rule, which leads to a SignatureDoesNotMatch error from the backend.
Is it a bug in console code? And if anyone has an idea on how to configure Apache not to fix PUT requests, it would be greatly appreciated.

harshavardhana:
Bonus: add user-agent for Console UI requests

fixes #2305
fixes #2192

### DIFF
--- a/restapi/client-admin.go
+++ b/restapi/client-admin.go
@@ -37,7 +37,7 @@ const globalAppName = "MinIO Console"
 
 // NewAdminClientWithInsecure gives a new madmin client interface either secure or insecure based on parameter
 func NewAdminClientWithInsecure(url, accessKey, secretKey, sessionToken string, insecure bool) (*madmin.AdminClient, *probe.Error) {
-	s3Client, err := s3AdminNew(&mcCmd.Config{
+	admClient, err := s3AdminNew(&mcCmd.Config{
 		HostURL:      url,
 		AccessKey:    accessKey,
 		SecretKey:    secretKey,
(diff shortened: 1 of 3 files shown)


bexsoft (approved):
LGTM

reivaj05 (approved):

PR #2303, opened 2022-09-09
Replacing all buttons from console with mds

bexsoft:
## What does this do?

Replaces all buttons in Console, Operator Console & DirectPV Console to use mds standard buttons

## How does it look?
Screen Shot 2022-09-09 at 14 38 07
Screen Shot 2022-09-09 at 14 38 02
Screen Shot 2022-09-09 at 14 37 55
Screen Shot 2022-09-09 at 14 37 42
Screen Shot 2022-09-09 at 14 37 37
Screen Shot 2022-09-09 at 14 30 28
Screen Shot 2022-09-09 at 14 29 28
Screen Shot 2022-09-09 at 14 29 16

PR #2307, opened 2022-09-09
Inherits error props to tooltip wrapper child buttons

bexsoft:
requires #2303 

## What does this do?

Inherits error props to child buttons when tooltip wrapper is set


prakashsvmx (approved):
LGTM :+1:

dvaldivia (approved):

PR #2308, opened 2022-09-10
fix: login function was returning fixed error

Alevsk:
On login failure console is returning a fixed error message instead of the real error, this makes debugging extremely difficult

After the fix:

Screenshot from 2022-09-10 16-03-33
Screenshot from 2022-09-10 16-04-01


dvaldivia (approved):

prakashsvmx (approved):
LGTM :+1:

PR #2312, opened 2022-09-14
Return 401 for Login Errors

dvaldivia:


restapi/errors.go
```diff
@@ -95,6 +99,11 @@ func ErrorWithContext(ctx context.Context, err ...interface{}) *models.Error {
 				errorCode = 401
 				errorMessage = ErrInvalidLogin.Error()
 			}
+			// If the last error is ErrInvalidLogin, this is a login failure
+			if errors.Is(lastError, ErrInvalidLogin) {
+				errorCode = 401
+				errorMessage = err1.Error()
+			}
```
bexsoft, line 106: Not sure if this is correct as there was an issue some time ago with login responses and a change from `minio-go`

dvaldivia: the thing is @Alevsk reverted my change to return 401 on login errors on https://github.com/minio/console/pull/2308 because I was causing all errors to be the same, but if we are passing multiple errors, we are not doing much with the rest of the errors, thus checking always the last error on the list

bexsoft (dismissed):
LGTM, we need to rethink errors system

restapi/errors.go
```diff
@@ -81,6 +81,10 @@ func ErrorWithContext(ctx context.Context, err ...interface{}) *models.Error {
 	var exists bool
 	if len(err) > 0 {
 		if err1, exists = err[0].(error); exists {
+			var lastError error
+			if len(err) > 1 {
+				lastError = err[len(err)-1].(error)
```
harshavardhana, line 86: can this crash? if for some reason `err[len(err)-1]` is `nil` ?

This code however looks very complex and not clean - this must be fixed.

harshavardhana (dismissed):

restapi/errors_test.go
```diff
@@ -96,11 +96,11 @@ func TestError(t *testing.T) {
 	}
 	tests = append(tests,
 		testError{
-			name: "passing multiple errors",
+			name: "passing multiple errors but ErrInvalidLogin is last",
```
dilverse, line 99: does this message propagate to the enduser if so should we make this user friendly?  :)

dvaldivia: yeah, that is a concern, we are going to need to change the whole error propagation function

restapi/errors.go
dvaldivia: fixed @harshavardhana

bexsoft (approved):
LGTM

bexsoft (approved):

dilverse (approved):

Alevsk (changes requested):
We need to rethink this because currently, im not sure if the error code would be 401 for this particular case, minio is not even reachable

image

Alevsk (approved):

PR #2313, opened 2022-09-15
Limit concurrent downloads & uploads

### LINKED ISSUE #2163: Limit concurrent files upload
Hi,

It seems that the number files that can be uploaded concurrently via the console is not limited.
For example, if I drag/drop 1000 files to the console, 1000 POST requests will be created at the same time.

If MinIO server concurrent connections are limited, via [MINIO_API_REQUESTS_MAX](https://github.com/minio/minio/blob/master/docs/throttle/README.md#configuring-connection-limit) or via a reverse proxy server (typically Nginx), some of the file uploads will fail.

I haven't found a way to limit this in the console.

I think passing an environment variable to the console, like `CONSOLE_MAX_SIMULTANEOUS_UPLOADS`, could be interesting.

For example, if we set `CONSOLE_MAX_SIMULTANEOUS_UPLOADS` to 100 and try to upload 1000 files, the first 100 will be uploaded and the 900 others will wait in queue.

What do you think about that?

bexsoft:
fixes #2163 

## What does this do?

Limits concurrent downloads & uploads in console Object Manager

To set variables, you need to start console server with the following flags: `CONSOLE_MAX_CONCURRENT_UPLOADS` 
 & `CONSOLE_MAX_CONCURRENT_DOWNLOADS`

if values are set to 0, then no limit is set


portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ObjectDetailPanel.tsx
```diff
@@ -344,7 +347,7 @@ const ObjectDetailPanel = ({
       })
     );
 
-    downloadCall.send();
+    //downloadCall.send();
```
prakashsvmx, line 350: could we remove the comment?

portal-ui/src/screens/Console/Common/ObjectManager/ObjectManager.tsx
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import React, { Fragment } from "react";
+import React, { Fragment, useEffect } from "react";
```
prakashsvmx, line 17: `useEffect` is unused. so it could be removed

portal-ui/src/screens/Console/ObjectBrowser/RenameLongFilename.tsx
```diff
@@ -126,7 +129,7 @@ const RenameLongFileName = ({
       })
     );
 
-    downloadCall.send();
+    //downloadCall.send();
```
prakashsvmx, line 132: commented line could be removed

swagger-console.yml
```diff
@@ -3692,6 +3692,8 @@ definitions:
         type: array
         items:
           type: string
+      isDirectPV:
```
prakashsvmx, line 3695: do we need this flag?

prakashsvmx:
Observation:
set a limitation of `CONSOLE_MAX_CONCURRENT_UPLOADS=4`
Drag and drop a set of files and the crash can be observed.

```
TrafficMonitor.tsx:88 Uncaught TypeError: Cannot read properties of undefined (reading 'send')
    at TrafficMonitor.tsx:88:1
 ```

File: swagger-console.yml (hunk 4 above)
Comment (bexsoft) [reply swagger-console.yml:3695]: Yes, it is a flag that was deleted accidentally

File: portal-ui/src/screens/Console/ConsoleKBar.tsx [hunk 5]
```diff
@@ -19,6 +19,7 @@ import Console from "./Console";
 import { useSelector } from "react-redux";
 import CommandBar from "./CommandBar";
 import { selFeatures } from "./consoleSlice";
+import TrafficMonitor from "./Common/ObjectManager/TrafficMonitor";
```
prakashsvmx, line 22: `TrafficMonitor` is unused. could be removed.

portal-ui/src/screens/Console/Console.tsx
```diff
@@ -59,6 +59,7 @@ import {
   setSnackBarMessage,
 } from "../../systemSlice";
 import { selFeatures, selSession } from "./consoleSlice";
+import TrafficMonitor from "./Common/ObjectManager/TrafficMonitor";
```
prakashsvmx, line 62: TrafficMonitor is unused. could be removed.

bexsoft:
> Observation: set a limitation of `CONSOLE_MAX_CONCURRENT_UPLOADS=4` Drag and drop a set of files and the crash can be observed. ( 4 files are uploaded which can be seen after refresh)
> 
> ```
> TrafficMonitor.tsx:88 Uncaught TypeError: Cannot read properties of undefined (reading 'send')
>     at TrafficMonitor.tsx:88:1
> ```
> 
> * could we show a warning/error when there are more files than the limit?

fixed

dvaldivia (dismissed):
For `CONSOLE_MAX_CONCURRENT_UPLOADS=2` I see sets of 2 but others sets of uploads are let through, at some point the desired effects happens

Screen Shot 2022-09-20 at 3 53 33 PM

bexsoft:
> For `CONSOLE_MAX_CONCURRENT_UPLOADS=2` I see sets of 2 but others sets of uploads are let through, at some point the desired effects happens
> 
> Screen Shot 2022-09-20 at 3 53 33 PM

Fixed

prakashsvmx (approved):
Tested. Changes look good to me :+1:

reivaj05 (approved):

dvaldivia (approved):

PR #2319, opened 2022-09-16
Added permissions guidance on ListBuckets and AddBuckets screens

jinapurapu:
If a feature is disabled due to permissions, a tooltip will advise the user which permission is needed to use the feature.
Screen Shot 2022-09-15 at 4 47 02 PM
Screen Shot 2022-09-15 at 4 45 48 PM
Screen Shot 2022-09-15 at 4 44 41 PM


dvaldivia (dismissed):

portal-ui/src/screens/Console/Buckets/BucketDetails/BrowserHandler.tsx
```diff
@@ -88,6 +89,8 @@ const BrowserHandler = () => {
     navigate(`/buckets/${bucketName}/admin`);
   };
 
+  const configureBucketAllowed = hasPermission("*", [IAM_ROLES.BUCKET_ADMIN]);
```
bexsoft, line 92: In this specific case I think '*' should be `bucketName` as there may be cases where configuration is not set for a specific bucket. @Alevsk Can you please confirm?

dvaldivia (approved):

portal-ui/src/screens/Console/Buckets/ListBuckets/ListBuckets.tsx
```diff
@@ -384,6 +393,13 @@ const ListBuckets = ({ classes }: IListBucketsProps) => {
                         MinIO uses buckets to organize objects. A bucket is
                         similar to a folder or directory in a filesystem, where
                         each bucket can hold an arbitrary number of objects.
+                        <br></br>
+                        <br></br>
+                        {canListBuckets
+                          ? ""
+                          : "In order to view the buckets on this server, you require " +
```
dilverse, line 400: Should we be consistent with the messaging? since in other places we use `You require additional permissions in order to....`

dilverse (approved):
Message consistency can be addressed in separate PR :)

PR #2323, opened 2022-09-20
Skip TLS verif for local address

vadmeste:
If Console is talking locally to MinIO, skip the TLS verification if any.

This will allow users to avoid defining the correct MINIO_SERVER_URL domain
address, when TLS is enabled, useful in a bare-metal setup.


restapi/client-admin.go
```diff
@@ -477,17 +480,48 @@ func newAdminFromCreds(accessKey, secretKey, endpoint string, tlsEnabled bool) (
...
+	if h, _, err := net.SplitHostPort(hostName); err == nil {
+		if ips, err := net.LookupIP(h); err == nil {
+			for _, ip := range ips {
+				if ip.IsLoopback() {
+					isLocalAddress = true
+				}
+			}
+		}
+	}
+
```
harshavardhana, line 519: Please simply to always use 127.0.0.1 instead

harshavardhana (changes requested):

restapi/client-admin.go
```diff
@@ -477,17 +480,48 @@ func newAdminFromCreds(accessKey, secretKey, endpoint string, tlsEnabled bool) (
...
+	client = PrepareConsoleHTTPClient(isLocalAddress)
```
Alevsk, line 520: Based on the internal conversation we cannot trust on any dns/nslookup resolution and instead implement this logic only if the provided address is the 127.0.0.1  string, we may thing about the corner case of the ipv6 local address as well

Alevsk (changes requested):

restapi/client-admin.go
vadmeste: > Based on the internal conversation we cannot trust on any dns/nslookup resolution and instead implement this logic only if the provided address is the 127.0.0.1 string, we may thing about the corner case of the ipv6 local address as well

Yes, we cannot trust DNS resolution but I think we can trust ip.IsLoopback(), I like loopback because it also supports ipv6 localhost (and I hate parsing ipv6)

restapi/client-admin.go
vadmeste: > Please simply to always use 127.0.0.1 instead

127.0.0.1 won't always work I guess, this looks a better way to support both ipv4 and ipv6 localhost

harshavardhana: 127.0.0.1 will always work @vadmeste it will work even with IPv6 is enabled.

restapi/client.go
```diff
@@ -438,21 +443,26 @@ func pathJoinFinalSlash(elem ...string) string {
...
-	s3Config.Insecure = insecure
 
 	s3Config.HostURL = endpoint
 	s3Config.AccessKey = accessKey
 	s3Config.SecretKey = secretKey
 	s3Config.SessionToken = sessionToken
 	s3Config.Signature = "S3v4"
+
+	insecure := false
+	if u, err := url.ParseRequestURI(endpoint); err == nil && u.Host != "" {
```
harshavardhana, line 462: This is still url.Parse()

harshavardhana: This should be converted into a function where it says `isLocalURL(endpoint)` and used everywhere.

harshavardhana (approved):

harshavardhana:
PTAL at the failed tests @vadmeste

restapi/client-admin.go
```diff
@@ -487,6 +487,13 @@ var httpClients = struct {
 	m: make(map[string]*http.Client),
 }
 
+func isLocalAddress(address string) bool {
+	if u, err := url.Parse(address); err == nil && u.Host != "" {
```
harshavardhana, line 491: use `u.Hostname()`  so that you can remove the port and then 

```
ip := net.ParseIP(u.Hostname())
return ip.IsLoopback()
```

harshavardhana (changes requested):

restapi/client-admin.go
vadmeste: yes, I like loopback better as well, done

restapi/client-admin.go
```diff
@@ -477,17 +480,54 @@ func newAdminFromCreds(accessKey, secretKey, endpoint string, tlsEnabled bool) (
...
+
+// isLocalAddress returns true if the url contains an IPv4/IPv6 hostname
+// that points to the local machine - FQDN are not supported
+func isLocalIPAddress(ipAddr string) bool {
+	if ipAddr == "" {
+		return false
 	}
-	return httpClient
+	ip := net.ParseIP(ipAddr)
+	return ip.IsLoopback()
```
harshavardhana, line 507: Check for ip to be non-nil other wise this function can crash.

harshavardhana (changes requested):

restapi/client-admin.go
harshavardhana: @vadmeste ^^

vadmeste: > Check for ip to be non-nil other wise this function can crash.

Fixed, good catch

harshavardhana (approved):

Alevsk (approved):
Tested, works fine, this verification will only apply if the provided IP is `127.0.0.1`

harshavardhana:
> Tested, works fine, this verification will only apply if the provided IP is `127.0.0.1`

Yes this must be done via server setting as well. MinIO must set this to local IP always for embedded console. 

// cc @vadmeste can you make this server change as well?

PR #2331, opened 2022-09-21
Add support for edit/add/remove environment variables to MinIO tenant

Alevsk:
image


Alevsk:
@harshavardhana @aead @dvaldivia guys I need to get your opinion on:

- Should we allow people to edit secrets/passwords/auth token environment variables from this UI or those more sensitive values should be have to be manually edited via `kubectl`
- Right now we are returning all the content of the tenant configuration file to the user, including sensitive data such as `MINIO_ROOT_PASSWORD`, etc, technically any user that logs to operator with a valid `k8s token` have access to this data already, would it be worth it to return `REDACTED` values instead or something like that just to prevent this credentials in plain text to reach the client? again the user already have access to all of this via the k8s token

portal-ui/src/screens/Console/Tenants/TenantDetails/TenantConfiguration.tsx
```diff
@@ -0,0 +1,320 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2021 MinIO, Inc.
```
jinapurapu, line 2:
```suggestion
// Copyright (c) 2022 MinIO, Inc.
```

jinapurapu (changes requested):

operatorapi/tenants.go
```diff
@@ -972,6 +990,110 @@ func getSetTenantAdministratorsResponse(session *models.Principal, params operat
 	return nil
 }
 
+func getTenantConfigurationResponse(session *models.Principal, params operator_api.TenantConfigurationParams) (*models.TenantConfigurationResponse, *models.Error) {
```
reivaj05, line 993: It needs tests for the new code

Alevsk: added tests

reivaj05 (approved):

jinapurapu (changes requested):
After confirming save and restart, Save button is disabled but you can add/delete fields using the +/- buttons, this may be confusing to users.
Screen Shot 2022-09-27 at 10 59 53 AM
Screen Shot 2022-09-27 at 11 00 52 AM

Alevsk:
> After confirming save and restart, Save button is disabled but you can add/delete fields using the +/- buttons, this may be confusing to users. Screen Shot 2022-09-27 at 10 59 53 AM Screen Shot 2022-09-27 at 11 00 52 AM

fixed, thanks!!

jinapurapu (changes requested):
It is possible to save an empty KV pair, or an individual key or value. 

Screen Shot 2022-09-28 at 10 44 37 AM

Alevsk:
> It is possible to save an empty KV pair, or an individual key or value.
> 
> Screen Shot 2022-09-28 at 10 44 37 AM

You should be able to save keys with empty values (in case you want to reset something to the default value), on the other hand you should not be able to save empty keys and values, will check if this is happening because i have some logic around to prevent that

jinapurapu (changes requested):
Operator API test is failing here locally as well
```
--- FAIL: Test_TenantInfoTenantAdminClient (0.00s)
--- FAIL: Test_TenantInfoTenantAdminClient/Return_Tenant_Admin,_no_errors (0.00s)

```

reivaj05 (dismissed):

jinapurapu (approved):

bexsoft (approved):

PR #2333, opened 2022-09-22
Fix oauth state generation for OIDC login

donatello:
This is a regression from 118cf97e1d2313fe67839210706c45fcb5ed4e1d when env var support for passing console configuration from MinIO was removed.

This change ensures that all MinIO nodes in a cluster are able to verify state tokens generated by other nodes in the cluster. Without this, it is necessary to use sticky sessions in a loadbalancer to ensure that OIDC authorization code login flow steps for a client happens on the same minio node.

Fixes https://github.com/minio/minio/issues/15527


harshavardhana (dismissed):

donatello:
Remaining CI failure doesnt seem related to the change

harshavardhana (approved):

Alevsk (approved):

PR #2339, opened 2022-09-24
Add filtering of configuration parameters

dormanze:
fix https://github.com/minio/minio/issues/15736
Filter out \n ,prevent the error of minio server interface caused by foreground input.

before:
image
after:
image


harshavardhana:
@dormanze it is better to not replace \n with empty character this PR fixes couple of other issues as well https://github.com/minio/console/pull/2342

dormanze:
> @dormanze it is better to not replace \n with empty character this PR fixes couple of other issues as well #2342

Yes, I agree with you,I will close my pr.

PR #2350, opened 2022-09-30
Site replication support for different accesskey and secretkey in different sites

prakashsvmx:
How does it look:

Credentials can be different in the peer sties participating in Site Replication. So add input fields to collect and configure from API.
- AccessKey and SecretKey are required for every site whenever a new site is added.
- Current site endpoint is populated automatically and not editable
- Saved ( participating) sites are listed with name and endpoint as read only

image


dvaldivia (dismissed):
Tested. LGTM

bexsoft (changes requested):
When trying to edit replication sites I think we must specify that we need to re-enter credentials to all replication sites:

Screen Shot 2022-10-07 at 12 22 01

bexsoft:
Please disable save button if all fields are not completed:

Screen Shot 2022-10-07 at 12 24 19

bexsoft (changes requested):
There is an issue trying to create a new replication set, for `This Site` endpoint field is readonly and empty:

Screen Shot 2022-10-11 at 15 17 54
Screen Shot 2022-10-11 at 15 18 15

prakashsvmx:
@bexsoft  could you please confirm if `make install` was done?
there are back-end changes  done to get the environment variable to populate the end point of current site .
it would be of great help to share steps to replicate as well.

bexsoft (approved):
LGTM

jinapurapu (approved):

PR #2352, opened 2022-10-03
update madmin-go and mc deps for new changes

harshavardhana:


harshavardhana:
SSO integration test will fail due to madmin-go breakage and since this depends on server and server doesn't have the right dependencies yet.

harshavardhana:
Need this merged for batch replication PR and 1 more dependent PR.

PR #2355, opened 2022-10-04
do not log random errors using Go logger

harshavardhana:
we need to make sure that we print in consistent
format for the logs, remove the unnecessary logs
everywhere used via `log.Print*`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,24 +14,19 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-    - typecheck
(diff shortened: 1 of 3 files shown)


prakashsvmx:
@harshavardhana some lint errors .PTAL

harshavardhana:
> @harshavardhana some lint errors .PTAL

fixed @prakashsvmx

prakashsvmx (approved):

reivaj05 (approved):

PR #2365, opened 2022-10-06
Adding bucket events test

adfost:
Test case for adding bucket events


integration/buckets_events_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_AddEventAPI(t *testing.T) {
+	assert := assert.New(t)
+
+	AddBucket("testevents5", false, false, map[string]interface{}{}, map[string]interface{}{})
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "Hello, %q", html.EscapeString(r.URL.Path))
```
bexsoft, line 39: Please remove this line

integration/buckets_events_test.go
```diff
@@ -0,0 +1,128 @@
...
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			log.Printf("Error reading body: %v", err)
+			http.Error(w, "can't read body", http.StatusBadRequest)
+		} else {
+			log.Printf(string(body))
```
bexsoft, line 45: Do we need to only print the body value?

bexsoft (changes requested):

bexsoft:
Please squash commits

integration/buckets_events_test.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	})
+
+	go http.ListenAndServe(":8081", nil)
```
dvaldivia, line 49: server should be stopped at the end of the test, else we are going to have a goroutine leak

integration/buckets_events_test.go
bexsoft: On this line you are not using `body` anywhere else, Not sure if this block is really needed

integration/buckets_events_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+			requestDataEvents := map[string]interface{}{}
+			requestDataEvents["conf"] = map[string]interface{}{}
+			requestDataEvents["conf"].(map[string]interface{})["ignoreExisting"] = tt.args.conf.ignoreExisting
+			requestDataEvents["conf"].(map[string]interface{})["configuration"] = map[string]interface{}{}
+			requestDataEvents["conf"].(map[string]interface{})["configuration"].(map[string]interface{})["id"] = tt.args.conf.configuration.id
+			requestDataEvents["conf"].(map[string]interface{})["configuration"].(map[string]interface{})["arn"] = tt.args.conf.configuration.arn
+			requestDataEvents["conf"].(map[string]interface{})["configuration"].(map[string]interface{})["notificationEventType"] = tt.args.conf.configuration.notificationEventType
+			requestDataEvents["conf"].(map[string]interface{})["configuration"].(map[string]interface{})["prefix"] = tt.args.conf.configuration.prefix
+			requestDataEvents["conf"].(map[string]interface{})["configuration"].(map[string]interface{})["suffix"] = tt.args.conf.configuration.suffix
```
reivaj05, line 106: It would be better to define a new struct for this config and then marshal it.This way is really difficult to read

integration/buckets_events_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+			requestDataJSON, _ := json.Marshal(requestDataEvents)
+			requestDataBody := bytes.NewReader(requestDataJSON)
+			request, err := http.NewRequest(
+				"POST", "http://localhost:8081/api/v1/buckets/testevents5/events", requestDataBody)
+			if err != nil {
+				log.Println(err)
+				return
```
reivaj05, line 114: `err != nil` should make the test fail? this way the test will pass

integration/buckets_events_test.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+			request.Header.Add("Cookie", fmt.Sprintf("token=%s", token))
+			request.Header.Add("Content-Type", "application/json")
+			response, err := client.Do(request)
+			if err != nil {
+				log.Println(err)
+				return
+			}
```
reivaj05, line 122: same as in line 114

PR #2366, opened 2022-10-06
Splitting Operator UI Tests for faster execution

cniackz:
### Objective:

To make `Operator UI Tests` run faster and more reliable.

### Story:

Currently, there are two main tests:

1. `login.ts`
2. `tenants.ts`

They depend on different libraries, which is ok, not the main point to change here.
But I am going to group them in different folders, letting main `uitls.ts` be on top of them in case other tests are added in the future.

### Additional Information:

This is going to make it just a bit faster. What is actually consuming lot of time here is the setup of the cluster with Operator and Tenant; not testcafe itself. But this is just the first PR. In my next PR, I will focus on the k8s side to see if we can actually decrease the time we consume in creating the cluster.

For reliability, I noticed that some marks render slower than the expected result, added some waits around them to let page to render prior checking on them, hopefully this will increase the reliability of our tests.

### Re-org:

* Isolating those tests that are more prone to failure when running together with other tests.

Screen Shot 2022-10-07 at 12 07 13 PM

### Results:

Screen Shot 2022-10-07 at 2 03 55 PM


jinapurapu (approved):

bexsoft (approved):
LGTM

bexsoft (approved):
LGTM

PR #2369, opened 2022-10-06
Set timeout in all permission tests

cniackz:
### Objective:

Kill permission tests if they exceed 10 minutes threshold.

### Additional information:

5 minutes is not enough for these 3 tests. At least 10 minutes is required.


bexsoft (approved):
LGTM

harshavardhana (approved):

PR #2371, opened 2022-10-07
Mute permissions tooltips

jinapurapu:
Screen Shot 2022-10-14 at 6 30 45 PM
Screen Shot 2022-10-14 at 6 31 30 PM


cniackz:
Tested locally, I can toggle to enable or disable:
toggle

cniackz (approved):
LGTM

Alevsk:
can you squash all the changes into a single commit @jinapurapu ? thanks.

reivaj05 (approved):

Alevsk:
[Screencast from 10-07-2022 02:59:01 PM.webm](https://user-images.githubusercontent.com/1795553/194668730-d4c92528-6d14-4d30-b9e8-ee977e542344.webm)

This is more like a design decision but should we automatically close the tooltip once the mouse leaves the button? looks like right now the tooltip is closed if you over the tooltip itself, if you dont over it the tooltip will stay there until you change to a different page

prakashsvmx:
@jinapurapu  could you explore

`enterDelay` and `leaveDelay` props for the tooltip to address @Alevsk  comments?

kaankabalak:
@jinapurapu The mute option seems to be listed under the Service Account page description, but seems to affect permissions in all other pages as well. Is this intentional?

If not, should we place this option somewhere where it's visible regardless of which page the user is on the app?

oscarocastellanos:
> @dvaldivia @jinapurapu if the user has all the permissions, the tooltip enable/disable is confusing as the result is none.
> 
> the toolips are overlapping if hovered quickly. Could we think of some ux which could be part of the help box itself which could be more contextual help? @oscarocastellanos

Will work on something @prakashsvmx 👍

oscarocastellanos:
What I'm getting from these videos and screenshots is that tooltips should behave as any regular tooltip, this is, it should just show one tooltip at a time, only on mouseover for the button, and it should immediately dissapear once the mouse goes to a different element. If we make the interaction instant, it shouldn't overlay tooltips. 

@prakashsvmx @jinapurapu

reivaj05 (approved):

reivaj05 (approved):

PR #2382, opened 2022-10-13
Add Role ARN as part of OIDC provider

donatello:
- RoleARN needs to be used in the STS API call when present. Code to use this value needs to be added.


donatello:
The CI failure is not related to this change.

bexsoft (approved):

harshavardhana (approved):

PR #2383, opened 2022-10-13
ListUsers screen Permission guidance SnackMessage

jinapurapu:
Screen Shot 2022-10-13 at 4 51 02 PM


kaankabalak:
@jinapurapu, did we decide to move away from tooltips? Seems like the message is displayed only when the disabled buttons are clicked.

A potential problem here would be that the users might not even click on these buttons as they are disabled, therefore I think tooltips on button hover would communicate the situation better here.

oscarocastellanos (commented):
Looks good, but  I agree with @kaankabalak about the functionality and I think, what he proposes it's the best option moving forward.

jinapurapu:
@kaankabalak  @oscarocastellanos I've changed this to trigger on hover for the disabled buttons, maintaining the snack message format for readability. We can stick with the tooltips if that is preferred though

PR #2396, opened 2022-10-18
Allow policy with "s3:*Object" to download using Console  

jinapurapu:


portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
```diff
@@ -1236,7 +1236,10 @@ const ListObjects = () => {
     uploadPath = uploadPath.concat(currentPath);
   }
 
-  const canDownload = hasPermission(bucketName, [IAM_SCOPES.S3_GET_OBJECT]);
+  const canDownload = hasPermission(bucketName, [
+    IAM_SCOPES.S3_GET_OBJECT,
+    IAM_SCOPES.S3_STAR_OBJECT,
```
ravindk89, line 1241: Q: this is basically a string match right?

Do we need to do a larger pass for other areas in which we should be interpreting `s3:*Object` as a wildcard match on anything with `Object` as a suffix?

ravindk89 (approved):

prakashsvmx (approved):
Changes look good to me.
Note:
Users might expect wildcard support for other group of actions. As per s3.
>You can also use wildcards (*) as part of the action name.
(https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_action.html)

PR #2424, opened 2022-11-02
KMS backend definition needs to use 'keystore' key name in KES yaml

vadmeste:
'keys' meaning in KES server-config.yaml was changed two years ago. 
Currently 'keystore' is used to define a KMS backend.

'root' was renamed to 'admin' as well 13 months ago.


vadmeste:
This is on hold because it can break if one user upgrades Console but not KES server version is still < 0.17.2

It is also not very easy to check for KES version right now, let's disucss this

Alevsk:
`KES` know exposes the https://pkg.go.dev/github.com/minio/kes/keserv api, we can drop the custom kes config struct we have in console and instead of the one from `KES`

dvaldivia:
there's some APIs in KES that could allow us to check the version @vadmeste

PR #2430, opened 2022-11-06
Removing slash at the end of GCS tier screen endpoint

thiago4int:
https://storage.googleapis.com/ will generate https://storage.googleapis.com//bucket-name and tier creation will fail


bexsoft (approved):
LGTM

reivaj05 (approved):

PR #2434, opened 2022-11-07
Increase STS Session duration to 12 hours

dvaldivia:

### DIFF
--- a/pkg/auth/token/config.go
+++ b/pkg/auth/token/config.go
@@ -23,19 +23,11 @@ import (
 	"github.com/minio/pkg/env"
 )
 
-// GetConsoleSTSDuration returns the default session duration for the STS requested tokens (defaults to 1h)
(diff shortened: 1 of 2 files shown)


pjuarezd (approved):

dilverse (approved):

harshavardhana:
So user will be logged in until 12 hours, note this doesn't change the behavior of the JWT (oauth2) you can't change the duration here beyond the JWT expiration.

pjuarezd:
Yes @harshavardhana is to keep the console session live during a longer period, right now it is too short (1 hour) and often we are being kicked off while still using Console

> So user will be logged in until 12 hours, note this doesn't change the behavior of the JWT (oauth2) you can't change the duration here beyond the JWT expiration.

If the authorizer (oauth2 or MinIO) have a shorter period than 12h, that is OK, we'll take it.

There is plans to have a timer and end the session over a certain time of inactivity to compliment the longer sts session duration risks.

bexsoft (approved):

PR #2437, opened 2022-11-07
Add deprecated message in Audit logs section

reivaj05:
Screen Shot 2022-11-08 at 11 28 03


portal-ui/src/screens/Console/Tenants/TenantDetails/TenantAuditLogsScreen.tsx
```diff
@@ -243,7 +246,37 @@ const LoggingScreen = ({ classes }: ILoggingScreenProps) => {
...
+              Current Audit logs functionality will be deprecated soon, please
+              refer to the
+              <a
+                href="https://min.io/docs/minio/kubernetes/upstream/operations/monitoring/minio-logging.html"
+                target="_blank"
+                rel="noreferrer"
+              >
+                {" documentation "}
+              </a>
+              in order to set an external service for logs
```
dilverse, line 271: `in order to setup an external service for logs` would be better than `in order to set an external service for logs` IMO

prakashsvmx (dismissed):

kaankabalak (approved):

dilverse (approved):

.github/workflows/jobs.yaml
```diff
@@ -1695,193 +1695,193 @@ jobs:
...
-      # Runs a set of commands using the runners shell
-      - name: Start Kind for Operator UI
-        run: |
-          "${GITHUB_WORKSPACE}/portal-ui/tests/scripts/operator.sh"
-
-      - name: Run TestCafe Tests
-        uses: DevExpress/testcafe-action@latest
-        with:
-          args: '"chrome:headless" portal-ui/tests/operator/tenant/test-4 --skip-js-errors -c 3'
+  # all-operator-tests-4:
```
kaankabalak, line 1698: Hi @reivaj05, quick question, why are we commenting out these tests?

reivaj05: Hi @kaankabalak With this PR the screen for audit logs will now be shown as the image above, and the old UI will be hidden by default, I need to disable all the tests that are using that old UI so the build can pass. I don't want to remove those tests yet until we actually deprecate audit logs in the operator itself

kaankabalak (approved):

bexsoft (approved):
LGTM

PR #2448, opened 2022-11-10
Add permission guidance if User doesn't have permission to view Objects in bucket

jinapurapu:
Text did not distinguish between an empty Bucket and the User not having permission to view Objects in the Bucket. Now, if User does not have required permission, they are informed.
Screen Shot 2022-11-09 at 4 26 21 PM


kaankabalak (approved):

prakashsvmx (approved):

reivaj05 (approved):

prakashsvmx (approved):
LGTM :+1:

reivaj05 (approved):

bexsoft (approved):

Issue #2453, opened 2022-11-14
Can't use 7-16 length numerical string as bucket name


imizimi95:
## Expected Behavior
7-16 character long numerical string should be accepted as bucket name.

## Current Behavior
From RELEASE.2022-09-07T22-25-02Z, if the bucket name contains only numbers from length 7 to 16 characters, minio mistakenly identifies it as an ip address and can not create the bucket with that name.
![2022-11-14_15-09-17](https://user-images.githubusercontent.com/3495619/201683713-fd444111-5da7-4ad4-8b05-8d52bac16a52.gif)

## Regression
Maybe. 3a3a4b2

## Additional notes
Please check the related backend validations too.

PR #2454, opened 2022-11-14
Escape dot char in ipAddressFormat regex (#2453)

imizimi95:


imizimi95:
😄 The regex check is also faulty in eslint: https://github.com/minio/console/actions/runs/3462446964/jobs/5781711327  
It's not a useless escape character.

bexsoft (changes requested):
Hi!
Same issue is still visible, I think expression must be changed:

PR #2457, opened 2022-11-15
fix crash in ui for empty log entries

prakashsvmx:
fix crash in ui for empty log entries


harshavardhana (approved):

bexsoft:
Let's include this PR in the release too please

PR #2466, opened 2022-11-17
Generate Tenant health report and upload to Subnet

jinapurapu:
Screen Shot 2022-11-29 at 1 13 53 PM

https://user-images.githubusercontent.com/65002498/204649239-6dd55ffa-bb3d-4d6b-be98-3c9173884e54.mov


.github/workflows/jobs.yaml
```diff
@@ -2517,7 +2517,7 @@ jobs:
           go tool cover -func=all.out | grep total > tmp2
           result=`cat tmp2 | awk 'END {print $3}'`
           result=${result%\%}
-          threshold=53.40
+          threshold=53.30
```
reivaj05, line 2520: Please add more tests to cover the new code that was added, we should be increasing coverage not reducing the threshold

reivaj05 (changes requested):

.github/workflows/jobs.yaml
jinapurapu: I added a test for the /health endpoint I made, which brought the coverage up from 53.0%, thought it was a rounding issue. What is included is the coverage check, is it just the API tests or others too?

operatorapi/tenants.go
```diff
@@ -2752,6 +2772,363 @@ func getUpdateDomainsResponse(session *models.Principal, params operator_api.Upd
...
+	if e := enc.Encode(header); e != nil {
+		return e
+	}
+	if e := enc.Encode(healthInfo); e != nil {
+		return e
+	}
+
+	return nil
+}
+
```
reivaj05, line 2811: subnet package already has functionality to make http requests, can that functionality be used here instead of adding new methods?

operatorapi/tenants.go
```diff
@@ -2752,6 +2772,363 @@ func getUpdateDomainsResponse(session *models.Principal, params operator_api.Upd
...
+func subnetAPIKeyAuthHeaders(apiKey string) map[string]string {
+	return map[string]string{"x-subnet-api-key": apiKey}
+}
+
+func subnetURLWithAuth(uploadURL string, apiKey string) (string, map[string]string, error) {
+	if len(apiKey) == 0 {
+		// API key not available in minio/mc config.
+		// Ask the user to log in to get auth token
+		subnetHTTPClient := &xhttp.Client{Client: restapi.GetConsoleHTTPClient("")}
+		token, thisThing, e := restapi.SubnetLogin(subnetHTTPClient, "jill@min.io", "testPassword1!")
```
reivaj05, line 2874: Why these credentials are hardcoded?

.github/workflows/jobs.yaml
reivaj05: You can view [here](https://play.min.io:9443/buckets/builds/browse/bWluaW8vY29uc29sZS8yNDY2LzM2NDQxNzM3Mzkv) the coverage of your build, most of the handlers you added and the code in tenants.go file is not tested

operatorapi/tenants.go
```diff
@@ -2752,6 +2772,363 @@ func getUpdateDomainsResponse(session *models.Principal, params operator_api.Upd
...
+
+	return respCode, nil
+}
+
+// UTCNow - returns current UTC time.
+func UTCNow() time.Time {
+	return time.Now().UTC()
+}
+
+func getTenantHealthReport(session *models.Principal, params operator_api.TenantHealthReportParams) (*models.HealthReport, *models.Error) {
```
reivaj05, line 2959: It might be better to split this long function into smaller pieces

.github/workflows/jobs.yaml
jinapurapu: Thank you, I will clear out the redundant Subnet functions and add testing for the remaining new ones.

portal-ui/build/index.html
```diff
@@ -1 +1,3 @@
...
\ No newline at end of file
+
```
dvaldivia, line 1: No need to send this file

dvaldivia (changes requested):

PR #2473, opened 2022-11-21
Display warning for public access policy in bucket settings

Alevsk:
image


portal-ui/src/screens/Console/Buckets/BucketDetails/SetAccessPolicy.tsx
```diff
@@ -143,6 +143,19 @@ const SetAccessPolicy = ({
                 ]}
               />
             </Grid>
+            {accessPolicy === "PUBLIC" && (
+              <div
+                style={{
```
kaankabalak, line 148: Is there any specific reason why we're using inline styles here? Maybe it might be better for consistency purposes if we use a class here instead

dvaldivia: styles that are not meant to be re-used should be inlined as to avoid having dead styling code floating around

kaankabalak: Sounds good @dvaldivia

kaankabalak (approved):

prakashsvmx (approved):

reivaj05 (approved):

jinapurapu (approved):

PR #2480, opened 2022-12-01
Separate out dependency vulnerability checks in CI

donatello:
These tests are not "requirements" to run other tests. If there are vulnerabilities found, it will still show up as CI failure, it just wont hold back other CI tests.

### DIFF
--- a/.github/workflows/jobs.yaml
+++ b/.github/workflows/jobs.yaml
@@ -283,8 +283,8 @@ jobs:
           pip3 install semgrep
           semgrep --config semgrep.yaml $(pwd)/portal-ui --error
 
-  no-warnings-and-make-assets:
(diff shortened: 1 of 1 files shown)


.github/workflows/jobs.yaml
```diff
@@ -338,6 +335,57 @@ jobs:
...
+      - uses: actions/cache@v3
+        id: assets-cache
+        name: Assets Cache
+        with:
+          path: |
+            ./portal-ui/build/
+          key: ${{ runner.os }}-assets-${{ github.run_id }}
+          restore-keys: |
+            ${{ runner.os }}-assets-
+
```
donatello, line 388: I've removed the `yarn audit` step that was here and put it into its own job.

reivaj05 (approved):

jinapurapu (approved):

bexsoft (approved):
LGTM

PR #2500, opened 2022-12-12
Split Object Browser and Buckets Admin

dvaldivia:
Split the Object Browser and the Buckets Admin page

Screenshot 2022-12-12 at 12 12 37 PM

Screenshot 2022-12-12 at 12 12 39 PM

### DIFF
--- a/.github/workflows/jobs.yaml
+++ b/.github/workflows/jobs.yaml
@@ -252,10 +252,43 @@ jobs:
         working-directory: ./portal-ui
         continue-on-error: false
         run: |
-          yarn audit
+          yarn audit  
   
   
-  
+
+  testcafe-cache:
+    name: "Testcafe Cache Stage"
+    #    needs:
+    #      - ui-assets
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: [ 1.18.x ]
+        os: [ ubuntu-latest ]
+    steps:
+      - name: Read .nvmrc
+        id: node_version
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ env.NVMRC }}
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - uses: actions/cache@v3
+        id: testcafe-cache
+        name: Testcafe Cache
+        with:
+          path: |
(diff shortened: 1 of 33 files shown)


bexsoft:
https://github.com/dvaldivia/console/pull/5

.github/workflows/jobs.yaml
```diff
@@ -16,207 +16,6 @@ concurrency:
...
-          echo "start ---> make test-operator-integration";
-          make test-operator-integration;
-
-      - uses: actions/cache@v3
-        id: coverage-cache-operator
-        name: Coverage Cache Operator
-        with:
-          path: |
-            ./operator-integration/coverage/
-          key: ${{ runner.os }}-coverage-2-operator-${{ github.run_id }}
```
reivaj05, line 219: Are these tests no longer needed?

reivaj05: Nvm, they went to the bottom 🤦

reivaj05 (approved):

.github/workflows/jobs.yaml
dvaldivia: hahaha yeah, I re-arranged the file

cniackz:
Tested locally:

image

bexsoft (approved):

cniackz:
All is looking good, my only concern is the failure at test:

Screenshot 2022-12-14 at 6 17 55 PM

But maybe this isn't related as we are just changing the UI, right?.

In any case, the log is:

```
    user_api_bucket_test.go:2825: 
        	Error Trace:	/home/runner/work/console/console/integration/user_api_bucket_test.go:2825
        	Error:      	Not equal: 
        	            	expected: 204
        	            	actual  : 500
        	Test:       	TestReplication
        	Messages:   	Http Response: {"code":500,"detailedMessage":"The remote target does not exist","message":"an error occurred, please try again"}
--- FAIL: TestReplication (0.51s)
```

and the code around is:

```go
	// 7. Delete remaining Bucket Replication Rule with generic end point:
	// /buckets/{bucket_name}/delete-all-replication-rules
	response, err = DeletesAllReplicationRulesOnABucket(
		originBucket,
	)
	assert.Nil(err)
	if err != nil {
		log.Println(err)
		return
	}
	finalResponse = inspectHTTPResponse(response)
	if response != nil {
		assert.Equal(204, response.StatusCode, finalResponse)
	}
```

We can open another PR to fix this test

cniackz (approved):
LGTM

PR #2510, opened 2022-12-15
[WIP] - Executing only failing test

cniackz:
Debugging please ignore.


cniackz:
This is a Console UI update required because we need to ignore the "The remote target does not exist" message and return either 404 or 202 but not 500. Will try to update console source code.

PR #2514, opened 2022-12-15
Remove commented jobs

reivaj05:
This tests were testing the audit log section of a tenant which we don't support anymore and now we display a deprecated message instead, there is no reason to keep having this code


reivaj05:
Need to understand how tests were split

bexsoft (approved):

prakashsvmx (approved):

dvaldivia (approved):

PR #2515, opened 2022-12-15
Ignoring message from MinIO

cniackz:
### Objective:

Ignore MinIO Message on deletion rules.

### Issue:

https://github.com/cniackz/public/wiki/Chasing-MinIO-Bug-in-Replication

### Solution:

As discussed with @harshavardhana , we should ignore the message on delete in the API. After @poornas changed in PR https://github.com/minio/minio/pull/16221 this message is expected and cannot be considered as an error.

### Relevant change:

Screenshot 2022-12-16 at 9 51 50 AM


restapi/admin_remote_buckets.go
```diff
@@ -109,10 +109,10 @@ func registerAdminBucketRemoteHandlers(api *operations.ConsoleAPI) {
 
 	// delete all replication rules for a bucket
 	api.BucketDeleteAllReplicationRulesHandler = bucketApi.DeleteAllReplicationRulesHandlerFunc(func(params bucketApi.DeleteAllReplicationRulesParams, session *models.Principal) middleware.Responder {
-		err := deleteBucketReplicationRulesResponse(session, params)
-		if err != nil {
-			return bucketApi.NewDeleteAllReplicationRulesDefault(int(err.Code)).WithPayload(err)
-		}
+		_ = deleteBucketReplicationRulesResponse(session, params)
```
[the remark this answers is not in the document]
cniackz: That is correct, 404 is still valid, I will correct that, thank you Javier 👍

cniackz:
Thank you @dvaldivia I have correted the error that Javier pointed out. My next step in this PR is to use latest version of Swagger to reduce the number of changed files.

cniackz:
Ok, I am done with the changes, new Swagger version reduced the files changed and improved the way we handle the errors as per Javier feedback... Once all tests are passing, I will ping you guys in the channel for another round of review.

restapi/admin_remote_buckets.go
```diff
@@ -111,9 +111,14 @@ func registerAdminBucketRemoteHandlers(api *operations.ConsoleAPI) {
 	api.BucketDeleteAllReplicationRulesHandler = bucketApi.DeleteAllReplicationRulesHandlerFunc(func(params bucketApi.DeleteAllReplicationRulesParams, session *models.Principal) middleware.Responder {
 		err := deleteBucketReplicationRulesResponse(session, params)
 		if err != nil {
+			if err.Code == 500 && *err.DetailedMessage == "The remote target does not exist" {
```
reivaj05, line 114: Is minio returning a 500 when the target does not longer exists?

go.mod
```diff
@@ -67,18 +67,24 @@ require (
 	github.com/charmbracelet/bubbletea v0.22.1 // indirect
 	github.com/charmbracelet/lipgloss v0.6.0 // indirect
 	github.com/cheggaaa/pb v1.0.29 // indirect
+	github.com/cilium/ebpf v0.9.3 // indirect
```
dvaldivia, line 70: I think all the go.mod changes are unnecessary, let's remove them from this PR

dvaldivia (changes requested):

bexsoft (approved):

dvaldivia (approved):

restapi/admin_remote_buckets.go
reivaj05: @cniackz This is already merged but we shouldn't return a 204 if the remote target doesn't exist

cniackz: Hello @reivaj05 the remote target does exist and the rules are getting deleted but MinIO is doing more that just this is our trace log; so after showing this trace to Harsha and Poorna they told me to ignore this one as the rules are being deleted anyway; what I was thinking is something more complex, to change the logic or the way we actually delete these rules, to have an API that only does this deletion and no 500 is returned, but that will require more work and a deeper understanding on how we interact with MinIO and how we can get MinIO to just delete the rules. Maybe in a new PR we can improve this.

PR #2516, opened 2022-12-15
Fixed Delete all object versions selector functionality

bexsoft:
## What does this do?

Fixed an issue were deleting all versions selector was not working as intended. Also added a notification that this action is dangerous.

This change doesn't include the bypass option yet.

## How does it look?
Screenshot 2022-12-15 at 17 48 04
Screenshot 2022-12-15 at 17 48 00
Screenshot 2022-12-15 at 17 47 52
Screenshot 2022-12-15 at 17 47 49


dvaldivia (approved):

prakashsvmx (approved):
Verified with versioned bucket
with mc and ui.
LGTM :+1:

PR #2536, opened 2022-12-28
Added delete governance bypass option when deleting file versions

bexsoft:
## What does this do?

Added delete governance bypass option when deleting file versions

## How does it look?
Screenshot 2022-12-28 at 15 51 16
Screenshot 2022-12-28 at 15 51 08
Screenshot 2022-12-28 at 15 51 00
Screenshot 2022-12-28 at 15 50 50
Screenshot 2022-12-28 at 15 50 46
Screenshot 2022-12-28 at 15 50 31
Screenshot 2022-12-28 at 15 50 28


prakashsvmx (approved):
Tested.  LGTM :+1: 
- Versioned Bucket/Objects
- Lock with Retention set to compliance mode Bucket/Objects
- Lock with Retention set to governance mode Bucket/Objects

With compliance mode the error message can be more meaningful and could be updated later.
>We encountered an internal error, please try again.: cause(force-delete is forbidden in a locked-enabled bucket)

as the error is very generic and does not indicate that, compliance object cannot be deleted

Whereas deleting a version has the following message
>Object, '1.png (Version ID=469ef74c-c72e-4cb0-9ef6-db999f96e9ec)' is WORM protected and cannot be overwritten

dvaldivia (approved):

reivaj05 (approved):

PR #2551, opened 2023-01-05
Increase coverage

dvaldivia:


kaankabalak:
Seems like the coverage limit check is failing now with the following message:

```
Result:
55.3%
It is smaller than threshold (55.8%) value, failed!
```

dvaldivia:
@kaankabalak dang it, was trying to bring it above threshold hahaha will add some more

dilverse (approved):

kaankabalak (approved):
Approving as the coverage is increased

PR #2564, opened 2023-01-10
fix prometheus dashboard values for luxon migration

prakashsvmx:
UI Crash in master in prometheus widgets :
```
Uncaught TypeError: timeEnd.unix is not a function
```

Navigate to Prometheus dashboard  ( Monitoring -> Metrics), select a date range and click Sync observe the error in master.

With fix, it should not crash . 
Also verify the data displayed in the date range.


reivaj05 (approved):

bexsoft (approved):
LGTM

PR #2591, opened 2023-01-24
Fixed WS reconnection logic onError and onClose

bexsoft:
## What does this do?

Removes duplicated reconnection request when error is reached on Object Browser websocket, also removed the use of the connect call back as it will contain references to the initial request and it is not necessary as we already have information available.  


dvaldivia (approved):
Tested. Works.

harshavardhana (approved):

PR #2628, opened 2023-02-02
Expose error on Tenant if certificate is near expiration or expired

allanrogerr:
## What does this do?

- Added time left to certificates
```
if timeToExpire < 30 && timeToExpire >= 10 then text is orange
if timeToExpire < 10 then text is red
if timeToExpire < 1 day then text is verbose
```

## How does it look?
image


dvaldivia:
@allanrogerr how about instead of saying `Expires in: 0 hours, 0 minutes` we show `Expires in: EXPIRED`?

allanrogerr:
> @allanrogerr how about instead of saying `Expires in: 0 hours, 0 minutes` we show `Expires in: EXPIRED`?

Sure. This is what it would look like:

kaankabalak (approved):

dvaldivia (approved):

PR #2631, opened 2023-02-03
Updated copyright tags to 2023

bexsoft:
## What does this do?

Updated Copyright tags to 2023


dilverse (dismissed):

dvaldivia (changes requested):
We should only do this for `NOTICE` and autogenerated files, but not for any file we have not changed since it was created

reivaj05 (approved):

jinapurapu (approved):

allanrogerr (approved):
These look like swagger auto generated files and a NOTICE

dvaldivia (approved):

PR #2678, opened 2023-02-21
Hidden red notification in License menu when cluster is registered

bexsoft:
## What does this do?

Hidden red notification in License menu when cluster is registered as it is not required

## How does it look?

Screenshot 2023-02-21 at 17 41 16


kaankabalak (approved):

allanrogerr (approved):
Thanks!

Issue #2021, opened 2022-05-20
can't preview text/plain files without .txt extension


buptzhoutian:
For example, one file named `service.log` was uploaded to MinIO, the Content-Type of this object is "text/plain".

This `service.log` file can't be previewed. However, the same file uploaded as `service.log.txt` can be previewed.

prakashsvmx:
It depends on browsers too.

e.g
`.log`
 Chrome can load `.log` files in an iframe for Preview 
 Firefox Will not load.

`.txt`
for `.txt` both browsers will display.

prakashsvmx:
as per code the below are the supported extensions for preview.

```javascript
export const extensionPreview = (
  fileName: string
): "image" | "text" | "audio" | "video" | "none" => {
  const imageExtensions = [
    "jif",
    "jfif",
    "apng",
    "avif",
    "svg",
    "webp",
    "bmp",
    "ico",
    "jpg",
    "jpe",
    "jpeg",
    "gif",
    "png",
    "heic",
  ];
  const textExtensions = ["pdf", "txt", "json"];
  const audioExtensions = ["wav", "mp3", "alac", "aiff", "dsd", "pcm"];
  const videoExtensions = [
    "mp4",
    "avi",
    "mpg",
    "webm",
    "mov",
    "flv",
    "mkv",
    "wmv",
    "avchd",
    "mpeg-4",
  ];

```

dvaldivia:
We stopping previewing text files for security reasons. The alternative is to generate a share link and paste it in another tab.

PR #2680, opened 2023-02-23
Disabled Preview button for text files

bexsoft:
## What does this do?

Disabled Preview button for text files

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
@@ -138,7 +138,7 @@ export const extensionPreview = (
     "png",
     "heic",
   ];
-  const textExtensions = ["pdf", "txt", "json"];
+  const textExtensions = ["pdf"];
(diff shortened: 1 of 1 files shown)


reivaj05 (approved):

allanrogerr (dismissed):
Simple fix :)

jinapurapu (dismissed):

allanrogerr (approved):

jinapurapu (approved):

jankap:
Why exactly has the preview been disabled?

ianroberts:
What was the reasoning behind this decision?  I find the preview option very useful for plain text files as it means I can upload a `_README.txt` into folders that would otherwise be empty, explaining what sort of files should be dropped there by other people.  I suppose I could make a PDF instead, but why disable plain text?  Was this to fix some other CVE or something?

cesnietor:
I'm trying to find out as well :D. I'm also curious about it. We'll share the context if we end up figuring out. but seems it was for security reasons but the reasons are not explained though. https://github.com/minio/console/issues/2021#issuecomment-1132942518.

PR #2684, opened 2023-02-27
move to UBI micro image

harshavardhana:


pjuarezd:
As long as we keep using any of the following images as base we can still push it to ECR (and make it available for AWS Marketpalce). We are pushing two images to ECR, Console and Operator

>  UnsupportedImageError
You may get an UnsupportedImageError error when attempting to perform a basic scan on an image that was built using an operating system that Amazon ECR doesn't support basic image scanning for. Amazon ECR supports package vulnerability scanning for major versions of Amazon Linux, Amazon Linux 2, Debian, Ubuntu, CentOS, Oracle Linux, Alpine, and RHEL Linux distributions. Once a distribution loses support from its vendor, Amazon ECR may no longer support scanning it for vulnerabilities. Amazon ECR does not support scanning images built from the [Docker scratch](https://hub.docker.com/_/scratch) image.

https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-scanning-troubleshooting.html

pjuarezd (approved):

dvaldivia (approved):

PR #2693, opened 2023-03-06
Add playwright Test

cniackz:
### Objective:

Start exploring playwright to complement testcafe and if we like it, migrate to playwright little by little.

### How to manually test:

1. Have `console` running at port `5005` as we normally test
2. Under `playwright/tests` new added folder run:

```
npx playwright test
```

* This will execute `verify-title.spec.ts` which is a TypeScript test that checks just the title to contain `MinIO`
* Expected result is:

```sh
$ npx playwright test

Running 1 test using 1 worker

  ✓  1 example.spec.ts:3:5 › has title (624ms)

  1 passed (1.1s)
```

### Dependencies:

Screenshot 2023-03-07 at 9 33 31 AM

### DIFF
--- a/.github/workflows/jobs.yaml
+++ b/.github/workflows/jobs.yaml
@@ -1307,3 +1307,45 @@ jobs:
             echo "It is smaller than threshold ($threshold%) value, failed!"
             exit 1
           fi
+
+  playwright:
+    needs:
+      - compile-binary
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
+    - name: Install dependencies
+      run: |
+        echo "Run tests under playwright folder only"
+        echo "package-lock.json required"
+        cd $GITHUB_WORKSPACE/playwright
+        npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+
+    - uses: actions/cache@v3
+      name: Console Binary Cache
+      with:
+        path: |
+          ./console
+        key: ${{ runner.os }}-binary-${{ github.run_id }}
+
+    - name: Start Console, front-end app and initialize users/policies
+      run: |
+        (./console server) & (make initialize-permissions)
+
+    - name: Run Playwright tests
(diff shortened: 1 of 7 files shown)


.github/workflows/jobs.yaml
```diff
@@ -1307,3 +1307,45 @@ jobs:
...
+  playwright:
+    needs:
+      - compile-binary
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
```
reivaj05, line 1320: node-version: ${{ env.NVMRC }}

.github/workflows/jobs.yaml
```diff
@@ -1307,3 +1307,45 @@ jobs:
...
+      run: |
+        (./console server) & (make initialize-permissions)
+
+    - name: Run Playwright tests
+      run: |
+        echo "Run tests under playwright folder only"
+        cd $GITHUB_WORKSPACE/playwright
+        npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
```
reivaj05, line 1347: Do we need to run this step if tests failed?

playwright/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,209 @@
+name: Workflow
```
reivaj05, line 1: Why is this file needed?

.github/workflows/jobs.yaml
dvaldivia: the playwrigth create script proposes this and it uses 16 for the framework

playwright/.github/workflows/playwright.yml
cniackz: It is needed because we can use it to run only playwright in github actions, mainly for testing but can also be removed but manual intervention will be needed to run standalone mode.... Anyway I will soon close this PR because we have a better approach in other PR.

cniackz:
Thank you @reivaj05 for the review, but we are having new approach in https://github.com/minio/console/pull/2702, please help us review it there.

PR #2704, opened 2023-03-13
Introducer swagger typescript api

dvaldivia:
Introduces an auto-generated swagger API in typescript, this is to make sure all types are always up to date around the API Specification and to force all API's to be strongly typed.

Buckets is the first API to be migrated.


bexsoft:
Please fix semgrep checks

reivaj05 (approved):

bexsoft (approved):
LGTM

PR #2728, opened 2023-03-21
Fixed refresh button on ListObjects screen

jinapurapu:
Refresh button on ListObjects screen now reloads latest bucketInfo. 

https://user-images.githubusercontent.com/65002498/226708716-6fd070d6-e0f2-4056-b170-09d713e8b8da.mov

closes https://github.com/miniohq/engineering/issues/1181

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/ListObjects/ListObjects.tsx
@@ -416,7 +416,7 @@ const ListObjects = () => {
 
   // bucket info
   useEffect(() => {
-    if (loadingObjects || (loadingBucket && !anonymousMode)) {
+    if ((loadingObjects || loadingBucket) && !anonymousMode) {
       api
         .invoke("GET", `/api/v1/buckets/${bucketName}`)
         .then((res: BucketInfo) => {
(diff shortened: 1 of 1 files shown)


kaankabalak:
I'm not sure if I am doing anything wrong, however, even though it looks like an API request is being made now when the user clicks on the Refresh button which is great, I think the main problem seems to be that the count that comes in from the backend takes time to update to the correct value (took around 100 seconds on both cases below when I uploaded and deleted an object):

Screen Shot 2023-03-21 at 1 49 57 PM

Screen Shot 2023-03-21 at 1 52 24 PM

jinapurapu:
> I'm not sure if I am doing anything wrong, however, even though it looks like an API request is being made now when the user clicks on the Refresh button which is great, I think the main problem seems to be that the count that comes in from the backend takes time to update to the correct value (took around 100 seconds on both cases below when I uploaded and deleted an object):
> 
> Screen Shot 2023-03-21 at 1 49 57 PM
> 
> Screen Shot 2023-03-21 at 1 52 24 PM

Yes, I will look into that further, as well as making more clear the situation with deleted versioned objects (these seem to remain in the object count without explanation to the user, causing mismatch between the visible objects and the number shown in the Bucket Info section). This PR is just to fix the non-functional refresh button to start.

kaankabalak:
> Yes, I will look into that further, as well as making more clear the situation with deleted versioned objects (these seem to remain in the object count without explanation to the user, causing mismatch between the visible objects and the number shown in the Bucket Info section) to fix in future PRs. This PR is just to fix the non-functional refresh button to start.

Sounds good, I'll approve this PR as the lint failures seem to be unrelated 👍

kaankabalak (approved):

reivaj05 (approved):

bexsoft (approved):
LGTM, In the future we may need to simplify this in one call (perhaps adding this information to the websocket too. This may be handled in the update in real time task in future releases)

PR #2729, opened 2023-03-21
[WIP] - Fixing Unused params Linter Check

cniackz:
### Objective:

To improve our code to pass linter, regardless of the previous state, the complains from the linter are valid and if we are having vars that are unsed and things like that, we should correct the code first.

I think there was a recent change in revive and now it is checking for unused params.

### Spot 1:

* `pkg/logger/target/http/http.go` removed var here
* no need to append it here `console/pkg/logger/targets.go`

### Spot 2:

* Using function for testing, printed unused params `console/pkg/logger/logger_test.go`

### Spot 3:

* `pkg/logger/console.go:167:24: unused-parameter: parameter 'msg'`

### Spot 4:

* `restapi/logs.go:42:12: unused-parameter: parameter 'ctx'`

### DIFF
--- a/pkg/logger/logger_test.go
+++ b/pkg/logger/logger_test.go
@@ -25,6 +25,8 @@ import (
 )
 
 func testServer(w http.ResponseWriter, r *http.Request) {
+	fmt.Println(w)
+	fmt.Println(r)
 }
 
 func TestInitializeLogger(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/logger/logger_test.go
```diff
@@ -25,6 +25,8 @@ import (
 )
 
 func testServer(w http.ResponseWriter, r *http.Request) {
+	fmt.Println(w)
+	fmt.Println(r)
 }
```
reivaj05, line 30: func testServer(_ http.ResponseWriter, _ *http.Request)  ?

reivaj05: Didnt' see this was still in progress :P, i'll review again once you finish

Issue #2760, opened 2023-04-05
"Network error occured" when downloading large file


heiko-ma:
When downloading a large file (97GB in my test) via the console, the download interrupts after a while with the following error:
_Error: A network error occurred._

The browser console shows the following error
![image](https://user-images.githubusercontent.com/64192965/230032649-97946147-9fd1-453c-81b9-9d1159330c8a.png)
When I download die file via the direct link (<<URL>>/objects/download?prefix=dGVzdA==&version_id=null) the download completes without error. I suspect a problem in the JavaScript.

dvaldivia:
that is a valid assumption, perhaps the `XMLHttpRequest` object is not great for dealing with large files but we use it to show some progress when downloading the file, when you download the file directly you are not going through the JS, perhaps we should STAT the file and if it's beyond a certain mark just do direct download, unless we implement some chunking on the client

heiko-ma:
The nice thing about the JS-Solution is that it is relatively CPU friendly. I observed a high CPU/GPU-Usage due to scanning from Windows Defender (Microsoft Network Realtime Inspection Service). When passing the download through JS, it doesnt trigger the Inspection Service.

dvaldivia:
looking around, perhaps this [gist](https://gist.github.com/semlinker/8453bbad093caaf321b153285b350d84) approach which downloads the file using byte ranges, worth trying

allanrogerr:
1MB, 10MB, 100MB, 1G, 10G file uploaded and downloaded without issue. Testing 100G file.

allanrogerr:
@reivaj05 Failed at 100G with "Resource requested is unreadable, please reduce your request rate". My steps:

- clone console repo
- run make assets
- replace the console entry in the minio go.mod file
- make docker in minio
- load the docker image to kind
- create a minio tenant using this minio image
- attempt an upload/download of a 100G file
- face above issue

PR #2773, opened 2023-04-11
Fix download of large files in Console

reivaj05:
Fixes: https://github.com/minio/console/issues/2760


dvaldivia:
I found a memory problem while downloading a large file, to test this I did the following

1. Generate a `100GiB` file via `mkfile -n 100g ~/Desktop/LargeTestFile` (on macOS)
2. Upload the file via object browser, this completed just fine
3. attempt to download the file, then `Safari` reset the page due to memory pressure, but I saw the download continues even if the download progress component no longer reports it, after a few zombie connections, it stops al together
Screenshot 2023-04-12 at 11 41 37 AM

dvaldivia:
perhaps for Safari we can force a direct download? @reivaj05

reivaj05:
@dvaldivia I believe we won't be able to set anonymous access header if we force a direct download, is that right?

dvaldivia:
@reivaj05 when do we need to do that? :S

reivaj05:
Added way to download directly in safari if file size is more than 4G

bexsoft (changes requested):
Tested the PR with the following scenarios:

- Uploaded 5GB File, everything OK
- Downloaded File in Chrome, Need to find a way to display more constant batches for the progress bar as it updates only every 25%

- On Safari, the whole application freezes between batches and sometimes crashes once the download has been completed

- On Firefox Navigator Download manager & MinIO's object manager opens at the same time (not really an issue as current behavior opens Firefox manager after MinIO's object manager completes the download)
-  On Opera same behavior as Firefox is observed

cesnietor:
Tested  on Chrome with 4G and 100G files ✅ 
On Safari I also experienced the same behavior raised by @bexsoft  where the app crashes and kind of sends me to the Object browser page. This happens even with 4G files.  Small files seem to be working fine.

reivaj05:
@bexsoft i couldn't find a better way to display progress, current approach to get the chunks is `fileSize/1.5G`, i could reduce the chunk size i9n the code but smaller chunks will translate in more requests to minio and rate limit errors, with 1.5G we avoid the rate limit error

@bexsoft @cesnietor i'm going to let safari browser handle the downloads directly instead of setting a limit because seems 4G files is still too much for safari to handle

cesnietor:
Tested in Safari and is download is now being handled by the browser. ✅ 
Screenshot 2023-05-11 at 11 44 49 AM

cesnietor (dismissed):
LGTM

cesnietor (approved):
typo fixed

PR #2777, opened 2023-04-14
chore: disable animation for login page

cgpoh:
Following @dvaldivia to disable animation for login page until [kubernetes/kubectl#1368](https://github.com/kubernetes/kubectl/issues/1368) get resolved

### DIFF
--- a/portal-ui/public/config.js
+++ b/portal-ui/public/config.js
@@ -1,0 +1,1 @@
+window.DISABLE_ANIMATION = false
--- a/portal-ui/public/index.html
+++ b/portal-ui/public/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <script src="%PUBLIC_URL%/config.js"></script>
     <meta charset="utf-8" />
     <base href="/" />
     <meta content="width=device-width, initial-scale=1" name="viewport" />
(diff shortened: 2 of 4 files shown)


dvaldivia:
I don't think we should disable the animation on the console itself, since it's expected to expose the console via ingress or similar, for the operator UI perhaps not everyone will want to expose it so I deem it necessary to disable the animation, perhaps for Console if we really need this we should disable it via environment variable

cgpoh:
Thanks for the comment. We can’t use ingress due to security reason and we can only use port forward to login to minio console. I will try to use environment variable to allow turning on/off the animation.

dnskr:
Could someone review this please?

bexsoft:
Hi @cgpoh!, Thank you for your contribution

For this behavior to work, we will need some changes in MinIO as we don't support the ability to modify JS files directly into the server. We have created this PR that will help to achieve this behavior: https://github.com/minio/console/pull/2799 using env variables directly that can be read & managed directly from MinIO

We will be implementing the required change in MinIO as well. We will be closing this PR, Please feel free to reopen in case  you think something is missing

cgpoh:
> Hi @cgpoh!, Thank you for your contribution
> 
> For this behavior to work, we will need some changes in MinIO as we don't support the ability to modify JS files directly into the server. We have created this PR that will help to achieve this behavior: #2799 using env variables directly that can be read & managed directly from MinIO
> 
> We will be implementing the required change in MinIO as well. We will be closing this PR, Please feel free to reopen in case you think something is missing

Hi @bexsoft , thanks for your help!

PR #2780, opened 2023-04-18
fix to show or hide show deleted objects option based on versioning status

prakashsvmx:
fix to show or hide show deleted objects option based on versioning status

"Show Deleted Objects" option is displayed  for non versioned bucket. 
it should be displayed only for versioned state ( enabled, suspended) buckets

Tested in :
```
Locking enabled bucket
Versioning enabled bucket
Versioning suspended bucket
Non versioned bucket
```


jinapurapu:
Just checking that this is correct behavior - If versioning is suspended, deleted objects are still visible, but can't be restored, and objects deleted while versioning is suspended will show as deleted objects, but not be recoverable even once versioning is enabled. The only case the "Show deleted objects" box should not show is for never-versioned buckets, correct?

prakashsvmx:
@jinapurapu 

here is a summary of the behaviour in each of the buckets

```
Locking enabled bucket
  "Show deleted objects" option would be displayed  as versioning is enabled by default. 
  "Restore option is available per object  (from object versions list)"  is  available 
  Rewind option is available

Versioning enabled bucket
"Show deleted objects" option would be displayed  as versioning is enabled by default. 
 "Restore option is available per object (from object versions list)"  if multiple versions are/were available
 Rewind option is available
  
Versioning suspended bucket
"Show deleted objects" option would be displayed  as versioning is enabled by default. 
 "Restore option is available per object (from object versions list)"  if multiple versions are/were available
  Rewind option is availabe
  
Non versioned bucket
 "Show deleted objects" - Not available
 "Restore " - Not available
  Rewind - Not available
```

jinapurapu (dismissed):

dvaldivia:
I only found 1 problem @prakashsvmx if you are looking at a `non-version` bucket, and do `ctrl/cmd + k` to navigate to a `versioned` bucket, the `show deleted objects` option doesn't render, other than that, the change works fine

prakashsvmx:
@dvaldivia this seem to happen only with KBar actions.  and looks not due to these changes.

The object broweser in does not re render and populate the required states . i will investigate further in a new PR. 
with master branch,  it does not even refresh the object list when navigated via KBar action. I suspect some react router issue

prakashsvmx:
@dvaldivia  i have sent a fix for the re-rendering bug when the bucket name changes within the Object browser view  e.g using KBar. PTAL.

bexsoft:
Can you please help us to fix conflicts @prakashsvmx? Thank you!

bexsoft (approved):
LGTM

bexsoft:
> @prakashsvmx seems that the coverage check is failing, we'd need to add some tests for this.

I think this issue is not related to this PR, seems like threshold decreased while generating assets. We can manage this in a second PR

cesnietor:
Tested:

Bucked Locked:
- Shows `Show deleted objects`  ✅ 
- Item allows to see previous versions  ✅ 
- Allows deletion and restore  ✅ 
-- Found an issue where the UI shows the modal to disable versioning for the locked bucket but this is not possible so we should hide that propbably (can be addressed separetely). ⚠️ 

Screenshot 2023-05-10 at 10 14 07 AM

Bucket Versioned:
- Shows `Show deleted objects` ✅ 
- Item allows to see previous versions ✅ 
- Allows deletion and restore ✅ 


Bucket versioned suspended
- Shows `Show deleted objects` ✅ 
- Item allows to see previous versions ✅ 
- Allows deletion and restore **Only on those objects that were created while versioning was enabled** ✅ 
- If an object was added after Versioning was suspended, when trying to delete it, it shows the `Delete all versions` which is wrong since versioning was suspended right?. ⚠️ 

Screenshot 2023-05-10 at 10 11 35 AM

Bucket no versioning:
- Does not show `Show deleted objects` ✅ 
- Not able to see previous versions ✅ 
- Allows deletion ✅

dvaldivia (approved):

prakashsvmx:
@cesnietor Thank you so much for testing this out and the detailed feedback.

>- Found an issue where the UI shows the modal to disable versioning for the locked bucket but this is not possible so we should hide that propbably (can be addressed separetely [could we create a ticket for this pray ]). warning

yes,  i could send a fix.

>If an object was added after Versioning was suspended, when trying to delete it, it shows the Delete all versions which is wrong since versioning was suspended right?.

We would not exactly when the versioning was suspended.  So i believe this is fine as the older objects may have versions to delete.

PR #2805, opened 2023-05-07
fix: websockets goroutine leak for object list UI

jiuker:
listen on the done channel to ensure that we can clean up 
the WebSocket go-routine, once the caller is done with
writing the response.


bexsoft (changes requested):
Hi!

Thank you for your contribution, I have tested this change but loader stays forever if you try to navigate to a second or third bucket.

Please refer to the following gif for this issue:

jiuker:
Reproduction steps.
1. Add the following code to minio's startup code
```go
// Copyright (c) 2015-2021 MinIO, Inc.
//
// This file is part of MinIO Object Storage stack
//
// This program is free software: you can redistribute it and/or modify
// it under the terms of the GNU Affero General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version.
//
// This program is distributed in the hope that it will be useful
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU Affero General Public License for more details.
//
// You should have received a copy of the GNU Affero General Public License

package main // import "github.com/minio/minio"

import (
	"fmt"
	"os"
	"runtime"
	"time"

	// MUST be first import.
	_ "github.com/minio/minio/internal/init"

	minio "github.com/minio/minio/cmd"
)

func main() {
	go func() {
		ticker := time.NewTicker(time.Second)
		defer ticker.Stop()
		for _ = range ticker.C {
			runtime.GC()
			fmt.Println("go nums:", runtime.NumGoroutine())

		}
	}()
	minio.Main(os.Args)
}
```
3. Use server /somedir to start
The output will be 
```
go nums: 358
go nums: 358
go nums: 358
go nums: 358
go nums: 358
go nums: 358
go nums: 358
```
4. Login to the console, then wait a few seconds, the output will be stable all the time
5. Add a bucket, upload the file to this bucket, and wait in /browser/{bucket}.
Then use F5 to refresh the page and wait for the output to stabilize
Then you will find the number of output +1
```shell
go nums: 362
go nums: 362
go nums: 362
go nums: 362
go nums: 362
go nums: 362
go nums: 362
```
The number of times added is the number of refreshes
```go
go nums: 363
go nums: 363
go nums: 363
go nums: 363
go nums: 363
go nums: 363
```
Then use pprofile, you will find that this is the place where the go did not exit.

@bexsoft

jiuker:
This may not be a good solution, but there is really no good exit mechanism here.
@bexsoft

klauspost (approved):
Great find!

restapi/ws_objects.go
```diff
@@ -241,6 +241,9 @@ func (wsc *wsMinioClient) objectManager(session *models.Principal) {
 				LogInfo("Error while writing the message", err)
 				return
 			}
+			if writeM.RequestEnd {
```
dvaldivia, line 244: I don't think we should exit this goroutine, as we still re-use the socket for other subsequent listing requests over the socket, this goroutine should exit when `<-done` in line 250 is reached, if we have a leak perhaps, the socket closing is not properly killing the parent goroutine, so we should instead close the `done` channel when the socket gets closed

dvaldivia (changes requested):

restapi/ws_objects.go
```diff
@@ -232,18 +232,23 @@ func (wsc *wsMinioClient) objectManager(session *models.Principal) {
...
-				LogInfo("Error while parsing the response", err)
+		for {
+			select {
+			case <-done:
 				return
-			}
+			case writeM := <-writeChannel:
+				jsonData, err := json.Marshal(writeM)
+				if err != nil {
+					LogInfo("Error while parsing the response", err)
```
harshavardhana, line 242: This comment is wrong we are not parsing here we are marshalling.

harshavardhana (changes requested):

harshavardhana (approved):

cesnietor:
@jiuker  do you mind adding in the description what the change does and how it fixes the leak so that it is documented for posterity?

cesnietor (approved):
Tested adding objects, loading objects and everything looks good.

PR #2816, opened 2023-05-13
Added identifier field to Event destinations page & migrated to mds

### LINKED ISSUE #666: Adding of 2 lambda notifications of the same type
Hello.
We need to use namespace and access notifications for elasticsearch. When we have tried to add the second target through Console - it's just rewrite the first target (also, it has "elasticsearch:_" name). Through mc we can add as many target as we want.
Also after target adding we can not view info about target in Console.
I think that adding multiple notifications targets through Console is very useful. And viewer of current targets settings too.

bexsoft:
fixes #666

## What does this do?
- Added Identifier field to event destination page to allow multiple events of the same type. 
- Migrated all  visual components to mds

> **NOTE:** There is an issue where after adding a new Lambda Event Destination, list doesn't refresh, this is expected as you need to restart MinIO server to retrieve this new configuration. Also need to implement form validation in forms

## How does it look?
Screenshot 2023-05-12 at 18 51 12
Screenshot 2023-05-12 at 18 51 06
Screenshot 2023-05-12 at 18 50 58


reivaj05 (approved):

cesnietor (approved):
LGTM

allanrogerr (approved):

allanrogerr:
Lets fix #666 !

PR #2827, opened 2023-05-23
Changed Share Object logic to use Access Keys

### LINKED ISSUE #922: Links expire within an hour when sharing from web interface regardless of the value set
When i share a file from MinIO web interface its link expires within an hour despite choosing another value like 1 day or 2 days. Sharing files using third party S3 clients works fine

## Expected Behavior
Shared link must expire on the established date

## Current Behavior
When sharing files from web interface its download links expires 1 hour after their creation

## Possible Solution
The issue appear to be at web interface

1.Upload a file to a bucket
2.Share it from web interface
3.wait 1 hour or 2
4.Open sharing link

## Context
I was trying to share a video from a bucket on my personal website and video links stopped working. 2 days later i could share it without issues using a third party S3 client

## Regression
Is this issue a regression? No

## Your Environment
* Version used: RELEASE.2021-07-30T00-02-00Z
* Server setup and configuration: NGINX web server and the following config files
[api-nginx.txt](https://github.com/minio/minio/files/6925634/api-nginx.txt)
[panel-nginx.txt](https://github.com/minio/minio/files/6925635/panel-nginx.txt)
[minio-systemd.txt](https://github.com/minio/minio/files/6925652/minio-systemd.txt)
[minio-env.txt](https://github.com/minio/minio/files/6925657/minio-env.txt)

* Operating System and version : Ubuntu 20.04 (Linux asgardius-minio 5.8.0-1037-oracle minio/minio#38~20.04.1-Ubuntu SMP Fri Jul 16 01:02:14 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux)

bexsoft:
fixes #922 

## What does this do?

Changed Share Object Logic to use access keys instead of console login key. This is first step for this new change, as second phase we will auto generate auto expiring service accounts

## How does it look?

Before:
Screenshot 2023-05-23 at 9 41 15 AM

After:
Screenshot 2023-05-22 at 18 19 59
Screenshot 2023-05-22 at 18 19 47
Screenshot 2023-05-22 at 18 19 43
Screenshot 2023-05-22 at 18 19 32
Screenshot 2023-05-22 at 18 19 24


cesnietor:
Could we add something like `An Access Key with privileges for current user is required to sign this URL`  (adding the `for current user`) or something like that? Just to be clear that even being an admin won't list other service accounts (made me think that).

cesnietor:
I found that selecting 0 days, 0 hours, and 0 minutes hangs the UI.
Screenshot 2023-05-23 at 9 40 29 AM

swagger.yml
```diff
@@ -484,26 +484,19 @@ paths:
         - Object
 
   /buckets/{bucket_name}/objects/share:
-    get:
+    post:
```
reivaj05, line 487: I believe we should keep the method as get since the handler is not modifying the current state of the system, maybe access_key and secret_key can be passed as headers in the request and the rest as query parameters

reivaj05 (changes requested):

swagger.yml
bexsoft: As discussed, we will keep it as POST as this is the pattern we are following in other endpoints with this behavior. Also discussed with Daniel and its better to keep it as POST

cesnietor (approved):
Tested, LTM.
It nows gives the option to list user's service accounts or use other one by just typing it.

Screenshot 2023-05-23 at 3 37 21 PM

dvaldivia (approved):

grubjack:
trying to share a file under LDAP user with consoleAdmin permissions and get the error:

`{"Code":"AccessDenied","Message":"Access Denied.","Resource":"/minio/admin/v3/list-service-accounts","RequestId":"176498F6A04CC8B9","HostId":"dd90..."}`

under regular LDAP user get the error:

`{"Code":"AccessDenied","Message":"Access Denied.","Resource":"/minio/admin/v3/list-service-accounts","RequestId":"176498F6A04CC8B9","HostId":"dd90..."}`

PR #2839, opened 2023-05-29
Move Login to Typescript Swagger API

dvaldivia:
Making the `Login` screen more self-contained, and reworking it to use the typescript swagger API so we can remove some manually created types in favor of the auto generated ones.


prakashsvmx (approved):
Tested. LGTM :+1:

reivaj05 (approved):

PR #2843, opened 2023-05-30
Add tests for remote buckets handlers

reivaj05:


bexsoft (approved):

restapi/admin_remote_buckets_test.go
```diff
@@ -0,0 +1,350 @@
...
+	ctx := context.Background()
+	minioGetRemoteBucketMock = func(_ context.Context, _, _ string) (targets *madmin.BucketTarget, err error) {
+		return &madmin.BucketTarget{
+			Credentials: &madmin.Credentials{
+				AccessKey: "accessKey",
+				SecretKey: "secretKey",
+			},
+		}, nil
+	}
+	res, err := getRemoteBucket(ctx, &suite.adminClient, "bucketName")
```
cesnietor, line 156: can we assert the expected response instead of NotNil? just to make sure we are not returning a blank Structure or something we are not expecting.

reivaj05: No need, getRemoteBucket code will only return a non nil value if there was no error, also all callers of this function handle the case where res is an empty struct

reivaj05: I don't think asserting all the fields adds real value to the test unless we want to assert for a specific value in the field, which is not the case for this test

cesnietor: @reivaj05  what do you really want to test here? how do we make sure the function actually does what is supposed to? (get the response from the client and return it in our desired format) same for `listExternalBuckets`. The value is making sure the function uses the response from the client and returns in the fields that we want. If this test is not intended to do that then we need another one that does this.

reivaj05: @cesnietor  I just want to test that the response is not nil and there was no error, go is a typed language which already makes sure the value the function returns is already in the format we expect `models.RemoteBucket` with the fields we want, there is no data transformation and since i'm mocking the response in the client i know before hand what the values for those fields are going to be in the response so asserting the values would be like `assert.suite(true, true)` which adds no value to the test. Main reason for this test is to increase coverage and catch and fix issues in the code, for example the one i fixed when i was writing these tests in: https://github.com/minio/console/pull/2843/files#diff-59c5813590552d2e24877b71bb6c73db8663f1827416353b2f3f95c4e14a9e75R669

cesnietor: We chatted and I raised that returning `return &models.RemoteBucket{}, nil` would still pass the test. For test coverage is fine. We might have to consider the scope of the unittests since indeed Integration tests could also verify the response format.  Approving and letting you to decide whether we should assert the response fields or not.

cesnietor (approved):

restapi/admin_remote_buckets_test.go
reivaj05: @cesnietor Should be fine for now, going to try to find a way to assert fields in structs without to many lines of code

cesnietor (approved):

bexsoft (approved):

PR #2846, opened 2023-06-01
Hide Lookup Bind Password from the Console UI

cniackz:
### Objective:

To hide `Lookup Bind Password` from the Console UI as suggested by one of our customers.

### How it looks:

Screenshot 2023-06-03 at 7 12 23 AM

### Testing:

1. Tested that LDAP can be configured
2. Tested that LDAP can be used for access
3. Tested that Password is entered but never shown in the UI.

### DIFF
--- a/portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
+++ b/portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
@@ -49,10 +49,16 @@ import TabSelector from "../../Common/TabSelector/TabSelector";
 const enabledConfigLDAP = [
   "server_addr",
   "lookup_bind_dn",
-  "lookup_bind_password",
   "user_dn_search_base_dn",
   "user_dn_search_filter",
 ];
+
+const getFieldValue = (key: string, value: string) => {
+  if (key === "lookup_bind_password") {
+    return value ? "********" : "";
+  }
+  return value ? value : "";
+};
 
 const IDPLDAPConfigurationDetails = () => {
   const dispatch = useAppDispatch();
@@ -405,7 +411,7 @@ const IDPLDAPConfigurationDetails = () => {
                           <LabelValuePair
                             key={key}
                             label={value.label}
-                            value={fields[key] ? fields[key] : ""}
+                            value={getFieldValue(key, fields[key])}
                           />
                         ))}
                       </Fragment>


portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
```diff
@@ -54,6 +54,15 @@ const enabledConfigLDAP = [
   "user_dn_search_filter",
 ];
 
+// To hide Lookup Bind Password in the Console UI
+function isPassword(field: any) {
+  var result = false;
+  if (field == enabledConfigLDAP.at(2)) {
+    result = true;
+  }
+  return result;
```
cesnietor, line 63: no need to have a result variable, we can just do :
```
return field === enabledConfigLDAP.at(2)
```

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
```diff
@@ -54,6 +54,15 @@ const enabledConfigLDAP = [
   "user_dn_search_filter",
 ];
 
+// To hide Lookup Bind Password in the Console UI
+function isPassword(field: any) {
```
cesnietor, line 58: can we rename it as `isLDAPPassword` so that it documents itself.

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
cesnietor: notice the three equal operators for reference see [this](https://www.c-point.com/javascript_tutorial/jsgrpComparison.htm)

cesnietor:
Thinking about it, I think we should just not return it in the API. wdyt @cniackz ? otherwise this looks a bit hacky.

cniackz:
I think you are right @cesnietor we should instead modify the API to avoid retrieving this value.
But this will require more time for me to learn.
Moving this PR as a draft in the meantime.

cniackz:
I don't think `Workflow / Checking Lint (1.20.x, ubuntu-latest) (pull_request) ` is related to this change.

cniackz:
Yeah, confirmed, unrelated: https://github.com/minio/console/pull/2847

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
```diff
@@ -401,13 +405,15 @@ const IDPLDAPConfigurationDetails = () => {
...
                     {hasConfiguration && (
                       <Fragment>
-                        {Object.entries(formFields).map(([key, value]) => (
-                          <LabelValuePair
-                            key={key}
-                            label={value.label}
-                            value={fields[key] ? fields[key] : ""}
-                          />
-                        ))}
+                        {Object.entries(formFieldsWithOutPassword).map(
```
reivaj05, line 408: No need to have an extra variable, if we don't show the password in the details the user might think that there is no password set, it's better if we just display it as "*****"

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
```diff
@@ -401,13 +405,15 @@ const IDPLDAPConfigurationDetails = () => {
...
+                          ([key, value]) => (
+                            <LabelValuePair
+                              key={key}
+                              label={value.label}
+                              value={fields[key] ? fields[key] : ""}
```
reivaj05, line 413: cons getFieldValue = (key, value: string) => {
    if (key === "lookup_bind_password") {
      return value ? "********" : "";
    }
    return value;
  };

value={getFieldValue(fields[key] ? fields[key] : "")}

restapi/admin_config.go
```diff
@@ -148,8 +148,9 @@ func getConfig(ctx context.Context, client MinioAdmin, name string) ([]*models.C
 					Value: kv.EnvOverride.Value,
 				}
 			}
-
-			confkv = append(confkv, &models.ConfigurationKV{Key: kv.Key, Value: kv.Value, EnvOverride: envOverride})
+			if kv.Key != "lookup_bind_password" {
```
reivaj05, line 151: I think we shouldn't modify this since this is mirroring the same behavior as `mc admin config get`, if we are not changing the behavior for mc right now then console should do the same

reivaj05 (changes requested):
see comments

restapi/admin_config.go
cesnietor: It's intended to do this in the future I believe. but I'll  defer this to @cniackz

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
cesnietor: I agree with @reivaj05 , **IF** we can have ldap enabled without password set, then the user should know, so we 
 should display `****` or `Set Password` so that the user knows that it's set or not. If we don't allow (in mc also) to enable ldap without password then no need to display the `Set Password`.

Also the use of the variable might not super convenient cause, everytime we add a key  to the response in the api, or coming from the admin client, we'd need to update that map vs just letting backend drive what should be shown.  Are there any other fields that we want to hide?

restapi/admin_config.go
reivaj05: It's ok, I only meant to actually make the change in console once the change is done in mc

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
cniackz: For now only `Lookup Bind Password` should not be shown in plain text.

portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
cniackz: with a slight modification but it work and I understood your point, thanks Javi!.

restapi/admin_config.go
cniackz: I think `mc` change will take time if approved, I will ask @kannappanr but in the meantime, all we can do is to avoid displaying the password in plan text if provided since API will anyways provide the password, this does not depends on us but is more of a BE change.

allanrogerr:
Fixes 1199

allanrogerr (approved):

allanrogerr (approved):

allanrogerr:
Closing in place of a direct change to minio. PR number pending

Issue #2290, opened 2022-09-01
Pass client real IP address to minio api


victorserbu2709:
It is possible to configure minio console to pass real client IP address in some headers  when doing a request to minio api (eg. x-forwarded-for, x-real-ip)?

PR #2864, opened 2023-06-12
Pass Client IP address to MinIO on x-forwarded-for header

dvaldivia:
This PR introduces a custom `RoundTripper` for the transport we use when communicating with MinIO via the SDKs.

The intention is to communicate the IP address of the actual user to MinIO.

While reviewing this PR:

If the request was available, that was used as an inline solution, like when the client is instantiated from the handler
`context.Context` was used whenever was possible as the middleware can introduce the IP address and have it be pass it around
websocket can't use the same `context.Context` have their own remoteAddress call

Fixes https://github.com/minio/console/issues/2290


restapi/ws_handle.go
```diff
@@ -91,6 +92,7 @@ type WSConn interface {
 	writeMessage(messageType int, data []byte) error
 	close() error
 	readMessage() (messageType int, p []byte, err error)
+	remoteAddress() string
```
cesnietor, line 95: Can we add this one to `mockConn` so that it passes the lint errors.

dvaldivia: added

cmd/console/app_commands.go
```diff
@@ -41,8 +41,8 @@ func StartServer(ctx *cli.Context) error {
 	}
 
 	xctx := context.Background()
-	transport := restapi.PrepareSTSClientTransport(false)
-	if err := logger.InitializeLogger(xctx, transport); err != nil {
+	transport := restapi.PrepareSTSClientTransport(false, "127.0.0.1")
```
reivaj05, line 44: Could you add "127.0.0.1" as constant in `const.go` file?

reivaj05 (changes requested):

cmd/console/app_commands.go
dvaldivia: done

bexsoft (approved):

restapi/tls.go
```diff
@@ -23,8 +23,19 @@ import (
 	"time"
 )
 
+type ConsoleTransport struct {
+	Transport *http.Transport
+	ClientIP  string
+}
+
+func (t *ConsoleTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Add("X-Forwarded-For", t.ClientIP)
```
vadmeste, line 32: Do we really need to use X-Forwarded-For header? it is a known header that has its own purpose and Console is not a real proxy, I guess we should be able to use a different header name without issues.

dvaldivia: we are technically proxying requests for the users via the UI, and we need to communicate to MinIO the real IP address of the client behind the browser and MinIO is already reading this header by default so I don't think we should be using a custom header

reivaj05 (approved):

PR #2868, opened 2023-06-13
Refactor session to avoid duplicate calls to apis

cesnietor:
### Description:
Refactors the way we were calling `/session` api since previous implementaiton was making multiple requests (up to six) which could mean high latency if that requests take time to resolve. Now we call it only once and only at initial load. 

### Changes:
- [x] created Session thunk to control Session API state 
- [x] Create new state in consoleSlice to store the session state (loading, done, etc)
- [x] Updated Loging.tsx to default to `/browser` to avoid redirecting first to `/` and then to `/browser` when logging in.  

Before:
Screenshot 2023-06-12 at 10 50 46 PM

After:
Screenshot 2023-06-12 at 11 05 27 PM

If navigating between tabs it won't call session api.
Screenshot 2023-06-13 at 1 59 49 PM

### Test Steps:
Scenarios:
- With a minio running on port 9000, make assets and then make the binary `make assets && make` then run it on port 9090 like:
```
CONSOLE_ACCESS_KEY=minioconsole
CONSOLE_SECRET_KEY=minioconsole
CONSOLE_MINIO_SERVER=http://localhost:9000
CONSOLE_DEV_MODE=on ./console server
```
- Can login with valid user and the amount of calls to `/session` api are not more than one.
- Navigate between tabs (we should call `/session` only ONCE at initial load of whole page)
- Anonymous access works without issues (create bucket, set anonymous access to it  [path `/`] and try to access to it without loging in). Everything should work just fine.

### DIFF
--- a/portal-ui/src/ProtectedRoutes.tsx
+++ b/portal-ui/src/ProtectedRoutes.tsx
@@ -14,25 +14,18 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import React, { useEffect, useState } from "react";
+import { useEffect, useState } from "react";
 import { Navigate, useLocation } from "react-router-dom";
 import useApi from "./screens/Console/Common/Hooks/useApi";
 import { ErrorResponseHandler } from "./common/types";
 import { ReplicationSite } from "./screens/Console/Configurations/SiteReplication/SiteReplication";
 import { useSelector } from "react-redux";
-import {
-  globalSetDistributedSetup,
-  setAnonymousMode,
-  setOverrideStyles,
-  setSiteReplicationInfo,
-  userLogged,
-} from "./systemSlice";
 import { SRInfoStateType } from "./types";
 import { AppState, useAppDispatch } from "./store";
-import { saveSessionResponse } from "./screens/Console/consoleSlice";
-import { getOverrideColorVariants } from "./utils/stylesUtils";
 import LoadingComponent from "./common/LoadingComponent";
-import { api } from "api";
+import { fetchSession } from "./screens/LoginPage/sessionThunk";
+import { setSiteReplicationInfo } from "./systemSlice";
+import { SessionCallStates } from "./screens/Console/consoleSlice";
 
 interface ProtectedRouteProps {
   Component: any;
@@ -41,8 +34,14 @@ interface ProtectedRouteProps {
 const ProtectedRoute = ({ Component }: ProtectedRouteProps) => {
   const dispatch = useAppDispatch();
 
-  const [sessionLoading, setSessionLoading] = useState<boolean>(true);
   const userLoggedIn = useSelector((state: AppState) => state.system.loggedIn);
+  // sessionLoading is used to control rendering this page, if we only use console.sessionLoadingState,
+  // since the state chages it triggers the reload of this component multiple times, causing duplicate queries.
+  const [sessionLoading, setSessionLoading] = useState<boolean>(true);
+  const sessionLoadingState = useSelector(
+    (state: AppState) => state.console.sessionLoadingState
+  );
+
   const anonymousMode = useSelector(
     (state: AppState) => state.system.anonymousMode
   );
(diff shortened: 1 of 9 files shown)


portal-ui/src/ProtectedRoutes.tsx
```diff
@@ -53,56 +52,16 @@ const ProtectedRoute = ({ Component }: ProtectedRouteProps) => {
     return <Navigate to={{ pathname: `login` }} />;
   };
 
-  const pathnameParts = pathname.split("/");
-  const screen = pathnameParts.length > 2 ? pathnameParts[1] : "";
+  // update location path and store it
+  useEffect(() => {
+    dispatch(fetchSession(pathname));
+  }, [dispatch, pathname]);
```
dvaldivia, line 58: This effect will cause us to check the session every time the path changes, if we already checked the session we shouldn't need to check it again

portal-ui/src/screens/Console/consoleSlice.ts
```diff
@@ -15,40 +15,56 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+import { SessionResponse } from "../../api/consoleApi";
 import { AppState } from "../../store";
-import { SessionResponse } from "api/consoleApi";
+import { fetchSession } from "../../screens/LoginPage/sessionThunk";
+
+export enum SessionCallStates {
```
dvaldivia, line 22: can we put this in `consoleSlice.types.ts` to keep the slice simple?

portal-ui/src/screens/LoginPage/sessionThunk.ts
```diff
@@ -0,0 +1,95 @@
...
+            .then(() => {
+              dispatch(setAnonymousMode());
+            })
+            .catch((res: HttpResponse<ListObjectsResponse, Error>) => {
+              dispatch(setErrorSnackMessage(errorToHandler(res.error)));
+            })
+            .finally(() => {
+              // TODO: we probably need a thunk for this api since setting the state here is hacky,
+              // we can use a state to let the ProtectedRoutes know when to render the elements
+              dispatch(setSessionLoadingState(SessionCallStates.Done));
```
dvaldivia, line 86: do you really need to dispatch this? wouldn't the slice take care of this?

dvaldivia (changes requested):

portal-ui/src/ProtectedRoutes.tsx
cesnietor: I thought about that but if we are navigating through the app and you are logged in, what if someone wants to restrict your access while logged in, so calling it on every pathname would allow that. That's how it was before so assumed that was expected and kind of made sense.

portal-ui/src/screens/Console/consoleSlice.ts
cesnietor: we have types below but yes, I can bring them to a separate file, usually is better to keep related code closer to the code that uses it but we can have a types file.

cesnietor: Actually I'd like to keep it there, is still simple don't you think?

dvaldivia:
Seeing a session call every time I navigate @cesnietor 
Screenshot 2023-06-13 at 9 21 21 AM

cesnietor:
@dvaldivia yes it was like that already before.

portal-ui/src/screens/Console/consoleSlice.ts
dvaldivia: I think it's better to keep it in a separate file like all the types we have everywhere else, slices tend to get large :P

portal-ui/src/screens/LoginPage/Login.tsx
```diff
@@ -36,7 +36,8 @@ export interface LoginStrategyPayload {
 }
 
 export const getTargetPath = () => {
-  let targetPath = "/";
+  // default to browser view to avoid redirecting first to "/" and then to "/browser" on login
```
reivaj05, line 39: Comment might not be necessary

reivaj05 (approved):

portal-ui/src/ProtectedRoutes.tsx
cesnietor: We discussed this and indeed is a nice "feature" that we have where we can hide the features they have access if we for some reason want to. But this is something we will sacrifice since calling /session api everytime the path changes is expensive.  
I've updated the PR to reflect this. It will now only call session at initial load. Anonymous access still works.

portal-ui/src/screens/Console/consoleSlice.ts
cesnietor: done.

portal-ui/src/ProtectedRoutes.tsx
```diff
@@ -41,8 +34,14 @@ interface ProtectedRouteProps {
 const ProtectedRoute = ({ Component }: ProtectedRouteProps) => {
   const dispatch = useAppDispatch();
 
-  const [sessionLoading, setSessionLoading] = useState<boolean>(true);
   const userLoggedIn = useSelector((state: AppState) => state.system.loggedIn);
+  // sessionLoading is used to control rendering this page, if we only use console.sessionLoadingState,
+  // since the state chages it triggers the reload of this component multiple times, causing duplicate queries.
+  const [sessionLoading, setSessionLoading] = useState<boolean>(true);
```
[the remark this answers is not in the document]
cesnietor: agreed, done.

cesnietor:
@prakashsvmx @reivaj05 @dvaldivia @jinapurapu  I've updated the PR addressing the comments and also doing the following -> We will no longer call `/session` api on every change path since it's a bit expensive (even if the api resolves quickly). More comments in the thread above with Daniel

reivaj05 (approved):

cesnietor:
rebased.

dvaldivia (approved):

PR #2889, opened 2023-06-16
feat(ObjectManager): virtualize download/upload object list

### LINKED ISSUE #2878: Virtualize the "Downloads/Uploads" popup's items
I recently had to upload a large batch of image(around 2000), and noticed this popup's item list is not virtualized, hence is very laggy and memory consuming and makes the entire UI unresponsive.

image

I'd be happy to send a PR

ali-garajian:
closes #2878


cesnietor (approved):
Tested with 500 files of 40M size and component no longer lags.
Thanks for the contribution.
LGTM.

bexsoft (approved):
LGTM, There is a small issue with the last item in the list where is not completely visible (99%), this is not related to this change, will be fixed after migrating this panel to mds.

Issue #2890, opened 2023-06-16
'Cannot read properties of null (reading 'close')' issue while downloading folders


bexsoft:
## Description

There is an issue while downloading multiple files or folders in Object browser:

<img width="432" alt="Screenshot 2023-06-16 at 12 16 04" src="https://github.com/minio/console/assets/33497058/c5ca4c56-bd71-44f9-a3be-7e23e6579b08">
<img width="1394" alt="Screenshot 2023-06-16 at 12 10 28" src="https://github.com/minio/console/assets/33497058/0fd59ce4-990e-4bcf-b64e-a443e349c401">

## How to reproduce

1. Go into a bucket with a folder in it
2. Click on the checkbox next to the folder name and click download
3. Expand Download manager. Issue is seen here

------
1. Go into a bucket with multiple files (more tan 100)
2. Click on the select all checkbox and click download button
3. A failure may be seen and multiple files will start to fail during download

## Expected Behavior

No errors should be seen and files should get downloaded correctly. Folders should be downloaded as zip files.

cesnietor:
Might be related to https://github.com/minio/console/pull/2773

cesnietor:
Code is panicking here https://github.com/minio/console/blob/master/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts#LL182C10-L182C10

dhzoo30:
Hi, I faced the same issue. I used latest version (v0.30.0). It happens regardless of file size.

PR #2891, opened 2023-06-16
Fix download of folders

reivaj05:
Fixes: https://github.com/minio/console/issues/2890

Folders don't have a size attribute when they are retrieved from minio, i believe this is for design so new way to download in chunks won't work in this case, for the moment i'll let the browser handle the download as we do it for safari and later i'll check minio code if there is a way to send the size for folders to console

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
+++ b/portal-ui/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
@@ -112,7 +112,9 @@ class DownloadHelper {
     let isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
     if (isSafari) {
       this.toastCallback();
-      this.downloadSafari();
+      this.downloadByBrowser();
+    } else if (!this.fileSize) {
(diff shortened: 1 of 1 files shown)


cesnietor (approved):
Tested LGTM, thanks for the fix.

cesnietor:
@reivaj05  I just noticed that on Chrome the snackBar message doesn't show up like it does on Safari when downloading the folder.

reivaj05:
@cesnietor That's right, it doesn't show as in Safari, i was't sure if we want to display it in this case, Safari always downloads with the browser so we display the message all the time, it can be added if we want the same behavior for this case

cc: @dvaldivia

cesnietor:
@reivaj05  yeah it indeed might not be needed since Chrome has a more visible indication on what's going on.
Screenshot 2023-06-16 at 2 24 49 PM

dvaldivia (approved):

kaankabalak (approved):

prakashsvmx (approved):

bexsoft (approved):

PR #2899, opened 2023-06-21
Added placeholder image to HelpMenu if links unavailable

jinapurapu:
Adds placeholder image with text describing the Help Menu status when outside image links are not available.

https://github.com/minio/console/assets/65002498/187e417f-6205-459b-8d0a-10f0601c20f5


cesnietor:
If there is no internet, they won't be able to go to the links (from `Learn More`) so what about disabling the whole help menu? Do we have info that can be shown without internet?

jinapurapu:
> If there is no internet, they won't be able to go to the links (from `Learn More`), which would make the user experience confusing, so what about disabling the whole help menu? Or do we have info that can be shown without internet?

Yes, that is why the text explains this for now. I am still working on disabling the Help Menu or replacing with a limited offline mode if the links can't be reached.

reivaj05 (approved):

allanrogerr (approved):

prakashsvmx (approved):

PR #2913, opened 2023-06-29
Use github.com/minio/pkg/v2

donatello:


cesnietor:
@donatello  is this still needed?

donatello:
We can close for now

PR #2922, opened 2023-07-05
Removing white spaces from provided values

cniackz:
To fix: https://github.com/minio/console/issues/2755

### Explanation:

All I am doing is to trim the spaces introduced by the user so that we avoid an invalid configuration:

```
API: SYSTEM()
Time: 14:05:47 UTC 06/30/2023
DeploymentID: ad79aacc-3ffb-4726-9ff3-6cc7ca58b27a
Error: Invalid site configuration: region ' us-west-rack-2' is invalid, expected simple characters such as [us-east-1, myregion...] (*fmt.wrapError)
       7: internal/logger/logger.go:258:logger.LogIf()
       6: cmd/config-current.go:493:cmd.lookupConfigs()
       5: cmd/config.go:218:cmd.initConfig()
       4: cmd/config.go:189:cmd.(*ConfigSys).Init()
       3: cmd/server-main.go:452:cmd.initConfigSubsystem()
       2: cmd/server-main.go:423:cmd.initServerConfig()
       1: cmd/server-main.go:662:cmd.serverMain()
```

I think this can be generic as in non of our configurations we need leading or trailing spaces to work, so trim seems to be the solution here.

### DIFF
--- a/portal-ui/src/screens/Console/EventDestinations/ConfTargetGeneric.tsx
+++ b/portal-ui/src/screens/Console/EventDestinations/ConfTargetGeneric.tsx
@@ -98,6 +98,7 @@ const ConfTargetGeneric = ({
 
   const setValueElement = (key: string, value: string, index: number) => {
     const valuesDup = [...valueHolder];
+    value = value.trim();
     valuesDup[index] = { key, value };
 
     setValueHolder(valuesDup);


portal-ui/src/screens/Console/Configurations/utils.tsx
```diff
@@ -401,8 +401,12 @@ export const fieldsConfigurations: any = {
 };
 
 export const removeEmptyFields = (formFields: IElementValue[]) => {
+  // Remove white spaces from strings
+  for (let i = 0; i < formFields.length; i++) {
+    // @ts-ignore
+    formFields.at(i).value = formFields.at(i).value.trim();
```
cesnietor, line 407: since we are iterating here already in the array, we can do the filter (set the valid array of fields) here  cause the `filter` below iterates again on it so we can save some computation there.

portal-ui/src/screens/Console/Configurations/utils.tsx
```diff
@@ -401,8 +401,12 @@ export const fieldsConfigurations: any = {
 };
 
 export const removeEmptyFields = (formFields: IElementValue[]) => {
```
reivaj05, line 403: It might be better to change  the file `console/portal-ui/src/screens/Console/EventDestinations/ConfTargetGeneric.tsx:103`  to `valuesDup[index] = { key, value.trim() };
`

I believe that's the place where the value is being stored

reivaj05 (commented):
see comment

portal-ui/src/screens/Console/Configurations/utils.tsx
bexsoft, line 407: This can be changed to be a map, and as @cesnietor says, the filter bellow can be applied here too.

harshavardhana (dismissed):

portal-ui/src/screens/Console/Configurations/utils.tsx
cniackz: This is a great idea @reivaj05, I just tested this change and it works perfectly!

harshavardhana (approved):

cesnietor (approved):

reivaj05 (approved):

bexsoft (approved):

PR #2927, opened 2023-07-06
Fixed subnet Health report upload 

jinapurapu:
Adds functions to get API key from Subnet using license key, to correctly generate SUBNET upload headers, process Health report for upload, generic subnet upload. 
Also adds UI improvements on HealthInfo screen to display error if empty Health report is returned, prompt user to retry failed reports.
Screenshot 2023-07-06 at 4 18 55 PM


restapi/admin_health_info.go
```diff
@@ -142,9 +139,17 @@ func sendHealthInfoToSubnet(ctx context.Context, healthInfo interface{}, client
 	if e != nil {
 		return "", e
 	}
-	apiKey := subnetTokenConfig.APIKey
+	apiKey, e := subnet.GetSubnetAPIKeyUsingLicense(subnetTokenConfig.License)
```
anjalshireesh, line 142: Maybe this should be done only if `subnetTokenConfig.APIKey` is empty?

restapi/client-admin.go
```diff
@@ -413,7 +411,10 @@ func (ac AdminClient) serverHealthInfo(ctx context.Context, healthDataTypes []ma
 		} else {
 			info.Minio.Info = minioInfo
 		}
-
+		healthV1 := mcCmd.MapHealthInfoToV1(info, nil)
```
anjalshireesh, line 414: Since this will be embedded with latest `minio`, you'll never have to deal with older versions of the health info. The version will always be `madmin.HealthInfoVersion`

`mc` has this kind of code because it needs to support older versions of `minio`.

pkg/subnet/utils.go
```diff
@@ -94,29 +107,33 @@ func subnetUploadReq(info interface{}, url string, filename string) (*http.Reque
...
 	}
 
 	if e := enc.Encode(info); e != nil {
-		return nil, e
+		return nil, "", e
 	}
 	zipWriter.Close()
 	temp := body
-	part, e := writer.CreateFormFile("file", filename)
+	part, e := writer.CreateFormFile("file", "filename")
```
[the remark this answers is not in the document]
jinapurapu: It's only a temporary "file" to make the correct header, I don't think the name is used anywhere else

reivaj05 (dismissed):

bexsoft (changes requested):
Behavior is not consistent all the times, some random errors are seen when you click on the generate diagnostics button

cesnietor:
@jinapurapu  how can I test this? seems I need to register my cluster but this is local. do you have some creds? please share privately.

bexsoft (approved):
LGTM

cesnietor (approved):
tested with @bexsoft lgtm

PR #2928, opened 2023-07-07
fix object metadata section crash due to un escaped characters

prakashsvmx:
Closes https://github.com/minio/console/issues/2877

fix object metadata section crash due to un escaped characters

upload an object with meta data containing special chars. like 
```js
{
        mymeta:"test%special-sybols@~`@#$%^&*()+-_=<>,.?/||\/",
        "more":"test",
        "test1":"test2"
    }
```

while viewing the object in console, it should not crash.

Result:
image


jiuker (approved):

bexsoft (approved):
LGTM

PR #2940, opened 2023-07-12
[DO-NOT-MERGE] govluncheck

bexsoft:


harshavardhana (changes requested):
it is better to split the `vulncheck` into a separate yaml - there are way too many changes here.

See here https://github.com/minio/minio/blob/master/.github/workflows/vulncheck.yml

bexsoft:
> it is better to split the `vulncheck` into a separate yaml - there are way too many changes here.
> 
> See here https://github.com/minio/minio/blob/master/.github/workflows/vulncheck.yml

Ok, will separate it, do you know if there is a way to updated the cached version of go for the workflows? If I return back to 1.20.x it automatically enforce v 1.20.5

bexsoft:
Created new branch & PR

Issue #2942, opened 2023-07-13
Base64-encoded url path works wrong in browser


andy812:
## NOTE
What i see in my browser:
https://my-minio.com/browser/archi/cHVibGljL2Nzcy9pMThuLw==
```
echo  "cHVibGljL2Nzcy9pMThuLw==" | base64 -d
public/css/i18n/
```
It means i see files in mybucket/public/css/i18n folder. 
When i share the link https://my-minio.com/browser/archi/cHVibGljL2Nzcy9pMThuLw== with someone, they see only bucket's root folder, not ./public/css/i18n/
<img width="286" alt="изображение" src="https://github.com/minio/console/assets/29895539/21d0b8be-9e65-4719-bbd2-b597a9ee6add">

## Expected Behavior
You should see content of ./public/css/i18n/ folder when you open https://my-minio.com/browser/archi/cHVibGljL2Nzcy9pMThuLw==

## Your Environment
RELEASE.2023-07-07T07-13-57Z

prakashsvmx:
Thank you for reporting. @andy812  :+1:

PR #2944, opened 2023-07-14
fix loading of objects at a path when url is shared opened elsewhere

prakashsvmx:
Fixes https://github.com/minio/console/issues/2942

When a path is shared and opened in a new browser instance, only the top level was loaded.

with this fix, the shared path would load as expected.


kaankabalak (approved):
LGTM

harshavardhana (approved):

PR #2965, opened 2023-07-27
feat: download multiple object selection as zip ignoring any deleted objects selected

prakashsvmx:
Current implementation:
Currently if the user selects multiple files and clicks download, each file is download individually.

This PR enhances the download experience.
When multiple objects/prefixes are selected and download is clicked  perform the download as zip.

This PR is inspired by the prefix download as zip

Note: 
It would ignore deleted objects if any selected via "Show deleted objects" option


cesnietor:
What issue is this solving @prakashsvmx ?
adding @reivaj05  as a reviewer since he did some work on the download feature.

prakashsvmx:
@cesnietor 
the current implementation of download makes multiple files to download individually.
<details>
<summary> Video of current multi select download implementation</summary>

https://github.com/minio/console/assets/23444145/4309696a-2461-48fc-868d-5ef07790db84

</details>

With this implementation, if user selects multiple objects and clicks download, all selected objects would be download as single zip.

If the selection contains any prefix, it would be download as a separate zip.

<details>

<summary> New Download current multi select download implementation </summary>

https://github.com/minio/console/assets/23444145/02afb739-bf31-4908-bf57-e94376705d30



</details>

harshavardhana:
should we not be downloading all files in a single zip?

prakashsvmx:
@harshavardhana  the prefixes can be large so i have taken this approach to download them individually.

harshavardhana:
> @harshavardhana the prefixes can be large so i have taken this approach to download them individually.
> 
> Also it would require listing. as the client ( console ui) does not have the objects list under a prefix.
> 
> Currently the prefixes are downloaded as zip.

if we are going to do this on client side for prefix, I think doing this on server side is better.

prakashsvmx:
Thank you for the guidance @harshavardhana .
I will explore this further.

harshavardhana:
So basically ask MinIO to give you a zip result "streaming" result @prakashsvmx

bexsoft (commented):
Tested, it works fine, It would be nice to display the download progress in the Upload / Download manager as well

cesnietor:
Thanks for adding tests

cesnietor (approved):
LGTM

bexsoft (approved):

PR #2972, opened 2023-07-31
Update Github jobs to use hardcoded testcafe version

cesnietor:
Updated jobs to use raw command for testcafe instead of using github-action since that one uses latest version of testcafe which has a known [issue](https://github.com/DevExpress/testcafe/issues/7898) for headless chrome which seems to have been introduced in v3.1.0.


bexsoft (approved):
LGTM

reivaj05 (approved):

PR #2978, opened 2023-08-03
Initial Tools Pages Migrations

bexsoft:
## What does this do?

Migrates the following pages to mds:

- Call Home
- Inspect
- Profile
- Health

## How does it look?
Screenshot 2023-08-02 at 18 23 21
Screenshot 2023-08-02 at 18 23 15
Screenshot 2023-08-02 at 18 23 10
Screenshot 2023-08-02 at 18 23 04

### DIFF
--- a/portal-ui/src/screens/Console/Common/TestWrapper/TestWrapper.tsx
+++ b/portal-ui/src/screens/Console/Common/TestWrapper/TestWrapper.tsx
@@ -15,120 +15,20 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import React, { Fragment, useEffect, useState } from "react";
-import { Theme } from "@mui/material/styles";
-import createStyles from "@mui/styles/createStyles";
-import withStyles from "@mui/styles/withStyles";
-import { Grid } from "@mui/material";
-import ArrowForwardIosIcon from "@mui/icons-material/ArrowForwardIos";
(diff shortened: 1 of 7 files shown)


cesnietor:
Screenshot 2023-08-03 at 11 09 12 AM
Screenshot 2023-08-03 at 11 09 04 AM
Button `Save Configuration` in Call Home  looks disable but it's still clickable and when clicked it calls and returns a 500

cesnietor:
Inspect redirects you to register your closer if you haven't registered when you fill the data and click `Inspect` is that expected? It's not clear that the feature is only available if the cluster is registered.

bexsoft:
> Inspect redirects you to register your closer if you haven't registered when you fill the data and click `Inspect` is that expected? It's not clear that the feature is only available if the cluster is registered.
> 
> Same for Profile

Yes it is expected, there is a note at the top of the page asking to register your cluster

bexsoft:
> Screenshot 2023-08-03 at 11 09 12 AM Screenshot 2023-08-03 at 11 09 04 AM Button `Save Configuration` in Call Home looks disable but it's still clickable and when clicked it calls and returns a 500

This should redirect you to Register your cluster as well

cesnietor (approved):
As discussed internally the features that require cluster registration should be disabled completely so that it's clear that they need to register the cluster. We can either merge this change for now since that is the current behavior and add as an improvement in another one or fix here.

pjuarezd (approved):

Issue #2967, opened 2023-07-27
Console appears to not save LDAP config, always missing bind pw


feorlen:
Configuring LDAP from Console doesn't appear to succeed because the Lookup Bind Password isn't saved. LDAP always shows as not configured. 

But `mc admin config get myminio identity_ldap` shows the expected configuration and LDAP works.

Tested with the LDAP server from Will Dinyes' `dinyes.co` IAM lab, which uses `mc`. Will has the same problem with Console, so it's not only my test VM.

## Expected Behavior

Console saves the LDAP configuration and shows LDAP enabled.

## Current Behavior

All the other settings are saved, but Console always says the Lookup Bind Password is missing. "LDAP Enabled" is always "No" even when LDAP login works.

## Steps to Reproduce (for bugs)

1. Install SNSD following docs.
1. Log in to Console as admin.
1. Configure LDAP, restart.
1. Log out and return to login page.
1. No LDAP shown in Other Authentication Methods.
1. Log in as admin, view LDAP configuration.
1. Bind password is missing and GUI reports LDAP is not enabled.
1. In terminal, `mc` shows expected configuration.
1. With `mc`, able to attach consoleAdmin policy to an LDAP user and login.

```
$ mc admin config get myminio identity_ldap
identity_ldap server_addr=blahblah.dinyes.co:389 srv_record_name= user_dn_search_base_dn=dc=minio,dc=io user_dn_search_filter=(cn=%s) group_search_filter= group_search_base_dn= tls_skip_verify=off server_insecure=on server_starttls=off lookup_bind_dn=cn=admin,dc=minio,dc=io 
```

## Your Environment

Ubuntu VM on Ubuntu host with KVM/Virtual Machine Manager

Guest OS:
Ubuntu 22.04.2 LTS (GNU/Linux 5.15.0-78-generic x86_64)

MinIO:
https://dl.min.io/server/minio/release/linux-amd64/archive/minio_20230718174940.0.0_amd64.deb

Terminal session from MinIO install
[ldap-console-bug-scrollback-annotated.txt](https://github.com/minio/console/files/12188411/ldap-console-bug-scrollback-annotated.txt)

Video of Console login and reproduction available privately (contains live URLS, etc)

cesnietor:
This might be caused by a recent change done in minio https://github.com/minio/minio/pull/17380, we'll try to reproduce.

cniackz:
[...]

PR #2983, opened 2023-08-04
Removing unneeded password

cniackz:
fixes: https://github.com/minio/console/issues/2967
### Objective:

To fix our LDAP configuration page: https://github.com/minio/console/issues/2967

### Reason:

Since change https://github.com/minio/minio/pull/17380 we are not longer receiving the password field due to security reasons. Hence the UI is failing because is still expecting it. If we remove it; then the coincidences will be same and page rendered. Already tested in my local! 👍


cesnietor (approved):
LGTM

harshavardhana (approved):

cesnietor:
I wonder if we could face gracefully, meaning, if key is not available then just don't show that value or show an `N/A` instead of breaking the whole section. Cause another key could suddenly be removed and would be breaking this again.

cniackz:
True, code can be enhanced. Let's do that on a new/separate PR. Here the intention is just to update the code for the new MinIO response without any change to the current logic.

PR #2987, opened 2023-08-09
truncate prefixes above the selected level to make the download layout flat

prakashsvmx:
truncate prefixes above the selected level to make the download layout flat

Since objects are selected at nested level, we could ignore the top level prefixes when zip is created to make the layout flat.


cesnietor:
Tested LGTM, objects are now being downloaded in their corresponding path as expected.
Screenshot 2023-08-08 at 9 32 25 AM

pjuarezd (approved):

Issue #2848, opened 2023-06-02
Service Account doesn't show status 


cesnietor:
If a Service Account is disabled e.g. via `UpdateServiceAccount(ctx, resource, opts)` api, we list them in the UI but there is no way to know whether a svcacct is enabled or disabled, also there is no way to change their status.

Steps:
- Create a user
- Create a svcacct
- Disable it via mc admin like:
```
mc admin user svcacct disable myminio myuserserviceaccount
```
- The ui still shows it but there is no way to know it's status

![Screenshot 2023-06-02 at 11 49 39 AM](https://github.com/minio/console/assets/11819101/9b016082-0cf3-4077-a4aa-2b853ab33652)

PR #2992, opened 2023-08-14
show service account status and expiry in ui

prakashsvmx:
show service account status and expiry in ui
Fixes https://github.com/minio/console/issues/2848

How does it look?
image

image

Note: There is no UI to update `expiry`, `name`  ,`description` and `status` . for now it is part of `mc` commands

### DIFF
--- a/models/service_accounts.go
+++ b/models/service_accounts.go
@@ -24,21 +24,116 @@ package models
 
 import (
 	"context"
+	"strconv"
 
+	"github.com/go-openapi/errors"
 	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
 )
 
 // ServiceAccounts service accounts
 //
 // swagger:model serviceAccounts
-type ServiceAccounts []string
+type ServiceAccounts []*ServiceAccountsItems0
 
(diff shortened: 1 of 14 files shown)


models/service_accounts.go
```diff
@@ -24,21 +24,116 @@ package models
...
+	"github.com/go-openapi/errors"
 	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
 )
 
 // ServiceAccounts service accounts
 //
 // swagger:model serviceAccounts
-type ServiceAccounts []string
+type ServiceAccounts []*ServiceAccountsItems0
```
kaankabalak, line 37: Is the 0 at the end expected for the type here? Or was this previously a placeholder which got generated from the swagger file?

prakashsvmx: Yes. It was auto generated by `make swagger-gen`

cesnietor (approved):
Tested disabling svcacct and status is being reflected.

jinapurapu (approved):

bexsoft (approved):
LGTM, `string[]` change in PolicyDetails.tsx is fine

PR #2993, opened 2023-08-14
Remove duplicate ref=con in License page buttons

kaankabalak:
The getButton function already handles appending the query parameters, therefore the linkTracker variable isn't needed


prakashsvmx (approved):

bexsoft (approved):
LGTM

PR #3017, opened 2023-08-25
[WIP] Use absolute path

cniackz:
### Objective:

To use absolute path in our imports as observed in: https://towardsthecloud.com/set-absolute-path-typescript

### DIFF
--- a/portal-ui/src/screens/Console/Buckets/BucketDetails/AccessDetailsPanel.tsx
+++ b/portal-ui/src/screens/Console/Buckets/BucketDetails/AccessDetailsPanel.tsx
@@ -24,16 +24,13 @@ import {
   CONSOLE_UI_RESOURCE,
   IAM_PAGES,
   IAM_SCOPES,
-} from "../../../../common/SecureComponent/permissions";
-import {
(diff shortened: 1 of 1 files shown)


cniackz:
Let's see if tests passes, if you like the idea, I can keep on adding more files to use only absolute paths instead.

jinapurapu (approved):
LGTM, this will simplify imports quite a bit!

prakashsvmx:
`tsconfig.json`  has the below config which enables this support. so we should be good to use  absolute imports :+1:   
@cniackz 

```
....
"baseUrl": "./src",
    "rootDir": "./src"
  },
  "include": ["src"]
```

prakashsvmx (approved):

cesnietor:
there is a pattern commonly followed by others which is doing the relative paths. I'd like this to be discussed with @bexsoft before we merge.

bexsoft (approved):
LGTM, need to review HOP use case

cniackz:
I don't really want to break HOP for this minor changes. Will keep code as is.

PR #3022, opened 2023-08-30
idp: Use 900 seconds as minimum expiry without returning an error

vadmeste:
Do not bother the user with an error if the IDP expiry is less than 
900 seconds, since the S3 spec sets a minimum of 900 seconds 
for STS expiration, use that minimum duration instead of returning 
an error


harshavardhana:
We should take this in since this fixes a regression @bexsoft @cesnietor - can we defer the minor change to move into a constant for later?

prakashsvmx (approved):

harshavardhana (approved):

PR #3044, opened 2023-09-12
Load the available CA in logout API

pjuarezd:
## What does this do?
Load the available CA in the  API when invoking the `openid-connect/logout` remote IDP endpoint.

Found that even though the CA certificates are properly provided to minio console, the invoke to the logout endpoint when using OIDP was not loading the CA certificates, causing several things:

1) API `/api/v1/logout` returned a 500 error.

2) API call returned `tls: failed to verify certificate: x509: certificate signed by unknown authority` error:

```json
{
    "detailedMessage": "Post \"https://kc-service.keycloak.svc.cluster.local/realms/oidc-iam/protocol/openid-connect/logout\": tls: failed to verify certificate: x509: certificate signed by unknown authority",
    "message": "an error occurred, please try again"
}
```

3) User session on IDP was not closed 

4) User session in Minio Console was not really closed, user is still logged in

idp no logout

### DIFF
--- a/restapi/user_logout.go
+++ b/restapi/user_logout.go
@@ -18,6 +18,7 @@ package restapi
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"net/http"
@@ -37,7 +38,7 @@ func registerLogoutHandlers(api *operations.ConsoleAPI) {
 	api.AuthLogoutHandler = authApi.LogoutHandlerFunc(func(params authApi.LogoutParams, session *models.Principal) middleware.Responder {
 		err := getLogoutResponse(session, params)
 		if err != nil {
-			return authApi.NewLogoutDefault(err.Code).WithPayload(err.APIError)
+			api.Logger("IDP logout failed: %v", err.APIError)
 		}
 		// Custom response writer to expire the session cookies
 		return middleware.ResponderFunc(func(w http.ResponseWriter, p runtime.Producer) {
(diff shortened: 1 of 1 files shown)


dvaldivia (approved):

restapi/user_logout.go
```diff
@@ -101,7 +102,14 @@ func logoutFromIDPProvider(r *http.Request, state string) error {
...
 		params.Add("refresh_token", refreshToken.Value)
-		_, err := http.PostForm(providerCfg.EndSessionEndpoint, params)
+		client := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					RootCAs: GlobalRootCAs,
+				},
+			},
+		}
+		_, err := client.PostForm(providerCfg.EndSessionEndpoint, params)
```
harshavardhana, line 112: Actually, there is no point in throwing an error here since it's an idempotent behavior, that was the main reason.

Logout is one way action so it is simply safe to not even bother with the error in http.PostForm() you can simply invalidate the cookie anyways that would force a re-login

Some HTTP API calls are idempotent such as DELETE in S3 API which never returns an error, logout is similar.

harshavardhana (changes requested):

restapi/user_logout.go
harshavardhana: We do not really care if the user session from remote IDP is not closed or not, that is up to them to manage inactive sessions.

The main reason is even a simple network error will do that too, IMO if we say something to indicate that "IDP handshake failed, logging out locally to allow for re-login" would work if you want to let them know why.

IMO we do not have to throw any error, no one would notice.

harshavardhana (changes requested):

restapi/user_logout.go
pjuarezd: Shuld we out a log line to the console stdout error instead?

restapi/user_logout.go
```diff
@@ -37,7 +38,7 @@ func registerLogoutHandlers(api *operations.ConsoleAPI) {
 	api.AuthLogoutHandler = authApi.LogoutHandlerFunc(func(params authApi.LogoutParams, session *models.Principal) middleware.Responder {
 		err := getLogoutResponse(session, params)
 		if err != nil {
-			return authApi.NewLogoutDefault(err.Code).WithPayload(err.APIError)
+			api.Logger("IDP logout failed: %v", err.APIError)
```
pjuarezd, line 41: Sending the error to api logger instead of prevent the cookies clean here @harshavardhana

harshavardhana (approved):

reivaj05 (approved):

harshavardhana (approved):

bexsoft (approved):
LGTM

PR #3047, opened 2023-09-13
Improvements to LDAP Configuration page

### LINKED ISSUE #3028: LDAP configuration acts like it saves, but then reverts after refresh
I think I could debug this if I could find the logs but can't seem to find where Minio keeps LDAP logging. 

No matter, when I configure the settings, and hit Save. the configuration saves, shows the settings, but when I click refresh it shows "LDAP Enabled No".

Click on Identity --> LDAP

I configure Minio LDAP to my LDAP server:
Server Insecure: Enabled
Server Address: freeipa.example.com:389
Lookup Bind DN: uid=miniobind,cn=users,cn=accounts,dc=example,dc=com
Lookup Bind Password: <miniobind password>
User DN Search Base:  cn=users,cn=accounts,dc=example,dc=com
User DN Search Filter: (uid=%s)

click "Save" I get forwarded to a page that shows the configuration with all the same settings, and the banner up top "The server is restarting."

"LDAP Enabled"
Yes
Server Address
freeipa.example.com:389
Lookup Bind Password:
<miniobind password>
etc...

I hit refresh button and then it shows
"LDAP Eanbled"
No

## Expected Behavior

When I click Save I would expect that the settings would be saved and the LDAP authenticaiton would work but something seems to be silently erroring. 

Enable Minio in TrueNAS 
Login 
Configure LDAP per instructions above.

## Your Environment

* Version used (`minio --version`):
 minio version RELEASE.2023-07-21T21-12-44Z (commit-id=d004c4538643eb05ea69b919557d5cf7cc332fae)
Runtime: go1.19.11 linux/amd64
License: GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
Copyright: 2015-2023 MinIO, Inc.
* Server setup and configuration: FreeNAS Scale 22.12.3.3
* Operating System and version (`uname -a`): Linux minio-84488457df-6x2sp 5.15.107+truenas minio/minio#1 SMP Tue Jul 25 00:05:02 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

bexsoft:
fixes #3028 

## What does this do?

- Fixed LDAP Enabled / Disabled calculation
- Removed Password Label from main page to avoid confusion
- Added Re-enter password notice to edit configurations page as this is required for security purposes
- Display LDAP ENV variables in case of configured

## How does it look?

Screenshot 2023-09-13 at 17 31 09
Screenshot 2023-09-13 at 17 26 54
Screenshot 2023-09-13 at 17 21 06
Screenshot 2023-09-13 at 17 18 54
Screenshot 2023-09-13 at 17 17 59
Screenshot 2023-09-13 at 17 17 33
Screenshot 2023-09-13 at 17 17 30
Screenshot 2023-09-13 at 17 17 27

### DIFF
--- a/portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
+++ b/portal-ui/src/screens/Console/IDP/LDAP/IDPLDAPConfigurationDetails.tsx
@@ -18,15 +18,20 @@ import React, { Fragment, useEffect, useState } from "react";
 import {
   Box,
   Button,
+  ConsoleIcon,
   EditIcon,
   FormLayout,
   Grid,
+  HelpBox,
   InputBox,
   Loader,
   PageLayout,
   RefreshIcon,
   Switch,
   Tabs,
+  Tooltip,
+  ValuePair,
+  WarnIcon,
 } from "mds";
 import { api } from "api";
 import { ConfigurationKV } from "api/consoleApi";
@@ -40,7 +45,6 @@ import {
 } from "../../../../systemSlice";
 import { ldapFormFields, ldapHelpBoxContents } from "../utils";
 import ScreenTitle from "../../Common/ScreenTitle/ScreenTitle";
-import LabelValuePair from "../../Common/UsageBarWrapper/LabelValuePair";
 import PageHeaderWrapper from "../../Common/PageHeaderWrapper/PageHeaderWrapper";
 import AddIDPConfigurationHelpBox from "../AddIDPConfigurationHelpbox";
 import LDAPEntitiesQuery from "./LDAPEntitiesQuery";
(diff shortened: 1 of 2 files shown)


cesnietor:
I'd say that we should disable the other buttons since if you click on them before restarting. 
Screenshot 2023-09-14 at 3 13 01 PM
An example of clicking refresh:
Screenshot 2023-09-14 at 3 17 41 PM
Also it's worth noting that if you dismiss the restart button, there's no other way to restart the server other than that message, we might have to allow that functionality (enhancement) somewhere else cause it can be easily dismissed.

cesnietor:
If you click Edit Configuration and then define the password and click Save.
Then you click again on `Edit Configuration`, the password is still there.
Shouldn't we forget it every time you click on Edit? since we do a post when we do Save.?
Screenshot 2023-09-14 at 3 15 18 PM

bexsoft:
> I'd say that we should disable the other buttons since if you click on them before restarting. Screenshot 2023-09-14 at 3 13 01 PM An example of clicking refresh: Screenshot 2023-09-14 at 3 17 41 PM Also it's worth noting that if you dismiss the restart button, there's no other way to restart the server other than that message, we might have to allow that functionality (enhancement) somewhere else cause it can be easily dismissed.



> I'd say that we should disable the other buttons since if you click on them before restarting. Screenshot 2023-09-14 at 3 13 01 PM An example of clicking refresh: Screenshot 2023-09-14 at 3 17 41 PM Also it's worth noting that if you dismiss the restart button, there's no other way to restart the server other than that message, we might have to allow that functionality (enhancement) somewhere else cause it can be easily dismissed.

We would need to discuss this restart banner as it can cause some other UX issues in other parts of the application

bexsoft:
> If you click Edit Configuration and then define the password and click Save. Then you click again on `Edit Configuration`, the password is still there. Shouldn't we forget it every time you click on Edit? since we do a post when we do Save.? Screenshot 2023-09-14 at 3 15 18 PM

Fixed

cesnietor (approved):

reivaj05 (approved):

PR #3051, opened 2023-09-19
fix: handle crash in error handling of access keys page

prakashsvmx:
The current response handler in UI crashes on any error from the back end.

This can be replicated easily by configuring LDAP  and trying to create "Access Key" or simulating an error scenario from backend.

<details>
Error with latest MinIO:

image

Replicate this in dev
Console_SA-Error

Fix:
image

</details>

### DIFF
--- a/portal-ui/src/screens/Console/Account/Account.tsx
+++ b/portal-ui/src/screens/Console/Account/Account.tsx
@@ -47,7 +47,7 @@ import { errorToHandler } from "api/errors";
 import HelpMenu from "../HelpMenu";
 import { ACCOUNT_TABLE_COLUMNS } from "./AccountUtils";
 import { useAppDispatch } from "store";
-import { ApiError, ServiceAccounts } from "api/consoleApi";
+import { ServiceAccounts } from "api/consoleApi";
 import {
   setErrorSnackMessage,
   setHelpName,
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

kaankabalak (approved):

portal-ui/src/screens/Console/Account/Account.tsx
```diff
@@ -105,8 +105,11 @@ const Account = () => {
           setRecords(sortedRows);
         })
         .catch(async (res) => {
```
kaankabalak, line 107: As we have removed the `await` portion, I don't think we would need `async` here

portal-ui/src/screens/Console/Account/AddServiceAccountScreen.tsx
```diff
@@ -85,8 +85,7 @@ const AddServiceAccount = () => {
 
         .catch(async (res) => {
```
kaankabalak, line 86: Similar to my other comment, as we have removed the `await` portion, I don't think we would need `async` here

prakashsvmx: Thank you :+1: . I have updated.

kaankabalak (approved):

harshavardhana:
Some tests are failing @prakashsvmx

portal-ui/src/screens/Console/Account/AddServiceAccountScreen.tsx
```diff
@@ -83,10 +83,9 @@ const AddServiceAccount = () => {
           });
         })
 
-        .catch(async (res) => {
+        .catch((res) => {
           setAddSending(false);
-          const err = (await res.json()) as ApiError;
```
cesnietor, line 88: ahh these casts are only needed when don't send a response type in a success response. 
Thanks for fixing this.

cesnietor (approved):

bexsoft (approved):

bexsoft:
Errors seems not be related to this PR. Need to investigate

Issue #3010, opened 2023-08-22
GetConsoleHTTPClient should not cache clientIP


Timost:
## NOTE
If this case is urgent, please subscribe to [Subnet](https://min.io/pricing) so that our 24/7 support team may help you faster.

## Expected Behavior
Minio console forwards the correct clientIP to the minio-go client

## Current Behavior
Minio console caches the first clientIP it sees and then reuses it along with the client even if the clientIP changed.

## Possible Solution
Cache *only* the client (if possible) and set the `clientIP` every time in [`GetConsoleHTTPClient`](https://github.com/dvaldivia/console/blob/master/restapi/client-admin.go#L532C16-L532C16)

## Steps to Reproduce (for bugs)
See original issue (below)

## Context
This a follow up to an issue/discussion on minio/minio: see https://github.com/minio/minio/discussions/17878#discussioncomment-6792502

## Regression
No

## Your Environment

Version used (minio --version): RELEASE.2023-08-04T17-40-21Z (didn't test with the latest version yet.
Server setup and configuration:
    4 vm
    1 podman pod with 1 minio container and 1 nginx container each
Operating System and version (uname -a): Host runs Ubuntu 22.04 with podman 4.6.0

cesnietor:
issue might have been introduced by https://github.com/minio/console/pull/2864 as mentioned in the other thread, we'll investigate.

Timost:
Hi,

Timost:
Hi,
Would you consider it, if I opened a PR for that ?
Or is it too delicate for an outsider.

cesnietor:
@Timost  we are already working on it.
You're more than welcome to help us testing when we send a PR.
Also is an open source project so anyone can send contributions :D 
Thanks.

Timost:
Great news ! 
I'll be waiting for the PR then !

Thank you ! 🙏

PR #3038, opened 2023-09-07
Fix regression on login path when using redirect url

cesnietor:
Fixes: https://github.com/minio/console/issues/3036

We changed the way of handling the errors in here https://github.com/minio/console/pull/3005 which caused a regression on the login path when using `MINIO_BROWSER_REDIRECT_URL` env variable.

### Test Steps
Please follow same steps as  https://github.com/minio/console/pull/2818
Note: Use correct command as explained in the comments like:
```
make build && CI=true MINIO_BROWSER_REDIRECT_URL=http://localhost:8000/console/subpath/ MINIO_SERVER_URL=http://localhost:9000 ./minio server --console-address ":9090" ~/Data
```

Go to http://localhost:8000/console/subpath/

It should redirect you to `http://localhost:8000/console/subpath/login` before it was redirecting to `http://localhost:8000/login`.

Other test:
- Login and then go to a different pat such as browser or access keys
- Delete the local storage for the page
- refresh the site, it should redirect you again to `http://localhost:8000/console/subpath/login`

PR #3056, opened 2023-09-21
Fix: cache clientIP in GetConsoleHTTPClient

adriangitvitz:
Fixes: #3010 

Changes:
* Remove cache for ClientIP 

This was tested using the provided nginx configuration from the following discussion
 ( https://github.com/minio/minio/discussions/17878#discussioncomment-6792502) 
 
 Local test environment, using LXC ( Ubuntu 22.04 ):
 * Nginx Proxy
 * Minio Server


restapi/client-admin.go
```diff
@@ -524,6 +524,7 @@ func GetConsoleHTTPClient(address string, clientIP string) *http.Client {
 	client, ok := httpClients.m[address]
 	httpClients.Unlock()
 	if ok {
+        client.Transport.(*ConsoleTransport).ClientIP = clientIP
```
harshavardhana, line 527: Why would do this kind of ract updates for in-memory copies, is there a better way to set this which is safer?

harshavardhana (changes requested):

restapi/client-admin.go
harshavardhana: And what happens if it's not ConsoleTransport - how is a crash protected here ?

cesnietor:
Can you please add more details to the description of the PR, including which issue it fixes, changes/context and how to test it. 
Some examples: https://github.com/minio/console/pull/3047,  https://github.com/minio/console/pull/3038.

restapi/client-admin.go
dvaldivia: Since we are using that `httpClients` cache, we need to update the clientIP everytime we pull the cached client, not sure why we introduced the cache in the first place since we need this clientIP be different for each client to properly communicate who is making the request. cc @vadmeste

harshavardhana: Yeah we cannot update ClientIP in this manner since its racy, its a pointer and its shared.

dvaldivia:
I tested the fix, it does addresses the problem, but let's remove the http client cache and make the casting of the transport more defensive

restapi/client-admin.go
```diff
@@ -520,17 +510,14 @@ func GetConsoleHTTPClient(address string, clientIP string) *http.Client {
...
-	client, ok := httpClients.m[address]
-	httpClients.Unlock()
-	if ok {
-		return client
+	client := PrepareConsoleHTTPClient(isLocalIPAddress(address), clientIP)
+
+	if client.Transport != nil {
+		if transport, ok := client.Transport.(*ConsoleTransport); ok {
+			transport.ClientIP = clientIP
+		}
```
harshavardhana, line 518: Remember this transport is a global object and pointer manipulation in this manner is racy.

We have to do it differently by adding this information as part of the request headers. Meaning this information must be sent differently than the way it's implemented.

harshavardhana (changes requested):

restapi/client-admin.go
harshavardhana: 👍🏼

dvaldivia (approved):
Tested. Works as expected.

dvaldivia (approved):
Tested.

bexsoft (approved):
LGTM, Tested along side with @dvaldivia & @adriangitvitz

Issue #3053, opened 2023-09-20
UI user with 'consoleAdmin' policy can't change his own password


nixargh:
**Local** user can change other's passwords but not his own. Well, he can enter a new one and save but no real change happens.
His password can be changed using another user.

## Expected Behavior
User with proper permissions can change his own password using UI button "Change Password".

## Current Behavior
UI shows no errors, save button became not active but password stays unchanged.

## Steps to Reproduce (for bugs)
1. Create user with *consoleAdmin* policy and password.
2. Log in to UI console.
3. Change password.
4. Try to login with a new password.

## Your Environment
* MinIO version used (`minio --version`):
```minio version RELEASE.2023-08-31T15-31-16Z (commit-id=e47e625f73bb1c9672e6c8e4c3b3f9a0e93423a0)
Runtime: go1.19.12 linux/amd64
License: GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
Copyright: 2015-2023 MinIO, Inc.
```
* Server setup and configuration: [this type of setup](https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-multi-node-multi-drive.html)
* Operating System and version (`uname -a`):
```Linux minio.dev 5.15.0-79-generic #86-Ubuntu SMP Mon Jul 10 16:07:21 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux```

prakashsvmx:
I checked the latest version and i am able to change the password.

however the user experience can be improved by providing feedback and forcing the user to re-login after changing the password.

![image](https://github.com/minio/console/assets/23444145/0365f6d7-2923-49bd-8973-cc00d2baff63)

nixargh:
> I checked the latest version and i am able to change the password.
> 
> Only MINIO_ROOT_USER login cannot change the password. all other users can change the password.
> 
> however the user experience can be improved by providing feedback when password change is success and forcing the user to re-login after changing the password.
> 
> ![image](https://user-images.githubusercontent.com/23444145/269320411-0365f6d7-2923-49bd-8973-cc00d2baff63.png)

@prakashsvmx Have you checked you can log in with the new password then? In my case I can login only with the old one.

nixargh:
OK, issue solved :) I was pushing another button and it doesn't work that way, so maybe you should hide it for user matching the logged one?
[...]

PR #3058, opened 2023-09-21
Improve change password  user experience

prakashsvmx:
improve change password  user experience and add doc reference to rotating credentials for `MINIO_ROOT_USER`.

Fixes https://github.com/minio/console/issues/3053

image

### DIFF
--- a/portal-ui/src/screens/Console/Account/ChangePasswordModal.tsx
+++ b/portal-ui/src/screens/Console/Account/ChangePasswordModal.tsx
@@ -26,11 +26,16 @@ import {
 import ModalWrapper from "../Common/ModalWrapper/ModalWrapper";
 
 import { modalStyleUtils } from "../Common/FormComponents/common/styleLibrary";
-import { setModalErrorSnackMessage } from "../../../systemSlice";
+import {
+  setModalErrorSnackMessage,
+  setSnackBarMessage,
+} from "../../../systemSlice";
 import { useAppDispatch } from "../../../store";
 import { api } from "api";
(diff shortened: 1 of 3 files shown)


bexsoft (approved):
LGTM

cesnietor:
When current password is wrong we show 403 Forbidden, can we show the message returned instead?
`"error please check your current password"`, to give the user more details.

cesnietor:
When console is using same credentials as root password, it doesn't allow you to update them.
Screenshot 2023-09-21 at 1 18 05 PM
Screenshot 2023-09-21 at 1 15 03 PM

If we know that in advance, then we shouldn't disable the fields or not show them.

cesnietor:
If I have a new user with no policies assigned to them, and try to update their password, the logs say:
```
ErrorWithContext:error please check your current password
%!(EXTRA <nil>, madmin.ErrorResponse=Access Denied.)
```
And the UI show the 403 forbidden. If it's not allowed shouldn't we disable the feature for them? So that they don't even try to do it.

prakashsvmx:
@cesnietor 
I have updated to show the correct error message.

for other items:
- In console i could not find a way to know if the logged in user is with root credentials.
- The permissions for change password and create access keys are provided with `conole-ui` by default. so we will not be able to disable it. ( even when no policies assigned)
-

cesnietor (approved):

cesnietor:
Thanks @prakashsvmx  lgtm

bexsoft (approved):

PR #3066, opened 2023-09-30
Object Browser Refactor

### LINKED ISSUE #3001: Bucket shows empty when a new object is uploaded
If an object is selected and a new object gets uploaded, the UI now lists no items instead of reloading and listing the new item.

Note:
Same happens when multiple objects get selected and deleted.

Before uploading object:
Screenshot 2023-08-16 at 10 34 08 AM

After uploading new object:
Screenshot 2023-08-16 at 10 34 24 AM

## Expected Behavior
It should list the new item including existing items

## Current Behavior
It shows an empty list

## Possible Solution

Path stays on the previous object selected which might be causing some issues.

- On a bucket with objects, select an object
- The menu on the right should show up (with actions Download, preview, etc.)
- Upload a new object
- [Bug] Here the list shows empty and still pointing to the previous object.

## Your Environment

* MinIO version used (`minio --version`): RELEASE.2023-08-09T23-30-22Z

bexsoft:
fixes #3001 

## What does this do?

Simplified the Object Browser code by changing the location of Websockets to be a middleware as well as simplifying the logic to handle the object browser paths

- Refactored navigation to be handled only with URL
- Refactored & simplified websocket
- Updated components to use mds
- Fixed an issue with Anonymous access and file selection
- Fixed an issue with anonymous access and download selection click
- Fixed an issue with object details selection on root path from a bucket
- Simplified reducer

### DIFF
--- a/portal-ui/src/websockets/objectBrowserWSMiddleware.ts
+++ b/portal-ui/src/websockets/objectBrowserWSMiddleware.ts
@@ -1,0 +1,231 @@
+// This file is part of MinIO Console Server
+// Copyright (c) 2023 MinIO, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import get from "lodash/get";
+import { Middleware } from "@reduxjs/toolkit";
+import { AppState } from "../store";
+import { wsProtocol } from "../utils/wsUtils";
+import {
+  errorInConnection,
+  newMessage,
+  resetMessages,
+  setRecords,
+  setReloadObjectsList,
+  setRequestInProgress,
+  setSearchObjects,
+  setSelectedBucket,
+  setSelectedObjects,
+  setSimplePathHandler,
+} from "../screens/Console/ObjectBrowser/objectBrowserSlice";
+import {
+  WebsocketRequest,
+  WebsocketResponse,
+} from "../screens/Console/Buckets/ListBuckets/Objects/ListObjects/types";
+import { decodeURLString, encodeURLString } from "../common/utils";
+import { permissionItems } from "../screens/Console/Buckets/ListBuckets/Objects/utils";
+import { setErrorSnackMessage } from "../systemSlice";
+
+let wsInFlight: boolean = false;
+let currentRequestID: number = 0;
+
+export const objectBrowserWSMiddleware = (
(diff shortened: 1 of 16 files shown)


jinapurapu:
After creating a path, "Loading" remains, drag and drop upload doesn't work. 



https://github.com/minio/console/assets/65002498/9de26a59-a502-4424-ae69-93de7663c089

bexsoft:
> After creating a path, "Loading" remains, drag and drop upload doesn't work.
> 
>  Screen.Recording.2023-10-02.at.11.09.53.AM.mov

Fixed: 
Screen Recording 2023-10-02 at 1 50 23 p m 2023-10-02 13_51_28

jinapurapu:
Clicking arrow icon to invert object list order causes "Loading" to hang:

https://github.com/minio/console/assets/65002498/c3700d31-2f33-48e2-ba30-9b788f2e39b3

portal-ui/src/websockets/objectBrowserWSMiddleware.ts
```diff
@@ -0,0 +1,231 @@
...
+              return;
+            }
+
+            if (
+              response.error ===
+              "The Access Key Id you provided does not exist in our records."
+            ) {
+              // Session expired.
+              window.location.reload();
+            } else if (response.error === "Access Denied.") {
```
cesnietor, line 103: could we use codes, instead of strings for these type of validations?
Cause the thext response could change any time and we wouldn't know that it could break this.

portal-ui/src/websockets/objectBrowserWSMiddleware.ts
```diff
@@ -0,0 +1,231 @@
...
+  WebsocketResponse,
+} from "../screens/Console/Buckets/ListBuckets/Objects/ListObjects/types";
+import { decodeURLString, encodeURLString } from "../common/utils";
+import { permissionItems } from "../screens/Console/Buckets/ListBuckets/Objects/utils";
+import { setErrorSnackMessage } from "../systemSlice";
+
+let wsInFlight: boolean = false;
+let currentRequestID: number = 0;
+
+export const objectBrowserWSMiddleware = (
```
cesnietor, line 44: do you mind adding more context on what this refactor is doing? vs the previous one? just fo have it documented in the PR.

cesnietor (commented):
Tested different scenarios with paths, tracing, anonymous buckets, deleting and adding multiple objects. All seems great.! Just left few comments.

portal-ui/src/websockets/objectBrowserWSMiddleware.ts
bexsoft: Let me check as we are not returning response codes in the web socket. We do have a test to validate this behavior

bexsoft:
> Clicking arrow icon to invert object list order causes "Loading" to hang:
> 
>  Screen.Recording.2023-10-02.at.1.40.10.PM.mov

Fixed

portal-ui/src/websockets/objectBrowserWSMiddleware.ts
bexsoft: Will fix this in an upcoming PR as some extra changes need to be done

cesnietor: sounds good.

cesnietor (approved):

jinapurapu (approved):

PR #3084, opened 2023-10-10
mui Grid component replacement

bexsoft:
## What does this do?

Replaced Grid component from mui with mds equivalent, also migrated PreviewFileContent component to mds

## How does it look?

Screenshot 2023-10-10 at 2 21 40 p m


prakashsvmx (approved):

bexsoft:
> CSV files are downloading on preview
> 
>  Screen.Recording.2023-10-11.at.12.45.46.PM.mov

Hi @jinapurapu!

This issue should not be related to this change as only styles are affected by this. will review accordingly in another PR

dvaldivia (approved):

PR #3088, opened 2023-10-12
Add UI to select number of non-current versions subject to ILM expiry rule

jinapurapu:
Adds unit selector to Add Lifecycle rule modal to enable selection to expire non-current versions of objects after a number of non-current versions OR a number of days, and UI changes to display non-current version count based rules.

Edit ILM Rule for version count based rules in a separate PR

https://github.com/minio/console/assets/65002498/b949e161-c712-4b24-9032-dadb44a0f089


bexsoft (approved):
LGTM

cesnietor:
@jinapurapu  please remove the console log:
```set to valid:  true  lifecycledays:  231906034276324167843127894132794372979234174138`

**Comment by cesnietor** [comment]
if the number is too long:
`231906034276324167843127894132794372979234174138207324072341`
the request fails like:
```
"parsing body body from \"\" failed, because json: cannot unmarshal number 2.3190603427632416e+59 into Go struct field AddBucketLifecycle.noncurrentversion_expiration_days of type int32"
```
And there is no ui indication that something went wrong.
Could we add a max allowed field in the input.
And also show the error if it occurs?

**Comment by cesnietor** [comment]
As UX this text doesn't have a way to let the user know that it's clickable:
Screenshot 2023-10-31 at 1 29 51 PM
So users may never found out about it.
We can fix this by simply adding something like an icon e.g. `^` next to the text?

**Comment by cesnietor** [comment]
0 value should not be allowed:
```

{detailedMessage: "expiration days cannot be set to zero",…}
detailedMessage
: 
"expiration days cannot be set to zero"
message
: 
"an error occurred, please try again"
```

**Comment by bexsoft** [comment]
> Since 0 days is not a valid value. We correctly disable the button but we don't say what the valid values are, or that `0` is not valid. not sure if this can be done via a simple helper text or something else @bexsoft . Screenshot 2023-10-31 at 1 39 40 PM

Yes, it can be done by setting the `error` prop for this specific field

File: portal-ui/src/screens/Console/Buckets/BucketDetails/AddLifecycleModal.tsx [hunk 1]
```diff
@@ -116,7 +117,11 @@ const AddLifecycleModal = ({
         valid = false;
       }
     }
+    if (!lifecycleDays || parseInt(lifecycleDays) === 0) {
+      valid = false;
+    }
     setIsFormValid(valid);
+    console.log("set to valid: ", valid, " lifecycledays: ", lifecycleDays);
```
bexsoft, line 124: Please remove this line

bexsoft (changes requested):

cesnietor:
Can we change the validation error to be something like: `value should be between 1-<the actual max>`. We can save them to google the value by actually telling them the max.

cesnietor:
Tested, it now shows the validation warnings as well as some docs on the days/versions.
Screenshot 2023-11-09 at 4 01 43 PM
Screenshot 2023-11-09 at 4 01 39 PM
Screenshot 2023-11-09 at 4 01 32 PM

cesnietor (approved):

cesnietor (changes requested):
Ah @jinapurapu the message should be either `less than or equal to 2147483647` or `less than 2147483648` cause it allows `2147483647`.
Screenshot 2023-11-09 at 4 06 41 PM

cesnietor (approved):

bexsoft (approved):

PR #3100, opened 2023-10-20
Support resolving of listen addrs to multiple IPv4/IPv6

vadmeste:
If you specify --console-address "myaddress:9001" while myaddress 
can be resolved to one IPv4 address and another IPv6 address, 
Console ignores the IPv6 address. This commit fixes that.

This PR is just copying restapi/server.go restapi/custom-server.go and modifying the latter
to support listening to multiple interfaces, IP4 and IPv6


cesnietor:
If swagger gets updated with changes for that file, we will be no longer getting them cause we are excluding that file. 
Wouldn't it make more sense to to contribute to https://github.com/go-swagger/go-swagger maybe others can also benefit from it, or does it only make sense for MinIO?
Also, which lines contain the changes? is hard to see what is cp and what was added.

vadmeste:
> If swagger gets updated with changes for that file, we will be no longer getting them cause we are excluding that file.
Wouldn't it make more sense to to contribute to https://github.com/go-swagger/go-swagger maybe others can also benefit from it, or does it only make sense for MinIO?

This may take time; also go-swagger needs to add a flag to ignore generating server.go and gives us the control. For example, some wants prefer FIPS ciphers for TLS communication and some won't like that. MinIO has a flag for that.

yes, I did not find another solution to make reviewing this easy; for now you can do vimdiff restapi/server.go restapi/custom-server.go or equivalent;

harshavardhana:
We have a real customer requirement here @cesnietor - go-swagger is not designed yet to provide more control to developers so we may have to take this route for now.

But I think we need to find a solution here because go-swagger will be slow to accept changes that we need.

dvaldivia (approved):
LGTM

harshavardhana (approved):

PR #3117, opened 2023-10-31
Fix days selector to ignore daylight savings

### LINKED ISSUE #3106: "Expires cannot be greater than 7 days" when sharing even when under 7 days
## Expected Behavior

Creating a share link for a duration less than or equal to 7 days should generate a link and not not raise a warning.

## Current Behavior

Selecting a duration larger than 6d23h already triggers the warning and fails to create links.

This is most likely due to not handling DST correctly (which changes this weekend). If the max limit of 7 days is not from the console but the MinIO backend itself, I wonder how AWS S3 handles this and whether MinIO has the same behaviour here.

## Possible Solution

Create a link for a shorter duration.

1. Open any bucket with a file inside
2. Select the "Share" action from the sidebar on the right
3. As the default duration is 7 days, there is instantly a warning and no link is generated

## Context

n/a

## Regression

Not sure, though likely not.

## Your Environment

* MinIO version used (`minio --version`): `minio version RELEASE.2023-10-25T06-33-25Z (commit-id=c60f54e5be7302d82d0d8fc404c056fea4e2bf4e)`
* Server setup and configuration: Single-Node, Docker
* Operating System and version (`uname -a`): `Linux <hostname> 6.1.0-11-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-4 (2023-08-08) x86_64 GNU/Linux`

nreisingercres:
Fixes #3106 by
1. Ignoring daylight savings
2. Displaying the timezone to avoid confusion with the date around daylight savings


bexsoft:
Please run `make prettify ` under `portal-ui` to remove the workflow error. Thank you in advance

bexsoft (approved):

cesnietor (approved):

PR #3128, opened 2023-11-14
[WIP] Enabled Dark Mode in Console

### LINKED ISSUE #2429: Dark mode for console
A dark mode / theme for the console UI would be nice.
The current UI uses a white background which is very disturbing, especially in dark environments.

bexsoft:
fixes #2429 

## What does this do?

- Dark mode for login page will be tied to system settings when no configuration is set
- Dark mode will be stored in Application storage once set
- If Set system will automatically set the Dark Mode configuration

## How does it look?

Screenshot 2023-11-14 at 1 51 44 p m
Screenshot 2023-11-14 at 1 50 02 p m
Screenshot 2023-11-14 at 1 49 48 p m
Screenshot 2023-11-14 at 1 49 35 p m
Screenshot 2023-11-14 at 1 49 27 p m
Screenshot 2023-11-14 at 1 49 19 p m

### DIFF
--- a/portal-ui/package.json
+++ b/portal-ui/package.json
@@ -10,7 +10,7 @@
     "local-storage-fallback": "^4.1.1",
     "lodash": "^4.17.21",
     "luxon": "^3.4.3",
-    "mds": "https://github.com/minio/mds.git#v0.12.1",
+    "mds": "https://github.com/minio/mds.git#v0.12.2",
     "react": "^18.1.0",
     "react-component-export-image": "^1.0.6",
     "react-copy-to-clipboard": "^5.0.2",
--- a/portal-ui/src/StyleHandler.tsx
+++ b/portal-ui/src/StyleHandler.tsx
@@ -42,7 +42,7 @@ const StyleHandler = ({ children }: IStyleHandler) => {
   return (
     <Fragment>
       <GlobalStyles />
-      <ThemeHandler darkMode={!!darkMode} customTheme={thm}>
+      <ThemeHandler darkMode={darkMode} customTheme={thm}>
         {children}
       </ThemeHandler>
     </Fragment>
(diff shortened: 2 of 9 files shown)


cesnietor:
Why only the login page uses the system settings ?
The behavior seems odd when you login in a black page and then get redirected to a white screen. This is in the very first time I open it.

cesnietor:
For the License page, should we also have a dark mode for the selected plan?
Screenshot 2023-11-14 at 2 03 22 PM

dvaldivia (dismissed):
🎉🎉🎉🎉

bexsoft:
> Why only the login page uses the system settings ? The behavior seems odd when you login in a black page and then get redirected to a white screen. This is in the very first time I open it.

This was an original requirement when we designed it

cesnietor:
On other note, I find the dark mode to be a bit tough on my eyes. There's not enough contrast. on the text with the dark background. Would it make sense to make the text brighter? probably just white?  or the background more dark? e.g. Github:
Screenshot 2023-11-14 at 2 07 14 PM

cesnietor:
> > Why only the login page uses the system settings ? The behavior seems odd when you login in a black page and then get redirected to a white screen. This is in the very first time I open it.
> 
> This was an original requirement when we designed it

@bexsoft  yes but we don't respect it once you login. it should remain in the systems preference unless you modify it?

bexsoft:
> > > Why only the login page uses the system settings ? The behavior seems odd when you login in a black page and then get redirected to a white screen. This is in the very first time I open it.
> > 
> > 
> > This was an original requirement when we designed it
> 
> @bexsoft yes but we don't respect it once you login. it should remain in the systems preference unless you modify it?



> > > Why only the login page uses the system settings ? The behavior seems odd when you login in a black page and then get redirected to a white screen. This is in the very first time I open it.
> > 
> > 
> > This was an original requirement when we designed it
> 
> @bexsoft yes but we don't respect it once you login. it should remain in the systems preference unless you modify it.
Let me review as this was the original requirement (Just the Login page should change to system settings configuration and then go back to original color)

cesnietor:
> (Just the Login page should change to system settings configuration and then go back to original color)
why is that? is a weird experience.

PR #3132, opened 2023-11-20
FIX: Use STS env variable to increase the IDP token expiration

adriangitvitz:
Share link duration is based on the token expiration, this increases the IDP token expiration so the share link is able to last longer, by using an env variable called MINIO_STS_DURATION


pkg/auth/idp/oauth2/provider.go
```diff
@@ -332,13 +332,12 @@ func (client *Provider) VerifyIdentity(ctx context.Context, code, state, roleARN
 		client.RefreshToken = oauth2Token.RefreshToken
 
 		expiration := token.GetConsoleSTSDuration()
-		if exp := getIDPTokenExpiration(); exp > 0 {
```
cesnietor, line 335: can we change the order and the `GetConsoleSTSDuration ` so that we only read the env variable once?
Here we are reading it twice.
We can do:
- read, use value if defined.
- if empty check for the `oauth2Token.Expiry.Sub(time.Now().UTC())`
- if nothing done use the default 12 h
- then check for the 15 min  minimum.

cesnietor: this would make it easier to understand from top to bottom.

cesnietor (approved):

dvaldivia (approved):

PR #3141, opened 2023-12-01
add encoded filename as part of delete api url

prakashsvmx:
@cesnietor , @poornas 

Console has different API (endpoint) urls for different scenarios to handle deletes. 

the below url is modified to pass filename in case of single object delete, in url params

http://localhost:5005/api/v1/buckets/test-v-bucket/objects?prefix=MS50eHQ%3D&non_current_versions=true&bypass=false

Context, this is done so that we can do special operations with apis that have the prefix defined. single objects where being sent in the array so a single delete object api was needed.

### DIFF
--- a/restapi/user_objects.go
+++ b/restapi/user_objects.go
@@ -64,6 +64,7 @@ func registerObjectsHandlers(api *operations.ConsoleAPI) {
 	})
 	// delete object
 	api.ObjectDeleteObjectHandler = objectApi.DeleteObjectHandlerFunc(func(params objectApi.DeleteObjectParams, session *models.Principal) middleware.Responder {
+		fmt.Println("ObjectDeleteObjectHandler", params.Prefix)
 		if err := getDeleteObjectResponse(session, params); err != nil {
 			return objectApi.NewDeleteObjectDefault(err.Code).WithPayload(err.APIError)
 		}
@@ -71,6 +72,8 @@ func registerObjectsHandlers(api *operations.ConsoleAPI) {
 	})
 	// delete multiple objects
 	api.ObjectDeleteMultipleObjectsHandler = objectApi.DeleteMultipleObjectsHandlerFunc(func(params objectApi.DeleteMultipleObjectsParams, session *models.Principal) middleware.Responder {
+		fmt.Println("ObjectDeleteMultipleObjectsHandler", params)
+
 		if err := getDeleteMultiplePathsResponse(session, params); err != nil {
 			return objectApi.NewDeleteMultipleObjectsDefault(err.Code).WithPayload(err.APIError)
 		}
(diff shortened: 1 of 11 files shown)


cesnietor:
I see what you are saying @prakashsvmx .
Humm here let's consider a multiple object call only those that delete different  object names, i.e. if it's a single file with multiple versions to be deleted (due to a versioned bucket), that for us is still a single delete call, with the `all-versions` enabled.

so here we don't need to do any changes to the `/delete-objects` call but rather have a separate one for single-object delete. so that we can plug the name and the path in the prefix param.

prakashsvmx:
@cesnietor 
the single delete is already implemented as part of `http://localhost:5005/api/v1/buckets/test-v-bucket/objects` API endpoint with `DELETE` method.

so i have renamed `path` to `prefix` to be consistent everywhere. and updated all single deletes to use this endpoint.

restapi/user_objects.go
```diff
@@ -64,13 +64,16 @@ func registerObjectsHandlers(api *operations.ConsoleAPI) {
...
+		fmt.Println("ObjectDeleteObjectHandler", params.Prefix)
 		if err := getDeleteObjectResponse(session, params); err != nil {
 			return objectApi.NewDeleteObjectDefault(err.Code).WithPayload(err.APIError)
 		}
 		return objectApi.NewDeleteObjectOK()
 	})
 	// delete multiple objects
 	api.ObjectDeleteMultipleObjectsHandler = objectApi.DeleteMultipleObjectsHandlerFunc(func(params objectApi.DeleteMultipleObjectsParams, session *models.Principal) middleware.Responder {
+		fmt.Println("ObjectDeleteMultipleObjectsHandler", params)
+
```
cesnietor, line 76: please remove

cesnietor:
@allanrogerr  vulns are unrelated to this change. They've been there for a while. We'll have to address them separately.

allanrogerr:
> @allanrogerr vulns are unrelated to this change. They've been there for a while. We'll have to address them separately.

Thats unfortunate. Vulnerabilities should take some priority IMO.

allanrogerr:
@prakashsvmx Will encoding the filename of multiple objects be addressed in a followup PR? That is encoding the files name of the payload of `api/v1/buckets/<bucket-name>/delete-objects` ?

allanrogerr (approved):
Tested. Please consider fixing vulnerabilities.

bexsoft:
> Tested. Please consider fixing vulnerabilities.

Fixed in https://github.com/minio/console/pull/3146

cesnietor:
>@prakashsvmx Will encoding the filename of multiple objects be addressed in a followup PR? That is encoding the files name of the payload of api/v1/buckets/<bucket-name>/delete-objects ?
for multiple files we send them in the payload, so no need to encode them.

Issue #3018, opened 2023-08-28
web console:local logins unreachable if SSO is down


bpedersen2:
If  the configured SSO (openid) provider is not reachable (due to expired certificate), then the frontend does not offer the alternate login with local credentials, instead only an error is displayed: 

```
An error has occurred
The backend cannot be reached.
```

I would expect the switch to login with a local password to be available in this case. 
minio Version: RELEASE.2023-08-16T20-17-30Z (go1.19.12 linux/amd64)

bexsoft:
Hi @bpedersen2! How are you doing?

In the Login page we offer the ability to login using embedded IDP (Local Password) in case of necessary.

To access this option, you have to select the `Other Authentication Methods` dropdown and then Select the `Use Credentials` option:

<img width="1498" alt="Screenshot 2023-09-25 at 12 10 29" src="https://github.com/minio/console/assets/33497058/66acd9d4-507e-45fb-acc3-c0d9737bc43d">

This will display the Login fields to use admin or embedded credentials:

<img width="1498" alt="Screenshot 2023-09-25 at 12 10 36" src="https://github.com/minio/console/assets/33497058/fea6f227-468b-4fe2-a6a9-6663dbcdbbf4">

Please let us know if you have further questions.

Regards.

bpedersen2:
The problem is (only a single oidc login configured)  that the selector is not shown, instead there is only an error message:

![screenshot-minio storage frm2 tum de_9001-2023 09 26-08_28_24](https://github.com/minio/console/assets/5336494/52a562de-900a-4099-bbfb-8c6602a25180)

bpedersen2:
To test, you should use e.g. a keycloak instance and shut it down. Testing with github/google/etc. will be tricky as you need to make them unreachable.

bpedersen2:
I suspect https://github.com/minio/console/pull/2258/files#r1336700937 could be the culprit here.

cesnietor:
thanks @bpedersen2 we'll reopen and investigate. Also feel free to send a PR and we can test it. It won't be prioritized right now.

ribeaud:
This happens during _local_ testing (http://localhost:9001/) as well. My **OpenID** provider is up and running.

@bpedersen2: by certificate expiration, do you mean HTTPS certificate? So, does it mean, I would not get this error if I am using HTTPS and the certificate is OK?

bpedersen2:
[...]

PR #3168, opened 2023-12-23
simplify the provider config init, loading and allow reachable IDPs

harshavardhana:
fixes https://github.com/minio/console/issues/3018

### DIFF
--- a/restapi/user_login.go
+++ b/restapi/user_login.go
@@ -178,8 +178,6 @@ func isKubernetes() bool {
 
 // getLoginDetailsResponse returns information regarding the Console authentication mechanism.
 func getLoginDetailsResponse(params authApi.LoginDetailParams, openIDProviders oauth2.OpenIDPCfg) (ld *models.LoginDetails, apiErr *CodedAPIError) {
-	ctx, cancel := context.WithCancel(params.HTTPRequest.Context())
-	defer cancel()
 	loginStrategy := models.LoginDetailsLoginStrategyForm
 	var redirectRules []*models.RedirectRule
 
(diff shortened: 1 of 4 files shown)


restapi/user_login.go
```diff
@@ -177,58 +177,69 @@ func isKubernetes() bool {
...
+			DisplayName: displayName,
+			ServiceType: serviceType,
 		}
+
+		redirectRules = append(redirectRules, &redirectRule)
 	}
+
+	if len(openIDProviders) > 0 && len(redirectRules) == 0 {
+		return nil, ErrorWithContext(ctx, errors.New(403, "no reachable or configured IDP found, login not allowed"), ErrOauth2Provider)
+	}
```
tsuzu, line 234: I'm sorry if my understanding is incorrect.
I think we should be able to sign in with credentials(ID/Password) even if all `openIDProviders` are unavailable.
but with this if statement, the error page will be shown when all `openIDProviders` are unavailable.
so it may be better that `getLoginDetailsResponse` never returns errors and instead returns `LoginDetails` with `LoginDetailsLoginStrategyForm`

harshavardhana: Oh yes that is fine. I kept it the way it is today however the idea was that we do not fail if one of the IDP is unreachable.

tsuzu: I get it. Thanks. However, the objective of #3018 is to ensure successful sign-in even when the sole oidc provider is unavailable. Therefore, I believe the issue shouldn't be marked as closed until this scenario is addressed.

.github/workflows/jobs.yaml
```diff
@@ -1169,7 +1169,7 @@ jobs:
           go tool cover -func=all.out | grep total > tmp2
           result=`cat tmp2 | awk 'END {print $3}'`
           result=${result%\%}
-          threshold=71.3
+          threshold=50
```
dvaldivia, line 1172: not needed anymore :P

dvaldivia (changes requested):

dvaldivia (approved):

PR #3178, opened 2023-12-29
Rename `portal-ui` to `web-app`

dvaldivia:
To keep the naming of the web app across projects consistent we are renaming `portal-ui` to `web-app`


bexsoft (approved):
LGTM

dilverse (approved):

harshavardhana (commented):
why can't we just call it `ui/` ?

dvaldivia:
@harshavardhana we could but we call it `web-app` everywhere else xD

PR #3196, opened 2024-01-17
Use swagger api for delete group

cesnietor:
It also catches correctly the error and shows it as snackbar.
Screenshot 2024-01-17 at 11 01 29 AM

### Test Steps:
1. Go to Administrator/Identity/Groups
2. Add a group with users
3. Delete the group (error should show as snack bar)

2. Add a group without users
3. Delete group (group should be deleted and list of groups regreshed.)


bexsoft (dismissed):
LGTM

bexsoft (approved):

dvaldivia (approved):

PR #3199, opened 2024-01-17
Use swagger api for delete single and multiple Service Accounts

cesnietor:
It also catches correctly the error and shows it as snackbar.
It updates swagger yaml to be able to handle array of strings as input.

### Test Steps
1. Go to User/Access Keys
2. Create a Key
3. Delete a single Key using the `garbage` icon
4. Key should be deleted and keys reloaded.

/// multiple Keys
1. Go to User/Access Keys
2. Create multiple Keys
3. Select multiple Keys
4. Click `Delete Selected`
5. Keys should be deleted and keys reloaded.


prakashsvmx (approved):
Verified. LGTM :+1:

cniackz (approved):
LGTM, thank you @cesnietor plus all tests are passing!.

Issue #3006, opened 2023-08-21
Downloads fail without error when a node in the cluster goes down


hashworks:
When one uses a load balancer with health checks to access a Minio cluster, downtimes of single nodes should not affect users too much or at the very least provide error reports to them. Currently the web client / console does not do that: Failed downloads are marked as a successful download if nodes fail during the transfer.

## Expected Behavior

Minimum: The download fails and reports an error message.
Best: The web client retries the failed parts for a documented amount of time and fails otherwise.

## Current Behavior

The web client marks the download as successful and stores the smaller, unfinished file without an error. Users are not informed about the file corruption.

I downloaded the same file two times:
![img-2023-08-21-221044](https://github.com/minio/console/assets/2362091/6749f8ba-eedd-4b4c-9b4d-315c7aae0abd)

The first download succeded, but during the second download I stopped a cluster node that was providing parts, which did not result in an error and provided a cut-off / corrupted file:
```
 % lsd -l ~/Downloads/output_file*
.rw-rw-r-- hashworks hashworks  10 GB Mon Aug 21 22:00:51 2023  /home/hashworks/Downloads/output_file
.rw-rw-r-- hashworks hashworks 1.3 GB Mon Aug 21 22:10:42 2023  /home/hashworks/Downloads/output_file(1)
```

## Steps to Reproduce

1. Set up a small Minio cluster, f.e. three nodes, access it using a load balancer with health checks
2. Upload a larger file
3. Try to download the file using the web client / console
4. Verify that a specific node is used to provide parts of the file (f.e. using nethogs)
5. Kill minio on said node
6. Note the successful download, but the insufficient file size

## Context

Other tools handle this better:
* `mc` reports an error and fails the download (`EOF`)
* `rclone` retries the failed parts and finishes the download using the degraded cluster

## Your Environment
* MinIO version used (`minio --version`): `RELEASE.2023-08-09T23-30-22Z (commit-id=eb55034dfe5ef82449796c83e3126b245c5aee05)`
* Server setup and configuration: 3x(Minio + nginx reverse proxy) + Hetzner Load Balancer
* Operating System and version (`uname -a`): `Linux test1.example.net 6.1.0-11-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-4 (2023-08-08) x86_64 GNU/Linux`
* Web browser: `Firefox 116.0.3`

cesnietor:
We'll try to handle the error for large files when we are not using direct download.

[...]

PR #3214, opened 2024-01-24
Fail request when error occurs during download

cesnietor:
fixes: https://github.com/minio/console/issues/3006

Checks whether the download was complete in the frontend.
In the backend, setting the headers doesn't work if it fails during io.Copy cause it already starterd writing to the body. So we just log it and handle it in the client layer.

On error:
After:
Screenshot 2024-01-24 at 5 06 38 PM

### Test Steps:
- Deploy a distributed minio e.g. using kind 4 nodes
- expose the service 
- Upload a big object e.g. 100Mb
- Go to the Browser UI
- Cordon one of the nodes
- Download the object (enabling throttling in your browser to emulate slow speed)
- Delete one of the minio pods (it won't start since we cordoned a node)
- In the UI the object download should eventually fail (before it was downloading an inclomplete object)

/// Other tests
- Test downloading multiple files (selecting multiple)
- Test downloading multiple files (selecting one, then download, then only other one and click download) so that the download box appears.

### DIFF
--- a/api/user_objects.go
+++ b/api/user_objects.go
@@ -448,8 +448,9 @@ func getDownloadObjectResponse(session *models.Principal, params objectApi.Downl
 		// override filename is set
 		decodeOverride, err := base64.StdEncoding.DecodeString(*params.OverrideFileName)
 		if err != nil {
-			ErrorWithContext(ctx, fmt.Errorf("unable to decode OverrideFileName: %v", err))
-			panic(err)
+			fmtError := ErrorWithContext(ctx, fmt.Errorf("unable to decode OverrideFileName: %v", err))
+			http.Error(rw, fmtError.APIError.DetailedMessage, http.StatusBadRequest)
+			return
 		}
 
 		overrideName := string(decodeOverride)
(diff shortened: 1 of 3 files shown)


api/user_objects.go
```diff
@@ -448,7 +448,8 @@ func getDownloadObjectResponse(session *models.Principal, params objectApi.Downl
 		// override filename is set
 		decodeOverride, err := base64.StdEncoding.DecodeString(*params.OverrideFileName)
 		if err != nil {
-			return
+			ErrorWithContext(ctx, fmt.Errorf("unable to decode OverrideFileName: %v", err))
+			panic(err)
```
prakashsvmx, line 452: could we just log error instead of  `panic` ?

cesnietor: panic is needed to cause a fail response @prakashsvmx , logging and just returning is what was causing the issue.

cesnietor: After longer investigation I found that Copy already starts writing to the body so we can't send anything to the status or headers anymore. So we need to handle it at the client level.
But it an error happens before that, we can set the proper failure headers and statusCode.

prakashsvmx (approved):

bexsoft (approved):
Tested, LGTM
Screenshot 2024-01-26 at 3 05 44 p m

web-app/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
```diff
@@ -143,7 +146,7 @@ export const download = (
             return;
           }
         }
-        errorCallback(`Unexpected response status code (${req.status}).`);
+        errorCallback(`Unexpected response, download incomplete.`);
```
vadmeste, line 149: Removing the status code does not help debugging, can we try to add more information here ?

cesnietor: status by itself was also not very helpful. You'll need to see the logs for now. We were not exposing the errors before either.
Exposing detailed errors is in our backlog.

Issue #3210, opened 2024-01-22
User's with IAM policies receive message "Access Denied" when trying to access a specific folder in bucket, and no one else's. Temporarily mitigated after refresh.


AuxiliumReleases:
## Expected Behavior

We have a bucket labelled "Clients", that further has multiple folders assigned to users that need to upload contents there.
I.e., if we have a folder named "greentreefrog", that implies that the user "greentreefrog" will be able to read and write in that folder while not seeing any other folders within that bucket.

Each user has their own password and username for their credentials, and they are logging in directly through the web interface.

## Current Behavior

Currently, if the user stated above logs in, and attempts to view the "Clients" bucket through the object browser, they are greeted with the notification "Access Denied", until they refresh the page which allows them in. This has only happened after we updated to the latest version of minio.

Below this line is the policy we apply to users that have folders relevant to them within the client bucket.
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "ListOnlyOwnFolder",
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": [
                "arn:aws:s3:::clients"
            ],
            "Condition": {
                "StringEquals": {
                    "s3:delimiter": [
                        "/"
                    ],
                    "s3:prefix": [
                        "${aws:username}/"
                    ]
                }
            }
        },
        {
            "Sid": "FullAccessToOwnFolder",
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::clients/${aws:username}",
                "arn:aws:s3:::clients/${aws:username}/*"
            ]
        },
        {
            "Sid": "AllowGetBucketLocation",
            "Effect": "Allow",
            "Action": [
                "s3:GetBucketLocation",
                "s3:GetBucketPolicy"
            ],
            "Resource": [
                "arn:aws:s3:::clients"
            ]
        }
    ]
}
```

## Steps to Reproduce (for bugs)

1. Create a bucket with the name "clients"
3. Add multiple folders inside the bucket, with one corresponding to a test user.
4. Make a test user with their username matching one of the folders within the bucket.
5. Assign the policy listed above to that user.
6. Log in as that user, you should be (intermittently) greeted with "Access Denied" which is mitigated on refresh.

## Context

Our clients are unable to upload important documents that are needed for nightly updates relating to software they use.

## Your Environment

* Version used (`minio --version`):
```
[...]

PR #3219, opened 2024-01-30
logout state clean up

prakashsvmx:
Fixes https://github.com/minio/console/issues/3210

Issue Replication steps
create a  the policy as in https://github.com/minio/console/issues/3210

As an admin,

- create a bucket `clients` and some `prefixes`  like `test-user` and  `test-user-1` and upload files to each prefixes. 

- create 2 users  `test-user`  and `test-user1`  and assign the policy. 

- in a different browser, login as `test-user` and navigate to the prefix and logout.  
now in the same browser login as `test-user-1` and navigate . observe "Access Denied"  this is due to previous state  (bucket, prefix is retained) is not reset and request is made to the old state prefix. ( hard reload i.e Refresh of browser window would work as expected)

Verify the fix.
build and run the PR.
With this fix, the above steps would not result in "Access denied" error.

### DIFF
--- a/web-app/src/screens/LogoutPage/LogoutPage.tsx
+++ b/web-app/src/screens/LogoutPage/LogoutPage.tsx
@@ -15,7 +15,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import React, { useEffect } from "react";
-import { useNavigate } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


bexsoft (approved):
LGTM

reivaj05 (approved):

PR #3248, opened 2024-02-27
Use native WebSocket API

bayasdev:
This PR removes the external [websocket](https://www.npmjs.com/package/websocket) dependency and replaces it with the native WebSocket API provided by JavaScript.

It also renames `c` to `socket` according to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) for the WebSocket API and replaces HealthInfoMessage timestamp type from `Date` to `string`.

### DIFF
--- a/web-app/package.json
+++ b/web-app/package.json
@@ -28,8 +28,7 @@
     "recharts": "^2.10.3",
     "styled-components": "^5.3.11",
     "superagent": "^8.0.8",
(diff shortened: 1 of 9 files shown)


harshavardhana:
This is beautiful!

harshavardhana (approved):

reivaj05 (approved):

PR #3249, opened 2024-02-27
Check the blob size instead of the progress reports

### LINKED ISSUE #3243: "Error: Unexpected response, download incomplete." in Firefox
In Firefox: When I click download on any object I get "Unexpected response, download incomplete" in the "Uploads / Downloads" widget (close to the end of the progress bar, > 90%). In the network tab the only request that I see (https://console..../api/v1/buckets/.../objects/download?...) has status 200. Nothing in the minio log. Nothing in the browser console.

When I do the same in Chromium instead of Firefox, then the file downloads normally.
~~What I noticed: In firefox the download response payload seems to be base64 encoded, in chromium it isn't.~~ (probably just a display thing in firefox' console)

## Expected Behavior

Downloads work identically in Firefox and Chromium.

## Current Behavior

In Firefox it fails.

This is the only Minio cluster of mine that has this behavior and I have no idea when it started. Other clusters I operate running the same version of Minio work fine.

## Context

https://github.com/minio/minio/issues/19091

## Your Environment
* Version used (`minio --version`):
    ```
    minio version RELEASE.2024-02-17T01-15-57Z (commit-id=b6e98aed0161bb526aaf11fa1672d30eb89dca4b)
    Runtime: go1.21.7 linux/amd64
    License: GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
    Copyright: 2015-2024 MinIO, Inc.
    ```
* Server setup and configuration: It's deployed in a kubernetes 1.29.2 (k0s) using the minio operator (5.0.12). It's single node (both k8s and minio) and single pool (openebs-hostpath volume on xfs).
* Operating System and version (`uname -a`): `Linux ... 6.7.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 17 Feb 2024 14:02:33 +0000 x86_64 x86_64 x86_64 GNU/Linux`

pschichtel:
Fixes #3243 

I just had the epiphany to just simply check the size of the blob which carries the downloaded content, since that has already been decompressed by the browser.

### DIFF
--- a/web-app/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
+++ b/web-app/src/screens/Console/Buckets/ListBuckets/Objects/utils.ts
@@ -78,7 +78,6 @@ export const download = (
   toastCallback: () => void,
 ) => {
   let basename = document.baseURI.replace(window.location.origin, "");
-  let bytesLoaded = 0;
   const state = store.getState();
   const anonymousMode = state.system.anonymousMode;
 
@@ -107,7 +106,6 @@ export const download = (
     "progress",
     function (evt) {
       let percentComplete = Math.round((evt.loaded / fileSize) * 100);
-      bytesLoaded = evt.loaded;
       if (progressCallback) {
         progressCallback(percentComplete);
       }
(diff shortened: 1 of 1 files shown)


pschichtel:
I don't think we can access the blob before the ready state is DONE, so the progress bar will remain slightly off in firefox.

pschichtel:
Meh, so the dependency vulnerability check fails, but I definitely haven't touched any dependencies. Maybe this rebase helps?

cesnietor (approved):
Makes so much sense. 
Thanks for figuring it out @pschichtel.
I've tested it in Chrome, Firefox and Safari. For complete downloads as well as incomplete ones.
All works as expected.

PR #3270, opened 2024-03-19
Reuse GetClusterRegInfo from mc

anjalshireesh:
The logic related to calculating no of pools has changed in mc, but still old in console. Ensuring that the function from mc is reused here will avoid such issues in future.


go.mod
```diff
@@ -155,3 +155,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/minio/mc => github.com/anjalshireesh/mc v0.0.0-20240319134657-0e1592c105d9
```
bexsoft, line 159: Please remove replace

bexsoft (changes requested):

go.mod
anjalshireesh: @bexsoft will remove it after the corresponding PR https://github.com/minio/mc/pull/4881 is merged in `mc`.

anjalshireesh: @bexsoft done, PTAL.

shtripat (approved):
lgtm

prakashsvmx (approved):

harshavardhana (approved):

PR #3271, opened 2024-03-20
Changed default Replicate Existing Objects behavior in MinIO UI

bexsoft:
fixes https://github.com/minio/console/issues/3181

## What does this do?

- Fixed replicate existing objects configuration saving
- Displayed full error message if error case ocurrs.
- Changed wording for this feature & enabled by default this setting

## How does it look?

Screenshot 2024-05-05 at 12 19 49 a m
Screenshot 2024-05-05 at 12 24 14 a m

### DIFF
--- a/api/admin_remote_buckets.go
+++ b/api/admin_remote_buckets.go
@@ -292,7 +292,7 @@ func addRemoteBucket(ctx context.Context, client MinioAdmin, params models.Creat
 	return bucketARN, err
 }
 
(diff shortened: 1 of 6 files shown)


poornas:
`mc replicate add` turns on existing-objects replication by default if --replicate flag is not specified. The console enabling of existing objects by default is consistent with that - it is just the docs link above and wording displayed on console that needs to reflect that

bexsoft:
> `mc replicate add` turns on existing-objects replication by default if --replicate flag is not specified. The console enabling of existing objects by default is consistent with that - it is just the docs link above and wording displayed on console that needs to change to reflect that

I see, Including @ravindk89 in the loop for this

dvaldivia (approved):

poornas (approved):

PR #3280, opened 2024-04-02
fix ux for keys permissions

prakashsvmx:
fix ux for kms permissions

Fixes #3279
Test the below with Encryption:

Import
Create Key
Delete Key.

The actions should be enabled or hidden based on the permissions. 

Start MinIO and use play.min.io for testing with KES. 

```
 CI=true MINIO_ROOT_USER=minio MINIO_ROOT_PASSWORD=minio123  MINIO_KMS_KES_ENDPOINT=https://play.min.io:7373 MINIO_KMS_KES_KEY_NAME=my-minio-sse-kms-key MINIO_KMS_KES_KEY_FILE=./root.key MINIO_KMS_KES_CERT_FILE=./root.cert  minio server --address ":22000" --console-address ":10000" /tmp/data{1...4}

```

Deny Actions explicitly like:
```
 {
            "Effect": "Allow",
            "Action": [
                "kms:*"
            ]
        },
        {
            "Effect": "Deny",
            "Action": [
                "kms:DeleteKey",
                "kms:ImportKey"
            ]
        },
```


reivaj05:
@prakashsvmx Will this work in the case where we have permissions to delete/create the key in console but not in kms policies?

bexsoft (commented):
LGTM, Let's wait on @reivaj05 response

prakashsvmx:
Console does not have access/control to kes policies i believe.  And the UI elements rely only on IAM policy actions.  @reivaj05  . Please share details if I am missing something. 
Thank you

reivaj05:
@prakashsvmx this is helpful as well but i believe the main issue is when you try to delete a key but the policy in kms doesn't have the permissions to do it, we already have a way to get all the allowed actions in Console/KMS/Status.tsx:111, we should use what you implemented and also the allowed apis we can call in KMS

prakashsvmx:
@reivaj05 , if `kms:*` is not specified, console does not even show `Encryption` menu. 

would it not lead to dual configuration/ permission errors.? for example if user has IAM policy with `kms:*` and there are kes policies preventing API access  

```
{
            "Effect": "Allow",
            "Action": [
                "kms:*"
            ]
        },
````
and kes config like:
```
 deny:
    #- /v1/key/delete/my-minio-key*
    - /v1/key/delete/sph-*
    -  /v1/key/delete/*
```
?

**Comment by prakashsvmx** [comment]
@reivaj05  I also set the below for further testing:

```
policy:
  my-app:
    allow:
    - /v1/*
    deny:
    #- /v1/key/delete/my-minio-key*
    #- /v1/key/delete/sph-*
    - /v1/key/delete*
    - /v1/key/delete/*
    identities:
    - 5071bfc596449501154ad7421a5ef6860450d0ede2beb650b024c435c5bde539 # The client.crt identity
    # - 29a5efb99fe00832132b36dd0d6ee3a41355b503631af145549cfd35b78ef123
```
but the [api](http://localhost:5005/api/v1/kms/apis) returns the below

```json
...
{
    "method": "DELETE",
    "path": "/v1/key/delete/",
    "timeout": 15
}
```

so i am not sure about a reliable way to combine both IAM and KES policy.

reivaj05:
@prakashsvmx I thought the kms returned what you had in the config, you are right, it's going to be difficult to combine both if the response is only something lik `"/v1/key/delete/"`

Maybe we can merge this PR and we can discuss further for the other scenarios

cesnietor (approved):

Issue #3255, opened 2024-03-07
Make ShareLink URL point to Console instead to MinIO Server


cesnietor:
Currently whenever we want to share a file, the share url contains a default url which if for some reason is not reacheable by the browser it will fail to access it.

The proposal is to make Console handle the GET of the object instead of sending them to MinIO server.
That way it doesn't matter what the  MinIO server DNS is, since you can already access Console.

Right now:
<img width="543" alt="Screenshot 2024-03-06 at 5 27 54 PM" src="https://github.com/minio/console/assets/11819101/e4928f57-ea8d-4336-81db-d51e7faf8101">

## Expected Behavior
Share url uses console backend url and port
Use some other endpoint e.g. `https://play.min.io:9443/api/v1/get-file/cestest/file-s2.txt?X-Amz-Algorithm=A...`

## Current Behavior
Share url uses  127.0.0.1 url

## Steps to Reproduce (for bugs)
1. Add an Object
2.Click on an object
3. Click on Share File

bexsoft:
Steps Required:

- Create local endpoint under console to receive share requests
- Create a method where we can review the presigned key from share file to validate if request is valid
- Get the file as if we were in local download through Console UI.

ravindk89:
To clarify - @harshavardhana @bexsoft based on internal conversations, would this remove the need for us to use `MINIO_SERVER_URL` at all?

I've seen two user reports from community about having issues when using that envvar, vs not using it at all and simply specifying the FQDN of console to `--console-address`.

cesnietor:
No @ravindk89  we still need that to communicate with MinIO this is just sending the share files to console instead of MinIO.

dberardo-com:
is it possible to generate this URL programmatically, example from the CLI or from the Kubernetes Operator ?

prakashsvmx:
you can use mc share or any sdk https://github.com/minio/minio-go to generate presigned url

dberardo-com:
and can those link be permanent ? as in, never expire ?

prakashsvmx:
> never expire ? 

No.. it is as per s3 Max 7 days
>A presigned URL remains valid for the period of time specified when the URL is generated. If you create a presigned URL with the Amazon S3 console, the expiration time can be set between 1 minute and 12 hours. If you use the AWS CLI or AWS SDKs, the expiration time can be set as high as 7 days.

 https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html

[...]

PR #3284, opened 2024-04-05
Use Console as proxy for share object logic

cesnietor:
solves: https://github.com/minio/console/issues/3255

Share File URL now points to Console to facilitate sharing the url in case MinIO endpoint is not exposed.

Changes:
- Update how we return the share url pointing now to Object Browser API instead of MinIO endpoint
- Adds new endpoint to validate share link url (base64encoded) and download file
- unittests to validate url

How it looks:
Screenshot 2024-04-05 at 2 49 32 PM

When url expired:
Screenshot 2024-04-05 at 2 50 58 PM

### Test steps:

- Click share file
- Object should be downloaded


bexsoft (dismissed):
LGTM, Tested in local env as well as making a custom MinIO build with this fix.

prakashsvmx:
@cesnietor , Do we still need to limit  the number of hours as in the screenshot?

image

api/public_objects.go
```diff
@@ -0,0 +1,127 @@
...
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			http.Error(rw, resp.Status, resp.StatusCode)
+			return
+		}
+
+		urlObj, err := url.Parse(*inputURLDecoded)
+		if err != nil {
+			fmt.Printf("error parsing url '%s': %s", *inputURLDecoded, err)
```
harshavardhana, line 75: Frivolous log please remove.

api/public_objects.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		// Add the filename
+		_, objectName := url2BucketAndObject(urlObj)
+		escapedName := url.PathEscape(objectName)
+		rw.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", escapedName))
+
+		_, err = io.Copy(rw, resp.Body)
+		if err != nil {
+			fmt.Println("error copying response body:", err)
```
harshavardhana, line 87: Another frivolous log

api/public_objects_test.go
```diff
@@ -0,0 +1,87 @@
+package api
```
harshavardhana, line 1: Missing copyright

harshavardhana (changes requested):

api/public_objects_test.go
cesnietor: done

api/public_objects.go
cesnietor: done

prakashsvmx (approved):
LGTM :+1:

bexsoft (approved):
LGTM

PR #3302, opened 2024-04-18
Adds ExpireDeleteMarker status to bucketLifecycleRule UI display

jinapurapu:
Screenshot 2024-04-18 at 3 53 01 PM


web-app/src/screens/Console/Buckets/BucketDetails/BucketLifecyclePanel.tsx
```diff
@@ -188,6 +189,22 @@ const BucketLifecyclePanel = () => {
...
       },
     },
+    {
+      label: "Expire Delete Marker",
+      elementKey: "expire_delete_marker",
+      renderFunction: (el: LifeCycleItem) => {
+        if (!el) {
+          return <Fragment />;
+        }
+        if (el.expiration) {
```
cesnietor, line 199: since delete_marker can be undefined let's do this so that it's clear when it can be set and when it can't:
```
 if (el.expiration && el.expiration.delete_marker !== undefined) {
          return <span>{el.expiration.delete_marker ? "True" : "False"}</span>;
        }
```
This way if Delete Marker can't be set, it will be undefined.
It would look something like:
Screenshot 2024-04-18 at 3 54 58 PM

cesnietor (changes requested):

web-app/src/screens/Console/Buckets/BucketDetails/BucketLifecyclePanel.tsx
```diff
@@ -86,6 +86,7 @@ const BucketLifecyclePanel = () => {
           .getBucketLifecycle(bucketName)
           .then((res) => {
             const records = get(res.data, "lifecycle", []);
+            console.log("records", records);
```
cesnietor, line 89: please remove

cesnietor:
@jinapurapu  could you please rename the commit to use imperative language (what it Does) instead of what was done.
some comments on it [here](https://cbea.ms/git-commit/#imperative) .

bexsoft (approved):

cesnietor (approved):

Issue #3304, opened 2024-04-22
Path encoded in share link url is not url-safe


cesnietor:
## NOTE
If this case is urgent, please subscribe to [Subnet](https://min.io/pricing) so that our 24/7 support team may help you faster.

## Expected Behavior
URL should redirect and download files without any issues

## Current Behavior
returns a not found since the path contains a `/` on it e.g.
```
https://play.min.io:9443/api/v1/download-shared-object/aHR0cHM6Ly8xMjcuMC4wLjE6OTAwMC9jZXN0ZXN0L0F1ZGlvJTIwaWNvbi5zdmc/WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD01U0JFMFlQM0ZXVlFRUDZNNE1IRiUyRjIwMjQwNDIyJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDQyMlQxODE4MjdaJlgtQW16LUV4cGlyZXM9NDMyMDAmWC1BbXotU2VjdXJpdHktVG9rZW49ZXlKaGJHY2lPaUpJVXpVeE1pSXNJblI1Y0NJNklrcFhWQ0o5LmV5SmhZMk5sYzNOTFpYa2lPaUkxVTBKRk1GbFFNMFpYVmxGUlVEWk5ORTFJUmlJc0ltVjRjQ0k2TVRjeE16ZzBOekU1Tml3aWNHRnlaVzUwSWpvaVVUTkJUVE5WVVRnMk4xTlFVVkZCTkROUU1rWWlmUS5BSXFMQnR4bkVLS0NfUzBJX3RWUEhMODMyZUJwTW95Rkl0el9hWWVEWGF5aFNnNm5YMXhua2dUUjJzbHhnbEhYbUVsTVdUOUJuTk8zQjJscTVVWEVEZyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmdmVyc2lvbklkPW51bGwmWC1BbXotU2lnbmF0dXJlPTA3NjdjM2Q0YWMzNWI4YzFkYWZlMjlmYTEyNGI2NWRiNGNjMDExMmJkNWUyZGZhOGMzZGRlNTQyZDYxYTQ0MzE=
```

## Steps to Reproduce (for bugs)

1. Deploy minio on a kubernetes environment
[...]

PR #3305, opened 2024-04-22
Use url-safe base64 encoding for download-shared-object api

cesnietor:
fixes: https://github.com/minio/console/issues/3304

Uses URLEncoding  instead of StdEncoding to ensure url-safe strings.
Includes test.

### Test steps:

- With different filenames e,.g. `with .jpg, .svg, .png, spaces, etc`
- Click share file
- Object should be downloaded


bayasdev (approved):
Perfect

allanrogerr:
LGTM - tested

PR #3325, opened 2024-05-01
Fix a leak in WS object browser

vadmeste:
```
goroutine 7399330769 [chan send, 70126 minutes]:
github.com/minio/console/api.(*wsMinioClient).objectManager.func2.1()
        github.com/minio/console@v0.46.0/api/ws_objects.go:135 +0x6f0
created by github.com/minio/console/api.(*wsMinioClient).objectManager.func2 in goroutine 7354918912
        github.com/minio/console@v0.46.0/api/ws_objects.go:95 +0x45e
```

This commit fixes the leak above

### DIFF
--- a/api/ws_objects.go
+++ b/api/ws_objects.go
@@ -262,23 +262,17 @@ func (wsc *wsMinioClient) objectManager(session *models.Principal) {
 
 	defer close(done)
 
-	for {
-		select {
-		case writeM, ok := <-writeChannel:
-			if !ok {
-				return
-			}
-			jsonData, err := json.Marshal(writeM)
-			if err != nil {
-				LogInfo("Error while marshaling the response", err)
-				return
-			}
+	for writeM := range writeChannel {
+		jsonData, err := json.Marshal(writeM)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


api/ws_objects.go
```diff
@@ -250,27 +260,25 @@ func (wsc *wsMinioClient) objectManager(session *models.Principal) {
...
 	}()
 
-	// Write goroutine
-	go func() {
-		for {
-			select {
-			case <-done:
+	defer close(done)
+
+	for {
```
harshavardhana, line 265: Thanks for doing this, was doing this in my other PR but then reverted since it became too long.

harshavardhana (dismissed):

api/ws_objects.go
bexsoft, line 265: Can you please help us to fix linter notification? 

`api/ws_objects.go:265:2: S1000: should use for range instead of for { select {} } (gosimple)`

Thank you.

harshavardhana (approved):

bexsoft (approved):

Issue #3327, opened 2024-05-03
"illegal base64 data at input byte" when using certain share URLs generated via Console `1.3.0`


coffee-squirrel:
## Expected Behavior

All Share URLs generated via the Console UI should be valid and not cause exceptions.

## Current Behavior

When attempting to use (e.g. `curl -o file.tar "${SHARE_URL}"`) certain share URLs generated via Console `1.3.0` (via MinIO Server `RELEASE.2024-05-01T01-11-10Z`) I get a `HTTP 500` response with a body like:

```
{"detailedMessage":"illegal base64 data at input byte 147","message":"an error occurred, please try again"}
```

Other objects don't seem to have this issue, and so far the only unique aspect I can identify is the object size (~80GB vs 1GB or less for others I've tested).  There aren't any unique characters in the problematic objects' paths, and there's nothing in the MinIO Server log.

I was previously on `minio` `RELEASE.2024-04-18T19-09-19Z` and ran into the issue #3305 was intended to fix, which led to moving to `minio` `RELEASE.2024-05-01T01-11-10Z`.

Share URLs generated for the same objects via `mc share download` (`mc` `RELEASE.2024-04-29T09-56-05Z`) work as expected (but of course they go directly to MinIO Server).

## Possible Solution

Maybe https://github.com/minio/console/blame/e68a74ba488f38ec5be481cc358ecb68642a387a/api/public_objects.go#L96 should be `inputURLDecoded, err := b64.URLEncoding.DecodeString(inputEncodedURL)` to match the change made in #3305.

## Steps to Reproduce (for bugs)

1. Be on `minio` `RELEASE.2024-05-01T01-11-10Z`
2. Try generating+using share links for various objects-- maybe one that's very large.

## Context

I occasionally use Console-generated share links, but will now need to drop down to `mc share download` for certain objects (not a big deal).

## Regression

Somewhat (related to #3284 and #3305).

## Your Environment

* MinIO version used (`minio --version`): `RELEASE.2024-05-01T01-11-10Z (commit-id=7926401cbd5cceaacd9509f2e50e1f7d636c2eb8)`
* Server setup and configuration: single node single drive
* Operating System and version (`uname -a`): Windows Server 2022 (yeah I know :))

cesnietor:
ahh nice catch @coffee-squirrel  please feel free to send the PR and I'll gladly review it.

cesnietor:
[...]

PR #3328, opened 2024-05-03
Fix URL safe string decoding for DownloadPublicObject API

cesnietor:
fixes: https://github.com/minio/console/issues/3327

Similar to https://github.com/minio/console/pull/3305 it uses url safe base64 decoding for the download of the object. Used in Share File functionality. The previous PR fixed the creation of the URL (encoding), this fixes the decoding of it.

Includes unittest.

### Test steps:

- With different filenames e,.g. `with .jpg, .svg, .png, spaces, etc`
- Click share file
- Object should be downloaded


bayasdev (approved):
LGTM

jinapurapu (approved):

bexsoft (approved):

PR #3338, opened 2024-05-08
Updated golang to 1.21 in Dockerfile

lexpierce:
The current Dockerfile uses golang:1.19, which doesn't match either go.mod or the libraries in use, and doesn't build correctly. This should be updated to Go 1.21 at a minimum.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN make build-static
 
 USER node
 
-FROM golang:1.19 as golayer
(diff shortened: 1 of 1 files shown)


harshavardhana:
As you can see, the changes here won't make any difference in fact, this docker file must be purged from the repo.

harshavardhana:
https://github.com/minio/console/pull/3340 is the necessary PR.

PR #3340, opened 2024-05-08
remove all unnecessary Dockerfiles from repo

harshavardhana:


cesnietor:
We need to remove also:
`.dockerignore`
Also update https://github.com/minio/console/blob/master/DEVELOPMENT.md since it has some stale docker docs.

cesnietor:
@harshavardhana  did you do `make assets`? there's no need to do that, we do that only during releases.

harshavardhana:
> @harshavardhana did you do `make assets`? there's no need to do that, we do that only during releases.

was trying to the pdfjs-dist fixed but looks like that's not the right way to do it.

DEVELOPMENT.md
```diff
@@ -1,191 +0,0 @@
...
-...
-```
-
-### 3. Build `MinIO`
-
-Still in the MinIO folder, run
-
-```shell
-make build
-```
```
cesnietor, line 78: This is still valid and useful for development @harshavardhana

cesnietor (approved):
LGTM

bexsoft (approved):
LGTM, UI Vulnerability will be handled in a separate PR

PR #3341, opened 2024-05-09
Fix empty version in health report download

anjalshireesh:
In the "Download" option after generating health report, the version header was getting printed with an empty version number. This was happening because the version variable was shadowed inside a for loop.

Couple of other improvements:
- avoid gzipping the health info twice
- do not maintain list of tests to perform in console, simply use list of all tests from madmin


prakashsvmx (approved):
Verified. Able to download from console  and upload health to SUBNET  
```json
{"version":"3"}
{"version":"3","timestamp":"2024-05-17T07:05:36.632919482Z",
...
}
```

jiuker (approved):

bexsoft (approved):

PR #3351, opened 2024-05-21
Add ability to disable the share redirect feature

### LINKED ISSUE #3348: Add ability to disable Console as proxy for share object logic
#3284 has its use case, but introduced some bugs (#3305, #3328, #3334).

## Expected Behavior
Users can decide if they want to use Console as a proxy for the share logic.

## Current Behavior
Users are forced to use this new feature.

## Possible Solution
It would be great if one could disable this feature using an environment variable (f.e. `MINIO_BROWSER_PROXY_SHARE=false`, default `true`), since it introduced unneeded complexity if the `MINIO_SERVER_URL` is publically accessible.

Run a MinIO cluster with a recent Console version and run into the linked bugs.

## Context
The new feature broke the share functionality on our production cluster since not all fixes are provided in the latest version (1.4.0).

## Regression
This wasn't needed before #3284.

hashworks:
Introduces the ENV variable `CONSOLE_BROWSER_REDIRECT` to disable the feature introduced in #3348 since it is only needed when the MinIO server is not publically reachable.

It defaults to `on` to keep current behaviour.

Resolves #3348.

### DIFF
--- a/api/config.go
+++ b/api/config.go
@@ -307,3 +307,7 @@ func getConsoleAnimatedLogin() bool {
 func getConsoleBrowserRedirectURL() string {
 	return env.Get(ConsoleBrowserRedirectURL, "")
 }
+
+func getConsoleBrowserRedirect() bool {
+	return strings.ToLower(env.Get(ConsoleBrowserRedirect, "on")) == "on"
+}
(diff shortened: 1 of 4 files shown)


cesnietor:
Thanks for sending the PR, I answered in the issue https://github.com/minio/console/issues/3348#issuecomment-2135746112

PR #3353, opened 2024-05-21
Upgrade to Yarn 4 and fix vulnerability check

ramondeklein:
See #3354. This upgrades Yarn to v4 to enable advisory suppression.


ramondeklein:
Grrr... `--ignore` isn't supported in the normal `yarn audit`, only in `yarn npm audit` (which doesn't work for our repo).

harshavardhana:
@ramondeklein how about we move to latest yarn - we shouldn't have to use yarn1 we don't have a specific dependency.

ramondeklein:
> @ramondeklein how about we move to latest yarn - we shouldn't have to use yarn1 we don't have a specific dependency.

I was just working on that 👍🏻

ramondeklein:
The following changes have been made:
- Enable corepack to enable versioned package managers (required to run Yarn 4).
- Use the .nvmrc file in all places where the `${{ env.NVMRC }}` was used to install NodeJS.
- Use `yarn npm audit` (Yarn 4) instead of `yarn audit` (Yarn 1) and ignore the specific advisory.
- Fix the script to detect dead code (it looks like Yarn 4 doesn't use the default shell to run commands).

harshavardhana (approved):

harshavardhana:
This coverage limit check is a BS check can this be removed 

```
Total: 63.75 KiB, Transferred: 63.75 KiB, Speed: 110.09 KiB/s
Access permission for `play/builds` is set to `public`
grep to obtain the result
Result:
22.7%
It is smaller than threshold (65.0%) value, failed!
Error: Process completed with exit code 1.
```

?

harshavardhana (approved):

harshavardhana:
> @harshavardhana Minimum coverage of 65% percent also seems very low to me. If you want test coverage, then you should go for 100% coverage. It's the exceptions that always result in bugs. The standard codepath typically works fine, because everyone is using that every day.
> 
> This check was already there (and already failing for a while), but I can remove it.

I have changed the threshold to 1.0 to avail enough time to address the coverage whenever we can.

harshavardhana:
Okay, I'm going to merge this since we need to proceed to make a Console tag soon.

dvaldivia (approved):

.github/workflows/jobs.yaml
```diff
@@ -1103,7 +1162,7 @@ jobs:
           go tool cover -func=all.out | grep total > tmp2
           result=`cat tmp2 | awk 'END {print $3}'`
           result=${result%\%}
-          threshold=65.0
+          threshold=1.0
```
bexsoft, line 1165: We can have a discussion on this as sometimes this value fluctuates when we build the UI (No Golang code), I think we can keep this value at least in 25

bexsoft (approved):

PR #3361, opened 2024-05-24
Provide workaround for circular dependency

ramondeklein:
TL;DR If you create a PR in this repo that depends on changed code in the https://github.com/minio/minio repository, then make sure you push that code to your fork of https://github.com/minio/minio with the same branch name before pushing to the console repo.

### Details
This PR only changes the build procedure to provide a workaround to the cricular dependency that may result in a failing build. Suppose that you create a PR in the console that depends on certain code in the https://github.com/minio/minio repository, then the Github actions often fail, because it tried to build against the `master` branch of the MinIO repository in the MinIO account.

This PR changes the behaviour, because it will try to build against the MinIO repository in the account where the PR is originating from. So if you create a PR in https://github.com/ramondeklein/console with branch `example`, then it will try to checkout https://github.com/ramondeklein/minio (branch `example`) before it falls back to https://github.com/minio/minio (branch `master`).

I would have liked to avoid repeating code, but Github actions still doesn't support YAML anchors 👎🏻


harshavardhana (approved):

harshavardhana:
this provides a proper alternative we should take this in.

harshavardhana:
PTAL @donatello

harshavardhana:
PTAL @dvaldivia

donatello (approved):
Yeah we can try it this way. The problem only happens occasionally.

ramondeklein:
@harshavardhana Can you merge? I don't am authorized to merge in console.

PR #3363, opened 2024-05-28
chore: update all deps

harshavardhana:


CREDITS
```diff
@@ -355,235 +355,146 @@ SOFTWARE.
 
 ================================================================
 
-github.com/cheggaaa/pb
-https://github.com/cheggaaa/pb
+github.com/charmbracelet/x/ansi
+https://github.com/charmbracelet/x/ansi
```
ramondeklein, line 359: The URL isn't valid, because it's a directory in this repo. I think a global reference for github.com/charmbracelet/x (pointing to https://github.com/charmbracelet/x) would be sufficient. Otherwise, just combine all modules in this repo on line 358 and use a single URL.

ramondeklein (approved):
The URL isn't a big deal, so if this is urgent we can merge. I would prefer to have the CREDITS correct though...

CREDITS
harshavardhana: The URL is about what it references from the `go.mod` @ramondeklein don't worry about the actual URL

harshavardhana:
> The URL isn't a big deal, so if this is urgent we can merge. I would prefer to have the CREDITS correct though...



> The URL isn't a big deal, so if this is urgent we can merge. I would prefer to have the CREDITS correct though...

It is well understood that this is a GitHub limitation and that the sub-directories URLs are not very human friendly. 

For an example this is fine -> https://cloud.google.com/go/auth/oauth2adapt

ramondeklein:
I guess Github doesn't resolve it, because it would be hard to resolve if you have a directory called `pulls` or `issues`. So they choose you always have to reference it via `<repo>/tree/<branch>/<path>`, so https://github.com/charmbracelet/x/tree/main/ansi would be fine, but don't bother changing. Nobody is reading this file anyway...

Issue #3362, opened 2024-05-26
Applying lifecycle tag rules don't show up in Console UI


orientalperil:
## NOTE

I'm trying to apply lifecycle rules to a bucket.   The code works on S3 but not Minio.

## Expected Behavior

Lifecycle rules should be applied to the bucket that include the tag filter.

## Current Behavior

A rule is created but does not have the tag filter on it.

## Steps to Reproduce (for bugs)

I used this code:

```python
import boto3
from django.conf import settings

RETENTION = 'retention'
EXPIRE_FAST = 'expire_fast'

client = boto3.client(
    service_name='s3',
    aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
    aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
    endpoint_url=settings.AWS_S3_ENDPOINT_URL,  # http://127.0.0.1:6000
)


def s3_lifecycle_configuration():
    client.put_bucket_lifecycle_configuration(
        Bucket=settings.AWS_STORAGE_BUCKET_NAME,
        LifecycleConfiguration={
            'Rules': [
                {
                    'ID': 'Expire fast',
                    'Filter': {
                        'Tag': {
                            'Key': RETENTION,
                            'Value': EXPIRE_FAST,
                        },
                    },
                    'Status': 'Enabled',
                    'Expiration': {
                        'Days': 1,
                    },
                },
            ]
        }
    )

s3_lifecycle_configuration()
```

## Context

I want to create a lifecycle rule that applies to files with certain tags.

## Your Environment

* Version used (`minio --version`): RELEASE.2024-05-10T01-41-38Z (go1.22.3 darwin/amd64)
* Server setup and configuration: Python 3.10 with boto3
* Operating System and version (`uname -a`): macOS 13.6

![image](https://github.com/minio/console/assets/622699/9779eb49-df5d-4872-9513-5f167583b0e8)

Interesting the export shows the rule as I expect.  But in the web console you cannot see the filter.

```
$  ./mc ilm export myminio/my-bucket
{
 "Rules": [
  {
   "Expiration": {
    "Days": 1
   },
   "ID": "Expire fast",
   "Filter": {
    "Tag": {
     "Key": "retention",
     "Value": "expire_fast"
    }
   },
   "Status": "Enabled"
  }
 ]
}
```

orientalperil:
Great thank you @harshavardhana

PR #3369, opened 2024-06-05
Add ILM rule tags generated by mc to getBucketLifecycle response

jinapurapu:
Some ILM tags generated by mc are being presented differently from those in Console. This adds tags that are in this form to the tags to be displayed with the ILM rule in the editLifecycleConfiguration modal.

Closes https://github.com/minio/console/issues/3362


api/user_buckets_lifecycle.go
```diff
@@ -102,7 +102,12 @@ func getBucketLifecycle(ctx context.Context, client MinioClient, bucketName stri
 				Value: tagData.Value,
 			})
 		}
-
+		if len(rule.RuleFilter.Tag.Key) > 0 {
```
ramondeklein, line 105: I think this is more idiomatic Go:
```suggestion
		if rule.RuleFilter.Tag.Key != "" {
```

ramondeklein (changes requested):

ramondeklein (approved):

sen-io (approved):

PR #3370, opened 2024-06-06
go1.22.4 upgrade for vulncheck

harshavardhana:


.github/workflows/vulncheck.yaml
```diff
@@ -5,9 +5,6 @@ on:
   pull_request:
     branches:
       - master
-  push:
-    branches:
-      - master
```
harshavardhana, line 10: removed it since its unnecessary after a merge

shtripat (approved):

cesnietor (approved):
LGTM thanks

PR #3403, opened 2024-07-09
ws: Calculate the client IP from the WS request headers as well

vadmeste:
Currently, the websocket code adds an IP to X-Forwarded-For where the IP is calculated
from the TCP connection established to Console, however the established TCP connection
can be coming from the load balancer as well, hence the necesssity to evaluate the client 
IP based on X-Forwarded-For or X-Real-IP headers

Look for client IPs in the websocket request connection.

### DIFF
--- a/api/ws_handle.go
+++ b/api/ws_handle.go
@@ -177,9 +177,19 @@ func serveWS(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	clientIP := getSourceIPFromHeaders(req)
+	if clientIP == "" {
+		if ip, _, err := net.SplitHostPort(conn.RemoteAddr().String()); err == nil {
+			clientIP = ip
+		} else {
+			// In case there's an error, return an empty string
+			LogError("Invalid ws.RemoteAddr() = %v\n", err)
+		}
+	}
+
 	switch {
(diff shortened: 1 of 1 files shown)


api/ws_handle.go
```diff
@@ -177,9 +177,19 @@ func serveWS(w http.ResponseWriter, req *http.Request) {
...
 	}
 
+	clientIP := getSourceIPFromHeaders(req)
+	if clientIP == "" {
+		if ip, _, err := net.SplitHostPort(conn.RemoteAddr().String()); err == nil {
+			clientIP = ip
+		} else {
+			// In case there's an error, return an empty string
+			log.Printf("Invalid ws.clientIP = %s\n", err)
+		}
```
harshavardhana, line 187: use conn.RemoteAddr().String() itself as is if the splitHostPort fails.

harshavardhana (changes requested):
no random logs please

api/ws_handle.go
vadmeste: In fact, this would never happen in theory unless minio/websocket goes fairly crazy. It is better to keep logging this instead for easier debugging. I am using LogError() now.

harshavardhana (approved):

ramondeklein (approved):

PR #3421, opened 2024-08-08
Change KMSKeyInfo.CreatedAt to a time.Time

marktheunissen:
Discussion here: https://github.com/minio/minio/pull/20223#discussion_r1706917853

Other pull requests:
- https://github.com/minio/madmin-go/pull/296/files


marktheunissen:
Don't know why this JS build is failing in the CI...

ramondeklein:
We had the same failure in EOS console since yesterday, so looks like an external issue.

go.mod
```diff
@@ -151,3 +151,5 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/minio/madmin-go/v3 => github.com/marktheunissen/madmin-go/v3 v3.0.0-20240808031524-8a6ddcc05e14
```
cesnietor, line 155: we can't send this change

cesnietor (changes requested):

go.mod
ramondeklein: I think there was a TODO before that this should be changed before merging. I guess it needs it for now, until https://github.com/minio/madmin-go/pull/296 is merged.

cesnietor: yes, so blocking it so we don't merge it by mistake :D

ramondeklein:
This should fix the Javascript action: https://github.com/minio/console/pull/3423

go.mod
marktheunissen: It was intentional, just to run the tests in CI. Was planning on removing before merge, normally I leave a comment myself on the line as a reminder, will try do so in future always.

cesnietor (dismissed):

go.mod
marktheunissen: It has been replaced with the correct import version now

cesnietor: thank you @marktheunissen

marktheunissen:
@cesnietor govet complained about `errorsApi.ServeError(w, req, errorsApi.New(http.StatusUnauthorized, err.Error()))` which was not something I touched in the PR.

I have just pushed a commit to change that to `errorsApi.ServeError(w, req, errorsApi.New(http.StatusUnauthorized, "%s", err.Error()))` which will probably fix the linting but ultimately doesn't do anything as the .New() call handled it fine before.

Another option is to change the linter configuration, let me know.

ramondeklein:
@marktheunissen We've seen this in more repositories. It looks like the most recent `golangci-lint` version is more strict. We have issued PRs for both [MinIO](https://github.com/minio/minio/pull/20263) and [mc](https://github.com/minio/mc/pull/5023) already. Note that this is in fact a proper warning that needs fixing.

ramondeklein:
> I have just pushed a commit to change that to `errorsApi.ServeError(w, req, errorsApi.New(http.StatusUnauthorized, "%s", err.Error()))` which will probably fix the linting, but ultimately doesn't do anything, as the .New() call handled it fine before.

I think it warns, because there could be some unexpected behavior when the error message contains formatting instructions, such as `%1000d` or some other extreme stuff. Not sure if it could cause a crash when `%s` is used without passing any arguments...

harshavardhana:
Please look at a more comprehensive update @marktheunissen at #3425

marktheunissen:
@ramondeklein I don't disagree.... :) generally speaking, the vetting is correct.

My comment was just about this specific case, where the function being called is `errorsApi.New()` which is:

```go
// New creates a new API error with a code and a message
func New(code int32, message string, args ...interface{}) Error {
	if len(args) > 0 {
		return &apiError{
			code:    code,
			message: fmt.Sprintf(message, args...),
		}
	}
	return &apiError{
		code:    code,
		message: message,
	}
}
```

My comment was " the .New() call handled it fine before.". What I meant is that passing 0 args bypasses the call to `fmt.Sprintf`, but the linter doesn't know that. Edge case I suppose

@harshavardhana has completed this work in the PR https://github.com/minio/console/pull/3425 so I'll close this, thanks!

PR #3424, opened 2024-08-12
Fix content type issue

ramondeklein:
When the application returns index.html, it should always use Content-Type: text/html to ensure that the returned HTML is rendered. It previously used the MIME-type that was derived from the extension, which may cause erratic behaviour for some extensions.

Although http.ServeContent detects the content automatically, based on the extension we do need to set Content-Type: text/html explicitly. The http.FileServer is kind of crap, because you can't provide special handling for the (pretty common) scenario that the file can't be found. This is solved by intercepting some calls (via the notFoundRedirectRespWr type), but this doesn't prevent that the response headers are already prepared. That's why we need to set the content-type back to HTML again.

I also changed the logic, so the SPA file-server application is generated only once.

IMPORTANT: This issue doesn't occur when running via yarn start. Run  make build-static to update the local front-end, rebuild and start the back-end and connect via http://localhost:8444/ to reproduce the issue.

### DIFF
--- a/api/configure_console.go
+++ b/api/configure_console.go
@@ -313,6 +313,12 @@ func AuthenticationMiddleware(next http.Handler) http.Handler {
 
 // FileServerMiddleware serves files from the static folder
 func FileServerMiddleware(next http.Handler) http.Handler {
+	buildFs, err := fs.Sub(portal_ui.GetStaticAssets(), "build")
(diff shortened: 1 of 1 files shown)


bayasdev (approved):
LGTM

bexsoft (approved):

PR #3425, opened 2024-08-16
update deps to latest releases

harshavardhana:


api/admin_kms.go
```diff
@@ -257,10 +257,15 @@ func listKeys(ctx context.Context, pattern string, minioClient MinioAdmin) (*mod
 	return &models.KmsListKeysResponse{Results: parseKeys(results)}, nil
 }
 
+// printDate - human friendly formatted date.
+const (
+	printDate = "2006-01-02 15:04:05 MST"
```
harshavardhana, line 262: keeping this here same as `mc ls` output - lets not use time.RFC3339 since is visibly verbose.

api/admin_kms.go
```diff
@@ -257,10 +257,15 @@ func listKeys(ctx context.Context, pattern string, minioClient MinioAdmin) (*mod
...
+)
+
 func parseKeys(results []madmin.KMSKeyInfo) (data []*models.KmsKeyInfo) {
 	for _, key := range results {
 		data = append(data, &models.KmsKeyInfo{
-			CreatedAt: key.CreatedAt,
+			CreatedAt: key.CreatedAt.Format(printDate),
```
ramondeklein, line 268: Doesn't this conflict with https://github.com/minio/madmin-go/pull/296?

ramondeklein (changes requested):

api/admin_kms.go
harshavardhana: conflict?

ramondeklein: I guess I wasn't awake. We need to format it, because now it's a `time.Time` structure. Not sure what I was thinking...

ramondeklein (approved):

PR #3429, opened 2024-08-30
Support madmin-go new RemoveTier API signature

vadmeste:
Requires https://github.com/minio/madmin-go/pull/301 And https://github.com/minio/mc/pull/5031


vadmeste:
This is not needed anymore, as we are adding a new v2 API in madmin-go

Issue #3427, opened 2024-08-29
MinIO Console UI should report failure to communicate with MinIO during login as "Server not reachable" instead of "Invalid Login"


ecerulm:
**Is your feature request related to a problem? Please describe.**

I had a new setup with HAProxy and 2 minio servers, and I couldn't login via the MinIO console at :9001. 

I was getting "Invalid Login" (both in the json response that I could see fromt he browser > developer tools> network tab) and in the UI as a red banner saying "Invalid login". 

The culprit was I used `MINIO_SERVER_URL` and that somehow prevented MinIO console from reaching the MinIO server. 

But really the "Invalid Login" was misleading me into thinking there was an issue with the credentials. At the end I got help from @marktheunissen in slack , and he identified that I could get rid of the `MINIO_SERVER_URL`. 

My point is that the error message in this case it should not be "Invalid Login" but "Server not reachable" or "Server invalid TLS" or something like that. That will help MinIO beginners like me that mistakenly thought MINIO_SERVER_URL was required when using HAProxy. 

**Describe the solution you'd like**
Any of the following
* at startup minio console part can do a sanity check about the reachability of the server , and print something in the startup log
* at runtime the console can rely some of the underlying error to the browser client. I understand that you don't want to leak the full error to the browser , but specific error like "server unreachable" , "server invalid TLS certificate", etc I believe do not leak important information about the setup

**Describe alternatives you've considered**

**Additional context**

![image](https://github.com/user-attachments/assets/501b2796-5750-4e87-9e76-ad1fd12aa1f9)

The correct message in the image above should have been "server unreachable" (or whatever was the actual top level cause),

ecerulm:
@harshavardhana mentioned in https://github.com/minio/docs/pull/1312#issuecomment-2317206253 
> It isn't done in vain, it was an ask by the customer to avoid reporting "login errors" to avoid brute force attacks on the login vector in the name of security by (obscurity)
> 
> We used to be more descriptive in the past but it was implemented to make it vague on purpose.

I understand that you don't want to leak information about the login process that could be used by an attacker to learn about the system. 

But in the specific case of "MinIO Console can't reach MinIO Server" I believe it would not leak anything. In a way, MinIO Console is already leaking that there is a connectivity issue between console and server since it takes 10 seconds to send the "Invalid Login". If you just write the incorrect credentials it tells you right away "Invalid Login" if the console can't communicate with server it "waits" 10 seconds and then gives the "Invalid Login".

[...]

PR #3432, opened 2024-09-03
Return network error when logging in and the network connection fails

marktheunissen:
Displays "unable to login due to network error" to the user when a network error is encountered.

- Works with scenarios where the DNS does not resolve, or the port is closed or there is a timeout, etc. 
- We have InsecureSkipVerify set on the http.Client now, so certificate errors are no longer detected by our http client as errors.

@harshavardhana Up to you, I know MINIO_SERVER_URL is deprecated but it looks like people still use it and get into trouble with DNS. 

Fixes: https://github.com/minio/console/issues/3427


harshavardhana:
```
=== NAME  Test_PolicyListGroupsAPI
    policy_test.go:740: 
        	Error Trace:	/home/runner/work/console/console/integration/policy_test.go:740
        	Error:      	Not equal: 
        	            	expected: "[\"testgroup12345\"]\n"
        	            	actual  : "[\"testgroup12345\",\"testgroup12345\"]\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-["testgroup12345"]
        	            	+["testgroup12345","testgroup12345"]
        	            	 
        	Test:       	Test_PolicyListGroupsAPI
=== RUN   Test_PolicyListGroupsAPI/List_Users_for_Policy_-_Invalid
--- FAIL: Test_PolicyListGroupsAPI (0.05s)
```

This looks like some duplicate groups are being listed @marktheunissen - needs a fix on the minio upstream side.

marktheunissen:
@harshavardhana I've tracked this error down to a specific location, this integration test is adding groups and then when getting them back, the duplicates are being introduced.

The group name `testgroup12345` is appended here: https://github.com/minio/minio/blob/master/cmd/iam-store.go#L995

And then the same group name `testgroup12345` is appended again here: https://github.com/minio/minio/blob/master/cmd/iam-store.go#L1005

The function was modified and no longer does `if store.getUsersSysType() == LDAPUsersSysType {` so I wonder if that would be the issue?

I can keep working on it but I thought you might be able to find the issue faster than me since it's the first time I've looked at this.

marktheunissen:
Made a simpler test that is easy to work with and reached out to a few people here: https://github.com/minio/minio/pull/20388/files

.github/workflows/vulncheck.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.22.5
+          go-version: 1.22
```
marktheunissen, line 22: Changing this to 1.22 instead of 1.22.7, as I think the rest of the build pipelines all use the latest go patch version... correct me if I'm wrong.

api/errors.go
```diff
@@ -111,6 +112,11 @@ func ErrorWithContext(ctx context.Context, err ...interface{}) *CodedAPIError {
 				errorCode = 401
 				errorMessage = ErrInvalidLogin.Error()
 			}
+			if errors.Is(err1, ErrNetworkError) {
+				detailedMessage = ""
+				errorCode = 401
```
ramondeklein, line 117: Shouldn't we report `503 Service Unavailable` instead?

marktheunissen: Yep will change it

ramondeklein (dismissed):

.github/workflows/vulncheck.yaml
bexsoft: Had to hardcode the version as github workflows sometimes keep old go versions and denies to upgrade by itself

bexsoft (changes requested):
Pleasew review my comment

PR #3435, opened 2024-09-09
Use setup-node@v4's new node-version-file directly

fulldecent:


ramondeklein (approved):

ramondeklein:
I guess we should use this everywhere else too. It's cleaner...

prakashsvmx (approved):

PR #3453, opened 2024-10-14
Fix Set Policy on Groups with spaces

cesnietor:
It was not possible to set a policy on a group with spaces:

Screenshot 2024-10-14 at 1 48 04 PM

We had a regex to validate the group name in the request but current UI doesn't allow the user to define that manually, the groups should already exists, so technically is not necessary. Right now this regex is wrong cause it doesn't allow spaces and it should. We could fix it but it but it might not allow other stuff, here we can rely on the server to return that error if something was wrong.

Screenshot 2024-10-14 at 2 53 29 PM

### Test Steps:

1. First:
- Create a group with spaces like e.g. `some group`
- Click on new group then Policies
- Click on Assign policies
- Assign policies and click save
- All should work

3. Also: Creating two groups
- In Listing Policies
- Select the two new groups
- Select ASsign policies
- Select policies to attach 
- Save (should work fine)


bayasdev (approved):
LGTM

reivaj05 (approved):

PR #3470, opened 2024-11-11
Deprecated Lifecycle and Tiering UI

bexsoft:
## What does this do?

Deprecated Lifecycle and Tiering UI, Functionality can still be configured using mc


dvaldivia (approved):

bayasdev (approved):

tkempken:
> ## What does this do?
> Deprecated Lifecycle and Tiering UI, Functionality can still be configured using mc

@bexsoft + @marktheunissen:
Is there a public statement available anywhere explaining why tiering and site replication were removed, or why they had to be?

Is there any information on whether these features will also be removed from mc/mcli or minio server itself in the future? I currently dont understand the reasoning and unfortunately I cant find any explanation despite extensive searching.

Thanks in advance!

ramondeklein:
The functionality will remain in MinIO and also in `mc` and `madmin` (MinIO administration API client for Go). The problem is that we can't allocate enough resources to keep all web interfaces up-to-date. We rather do it good, then do it half. That's why we have removed UI features that are not often used and were not up-to-date with the current status in MinIO itself.

We have released an Enterprise offering, but that includes much more than just MinIO. It allows multi-site management, including catalog, observability, improved firewall/loadbalancing, object prompting (AI) and much more. The enterprise UI is also a completely different product compared to MinIO console. That product is really aimed at enterprise. Small installations typically don't need those features.

PR #3472, opened 2024-11-12
Release v1.7.4

bexsoft:


bayasdev (approved):

reivaj05 (approved):

donatello:
Do we merge with the CI failure?

bexsoft:
> Do we merge with the CI failure?

This PDF issue is not a real failure, PDF is rendered correctly in the UI Locally I see an error related to pdf worker library required for this and testcafe. This needs further investigation

bayasdev (approved):

cesnietor (commented):
Please add Changelog.md

ramondeklein (commented):
@bexsoft Did you regenerate the files after rebasing to the latest `master`? When I run `make release`, then all generated files are different but that may be caused due to different versions. I do think the `build` folder should be identical and not depend on whoever runs `make release`.

For that reason, I am always a fan of letting the CI/CD build the final output instead. Then it's consistent.

bexsoft:
> @bexsoft Did you regenerate the files after rebasing to the latest `master`? When I run `make release`, then all generated files are different but that may be caused due to different versions. I do think the `build` folder should be identical and not depend on whoever runs `make release`.
> 
> For that reason, I am always a fan of letting the CI/CD build the final output instead. Then it's consistent.

Yes, @ramondeklein, all files are updated with the latest master change. we use a release script to generate these files, maybe could be related

ramondeklein (approved):
WIth the removal of functionality, this isn't a patch. It should at least be named v1.8.0. But I don't think version number is actually being used -outside MinIO-, so I guess it's fine...

feorlen:
@bexsoft corresponding docs ready to merge

cesnietor (approved):

reivaj05 (approved):

bayasdev (approved):

PR #3496, opened 2025-01-23
Support login request with leading or trailing whitespace from payload

### LINKED ISSUE #3422: Trailing whitespace in username leads to invalid credentials
## Expected Behavior

As a user, I would expect trailing spaces to be removed in case I copied them with trailing spaces from messages, mails or something else. 

## Current Behavior

The current version uses the `username` as it is. Therefore, trailing spaces are used as part of the login and cause an invalid login, which just leads to highlighted fields without any further information.

image

Leads to the following outcome:

image

The trailing space is hard to notice, and the UI is giving no hint.

## Additional information

Thanks for the awesome work ❤️

changchaishi:
Fixes #3422

### DIFF
--- a/api/user_login.go
+++ b/api/user_login.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"reflect"
 	"strings"
 
 	"github.com/go-openapi/runtime"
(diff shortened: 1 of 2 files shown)


models/login_request.go
```diff
@@ -24,6 +24,8 @@ package models
 
 import (
 	"context"
+	"reflect"
```
marktheunissen, line 27: Thanks for the PR! 

FYI, `// This file was generated by the swagger tool.` ... in other words, it should not be edited directly, but rather generated from the swagger definition. I don't see any corresponding edits to a swagger file that would generate this ... perhaps you still need to commit it?

changchaishi: Oh, I see, thanks. I did not understand it correctly. Now I had moved the logic to the handler layer.

api/user_login.go
```diff
@@ -124,11 +125,24 @@ func getConsoleCredentials(accessKey, secretKey string, client *http.Client) (*C
...
+func trimWhitespace(lr *models.LoginRequest) {
+	val := reflect.ValueOf(lr).Elem()
+
+	for i := 0; i < val.NumField(); i++ {
+		field := val.Field(i)
+		if field.Kind() == reflect.String {
+			field.SetString(strings.TrimSpace(field.String()))
+		}
+	}
+}
```
ramondeklein, line 138: We shouldn't use `reflect` for this. We should call `strings.TrimSpace()` on the relevant fields.

api/user_login.go
```diff
@@ -124,11 +125,24 @@ func getConsoleCredentials(accessKey, secretKey string, client *http.Client) (*C
...
+
 // getLoginResponse performs login() and serializes it to the handler's output
 func getLoginResponse(params authApi.LoginParams) (*models.LoginResponse, *CodedAPIError) {
 	ctx, cancel := context.WithCancel(params.HTTPRequest.Context())
 	defer cancel()
 	lr := params.Body
+	trimWhitespace(lr)
```
ramondeklein, line 145: No need to call a function here. Just trim the fields from within this function is totally fine.

integration/login_test.go
ramondeklein, line 1: Good coding ethics to add a test. 👍

ramondeklein (changes requested):

ramondeklein (approved):

PR #3500, opened 2025-02-07
fix: possible nil deref in admin_info

wooffie:
We check inStep here:
https://github.com/minio/console/blob/31a5d640ebe3422dddcdb96163afc083bd34c92b/api/admin_info.go#L1121
But not check here before use:
https://github.com/minio/console/blob/31a5d640ebe3422dddcdb96163afc083bd34c92b/api/admin_info.go#L1128


ramondeklein (approved):
LGTM

cesnietor (approved):
LGTM, thanks.

PR #3509, opened 2025-02-26
Implemented AGPL MinIO Object Browser simplified Console

bexsoft:
## What does this do?

New AGPL Object Browser simplified Console

## How does it look?

Screenshot 2025-02-25 at 9 55 57 p m
Screenshot 2025-02-25 at 9 48 30 p m
Screenshot 2025-02-25 at 9 48 17 p m

PR #3517, opened 2025-03-13
Clean-up after PR 3509

ramondeklein:
This PR implements the following changes:
- Obsolete after https://github.com/minio/console/pull/3509 has been removed.
- Change mock structures to avoid using global variables (can cause issues when running `go test` due to parallelism).
- Upgrade packages to fix Go and Javascript vulnerabilities

I was cleaning up the redundant code in upstream console and I noticed it was full of global variables for mocking. Global variables should be avoided and especially in Go unit tests, because tests may run in parallel when you run `go test ./...`. Unless `t.Parallel()` is called, tests _within a single package_ run sequentially. But Go does run multiple packages in parallel, so it may be tricky when the mocked type is used in multiple packages.

It's often easy to convert the globals to fields inside the struct, so I would propose to avoid global variables unless it's unavoidable. Made the change in console in [this commit](https://github.com/minio/console/pull/3517/commits/ba913314e1180095a8654f82246392bcf2c9eaec). Test code should mostly follow the same coding practices as normal code. We shouldn't be sloppy, because it's "just" test code.

PS: The mocking types were private, so I won't expect problems with the previous implementation. But it's often that a private type is made public later, so I rather keep code safe.


ramondeklein:
I tried to upgrade `@babel/runtime` to `v7.26.10`, but wasn't successful. @bexsoft Can you take a look?

cesnietor (commented):
LGTM just saw a test failing. plus some linting :
```
type `wsAdminClient` is unused (unused)
func `sendWsCloseMessage` is unused (unused)
```

ramondeklein:
@cesnietor These linter issues have already been fixed in commit https://github.com/minio/console/pull/3517/commits/0a6a9a1bcf7ee051210e3e7c49ecaac457e8a681. Not sure why you are still seeing them.

The failing test has nothing to do with this PR. It also fails on `master`. I did briefly look at it. It depends on an OpenID server running on `dex:5556`, so I guess this needs some additonal infra to run.

cesnietor:
Now disablebucket is failing :D @ramondeklein 
```
--- FAIL: Test_disableBucketEncryption (0.00s)
    --- FAIL: Test_disableBucketEncryption/Bucket_encryption_disabled_correctly (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1327fb8]

goroutine 180 [running]:
testing.tRunner.func1.2({0x1465b20, 0x236b050})
	/opt/hostedtoolcache/go/1.23.7/x64/src/testing/testing.go:1632 +0x230
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.23.7/x64/src/testing/testing.go:1635 +0x35e
panic({0x1465b20?, 0x236b050?})
	/opt/hostedtoolcache/go/1.23.7/x64/src/runtime/panic.go:791 +0x132
github.com/minio/console/api.minioClientMock.removeBucketEncryption(...)
	/home/runner/work/console/console/api/user_buckets_test.go:118
github.com/minio/console/api.disableBucketEncryption(...)
	/home/runner/work/console/console/api/user_buckets.go:499
github.com/minio/console/api.Test_disableBucketEncryption.func3(0xc000350680?)
	/home/runner/work/console/console/api/user_buckets_test.go:457 +0x90
testing.tRunner(0xc000350680, 0xc0000d4c40)
	/opt/hostedtoolcache/go/1.23.7/x64/src/testing/testing.go:1690 +0xf4
created by testing.(*T).Run in goroutine 179
	/opt/hostedtoolcache/go/1.23.7/x64/src/testing/testing.go:1743 +0x390
```

ramondeklein:
@cesnietor Sorry, I didn't check it right. All checks now pass...

cesnietor (approved):

PR #3526, opened 2025-04-03
Release v2.0.0

bexsoft:
## What does this do?

- Console Release v2.0.0
- Updated swagger-typescript-api generation


dvaldivia (approved):

feorlen:
Docs note: this is the release where the only thing you can still do in the console is log in and view/upload objects in an already existing bucket.

yardenshoham:
Why remove so much functionality?

harshavardhana:
> Why remove so much functionality?

Community version is going to back to what it was before all these changes, it is basically object browser.

cesnietor (approved):
Tested lgtm. Only noticed that we have a way to create buckets but no deletion which is weird.
Also the searchbar on the top looks out of context, maybe should be within the context of the bucket, looks as if it's for all buckets cause it's above the bucket name cc @bexsoft.

cesnietor:
@bexsoft  please update the changelog.md

bexsoft:
> @bexsoft please update the changelog.md

Updated.

PR #3541, opened 2025-05-16
Updated project dependencies

bexsoft:
## What does this do?

- Updated `web-app` project dependencies to the latest versions
- Updated mds to v1.1.3 which contains the new logo version for Object Store community version

## How does it look?

Screenshot 2025-05-16 at 12 05 09 p m


.github/workflows/jobs.yaml
```diff
@@ -689,23 +689,6 @@ jobs:
           path: |
             ./integration/coverage/
           key: ${{ runner.os }}-coverage-2-${{ github.run_id }}
-  react-tests:
```
cesnietor, line 692: why removing this?

bexsoft: Those tests were related to a set of functions that were removed. These functions were used in Operator console and are no longer needed here.

dvaldivia (approved):

PR #3557, opened 2025-06-18
Decreased Browser direct download threshold to 5GiB

bexsoft:
fixes https://github.com/minio/object-browser/issues/3539

## What does this do?

Decreased the direct download threshold to 5GiB so browser download manager handles these downloads instead of MinIO Object Manager on slower systems


reivaj05 (approved):

cesnietor (approved):
vulnerability issue will be addressed in separate PR

